PR #1, opened 2015-10-16
Daemonize BESS

apanda:
- By default bessd now starts daemonized.
- Use -w to run without daemonization.
- When daemonized output goes to syslogd


sangjinhan:
Done with #2. Thanks

PR #3, opened 2015-10-16
PMD driver reverses txq rxq parameter

apanda:
The PMD driver's rte_eth_dev_configure call had rxq and txq in the wrong order.
This is not a problem currently (we only have 1 rxq and txq) but might be
problematic later


sangjinhan:
Thanks :+1:

PR #4, opened 2015-10-19
Fix Makefile

apanda:
When statically linking DPDK, not passing in -Wl,--whole-archive means that some
constructors are not called, leading to issues.

This was the original cause for bess not finding PMD ports when using 82599 NICs.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -2,7 +2,7 @@ CC = gcc
 DPDK_LIB = dpdk
 
 ifndef RTE_SDK
-    RTE_SDK = ../deps/dpdk
+    RTE_SDK = $(abspath ../deps/dpdk)
 endif
 
 ifndef RTE_TARGET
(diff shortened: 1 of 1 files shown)


apanda:
Actually this is a good point, I keep running into this issue where I forget to unset RTE_SDK, and it really doesn't work, so maybe we should do this too.

sangjinhan:
Applied. Thanks!

Yeah, the original idea was to allow users to use their own version of DPDK, although I never tested this scenario myself. :)

Our changes from the original configuration are:
* CONFIG_RTE_BUILD_COMBINE_LIBS=y (was n)
* CONFIG_RTE_LIBNAME="dpdk" (was intel_dpdk)
* CONFIG_RTE_LIBRTE_KNI=n (was y)
* CONFIG_RTE_LIBRTE_VHOST=y (was n)

The first one allows us to not enumerate all the individual dpdk components in Makefile. The libname was changed since "dpdk" is default in 2.1. KNI causes build errors with very recent kernel versions, but anyways we have our own vport so KNI is not required.

Given that none of them are strictly required to build BESS (VHOST is the only exception, but we can selectively load the driver by checking CONFIG_RTE_LIBRTE_VHOST in vhost_driver.c), I thought supporting custom DPDK would be not a bad idea.

I think we have three options in the long term:
1. Forcing deps/dpdk
2. Enumerating all DPDK components in our Makefile (we used to do this)
3. Using DPDK-provided Makefile (as long as it's not too intrusive): http://dpdk.org/doc/guides/prog_guide/dev_kit_build_system.html#makefile-description

Comments? :)

PR #6, opened 2015-11-05
Add loopback support to vport and assign RXQs to CPUs smartly

apanda:


apanda:
Updated with both things. I see some increased in overall performance (6.66 Mpps for 4 vs 2.475 with 1) but is not linear. Also it slows down very quickly, e.g. with 8 cores I only get 4.4 Mpps.

sangjinhan:
Applied. Thanks.

For performance test, I added perftest/loopback_vport.bess. One bug found: if I repeat reset and running the script, vport stops echoing packets. Could you take a look?

apanda:
Sure.



On Nov 8, 2015, at 12:10 AM, Sangjin Han <notifications@github.com> wrote:

Applied. Thanks.

For performance test, I added perftest/loopback_vport.bess. One bug found:
if I repeat reset and running the script, vport stops echoing packets.
Could you take a look?

—
Reply to this email directly or view it on GitHub
<https://github.com/NetSys/bess/pull/6#issuecomment-154788475>.

PR #9, opened 2015-11-10
Fix resource leakage when deiniting port

apanda:


sangjinhan:
Thanks, applied. :+1:

PR #10, opened 2015-11-24
fix bug for increasing snobj buffer size when it needs to be more than doubled

keonjang:


colin-scott:
Google's writing code for us! :-)

keonjang:
Not yet :)

sangjinhan:
Applied. Thanks!

PR #11, opened 2015-12-09
Blobs were not decoded correctly by libbess-python

apanda:
We did not update offset correctly previously


sangjinhan:
Applied. Thanks!

PR #17, opened 2015-12-26
'show tc' now shows resource limits

joshreich:
Added setting output to 'show tc', updated README w/ instructions for binding phys ports to DPDK.


sangjinhan:
Applied. Thanks, and sorry for the delay!

PR #19, opened 2016-01-06
Fix nvport library and examples work with changes

apanda:
This also gets rid of the need to place libsn in /usr/lib and no longer depends
on having a dynamically linked dpdk build


sangjinhan:
Thanks. Could you also:

1. Change the filename (from vport_driver.c to something else... e.g., vport_zc)?
2. In the definition of "struct driver", NativeVport should be NativeVPort (large P). ZecoCopyVPort would also work.

sorry for being picky!

apanda:
Ping

sangjinhan:
Pong

PR #22, opened 2016-02-29
Fix problems with native apps

apanda:
Mempool naming has changed this tries to fix that problem, and fix the build.


sangjinhan:
I think the code should try from 1k to 512k, not the other way around. DPDK has a bug that a mempool name lingers even if its allocation has failed. For example if 512k alloc failed but the next trial 256k worked, then there would be both pframe0_512k and pframe0_256k, but the latter one should be used.

apanda:
Ah, I see. Will fix.

apanda:
Done, also it is from 16K to 512K (your code doesn't go below that).

sangjinhan:
Applied. Thanks.

PR #24, opened 2016-04-23
Multi-instance features and minor bessctl improvements

MurphyMc:
This primarily adds a few features to bessd to allow supporting multiple instances.  It also makes some tweaks to documentation/messages, fixes a bessctl bug, and adds a simple bessctl command.


sangjinhan:
Applied. Thanks!

PR #29, opened 2016-06-23
Add python 3 support to libbess-python

tootoonchian:
Retains python 2 compatibility.


tootoonchian:
Commit b2a4a03c7704378fe29fc40aff6a6f2c41aa339b breaks this patch (long doesn't exist in Python 3). Let me look into a way to resolve it first...

sangjinhan:
Will revisit this issue when Python 3 support is necessary

PR #38, opened 2016-10-12
build script should not return 0 when it fails

changlan:
This minor fix is useful for continuous integration.


sangjinhan:
Great, thanks!

PR #43, opened 2016-10-15
update gcc version to 5 on travis

changlan:
Make sure Travis works for the C++ branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
+sudo: required
 dist: trusty
 
+env:
+  - CXX=g++ CC=gcc
+
 install:
-  - sudo apt-get -qq update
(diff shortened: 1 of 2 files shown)


sangjinhan:
Would it be possible for travis to test building BESS with both gcc 4 and 5? We want to make sure BESS builds on the newest gcc, but at the same time we want to keep compatibility with the gcc version (4.7?)  used by Ubuntu Trusty.

changlan:
Yes. GCC 4 support is added. The reason I upgrade to GCC 5 is that the current build does not compile with GCC 4. But if you want to keep compatibility with default GCC on trusty, we can fix the code later.

sangjinhan:
Merged, thanks! 1002c1c and eee5825 should fix the build issues on gcc 4.

PR #45, opened 2016-10-17
Build fixes to address:

barath:
*  Missing log.h in module.h for log_err() symbol
*  Added -mavx compile flag to avoid build errors on AVX code.


sangjinhan:
Thanks! Non-AVX machines should not build BESS with AVX instructions. If a machine does support AVX, -mavx should have been implied by -march=native. I think the build error was caused due to missing header file utils/simd.h. 72bff87 should fix this issue.

PR #46, opened 2016-10-17
port drivers to c++

melvinw:


sangjinhan:
Thanks, applied!

sangjinhan:
You can use just "{}" as a member initializer instead of {0} or {{0}}, regardless of member types.

melvinw:
Gotcha

PR #51, opened 2016-10-17
Adds build checks for glog and gflags and documents their requirement.

### LINKED ISSUE #49: Software dependency needs to be updated
Since we are now using gflags and glog, the following files need to be updated.

build.py: `check_essential()`
docs/install.md

barath:
Fix #49

### DIFF
--- a/build.py
+++ b/build.py
@@ -127,8 +127,8 @@ def check_essential():
         sys.exit(1)
 
     required('pcap/pcap.h', 'libpcap-dev')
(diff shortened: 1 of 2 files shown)


sangjinhan:
Do we need both *-dev and the runtime library for glog and gflags? In most cases the *-dev package alone should suffice...

barath:
You're right -- not absolutely necessary.  I can remove it.

barath:
Updated to remove extra packages.

sangjinhan:
Merged, thanks!

PR #54, opened 2016-10-18
port/ditch some home-grown data structures

melvinw:


core/oldtests/htable.cc
```diff
@@ -402,23 +401,21 @@ static void perftest() {
...
-     dpdk_discrete_close},
-    {"rte_hash_lookup_data", dpdk_embedded_init, dpdk_data,
-     dpdk_embedded_close},
-    {"rte_hash_lookup_with_hash_data", dpdk_embedded_init, dpdk_data_hash,
-     dpdk_embedded_close},
-    {"rte_hash_lookup_bulk_data(x16)", dpdk_embedded_init, dpdk_data_bulk,
-     dpdk_embedded_close},
+      {"ht_get", bess_init, bess_get, bess_close},
+      {"ht_inlined_get", bess_init, bess_inlined_get, bess_close},
+      {"ht_inlined_get_bulk(x16)", bess_init, bess_inlined_get_bulk,
```
sangjinhan: This line should be guarded with #if __AVX__.
(Barath's laptop doesn't have AVX so build fails)

core/utils/htable.h
```diff
@@ -54,72 +77,127 @@ struct ht_bucket {
   ht_keyidx_t keyidx[ENTRIES_PER_BUCKET];
 } __ymm_aligned;
 
-struct htable {
-  /* bucket and entry arrays grow independently */
-  struct ht_bucket *buckets;
-  void *entries; /* entry_size * num_entries bytes */
+template <class K, class V, ht_keycmp_func_t C = memcmp,
```
sangjinhan: nitpick: We prefer typename over class :-)

sangjinhan:
One of the reasons why we switched to C++! 👍

PR #57, opened 2016-10-20
fix up metadata

melvinw:


core/metadata.cc
```diff
@@ -1,10 +1,12 @@
...
 #include <queue>
 #include <string.h>
+#include <vector>
 
 #include <glog/logging.h>
 
-#include "mem_alloc.h"
 #include "module.h"
 
 #include "metadata.h"
```
sangjinhan: This must be included first (Google style guide)

core/metadata.cc
```diff
@@ -29,8 +30,8 @@ bool scope_component_less(const struct scope_component *a,
   return a->offset < b->offset;
 }
 
-static struct scope_component *scope_components;
-static int curr_scope_id = 0;
+static std::vector<struct scope_component> scope_components;
+static std::map<Module *, int> module_scopes;
 
 char *get_scope_attr_name(scope_id_t scope_id) {
```
sangjinhan: This function seems not used anywhere (and the one below)

core/metadata.cc
```diff
@@ -29,8 +30,8 @@ bool scope_component_less(const struct scope_component *a,
   return a->offset < b->offset;
 }
 
-static struct scope_component *scope_components;
-static int curr_scope_id = 0;
+static std::vector<struct scope_component> scope_components;
+static std::map<Module *, int> module_scopes;
```
sangjinhan: int should be scope_id_t...? (not sure)

core/metadata.cc
```diff
@@ -43,25 +44,18 @@ int get_scope_attr_(scope_id_t scope_id) {
...
-    component->num_modules = 1;
-    component->modules = (Module **)mem_alloc(sizeof(Module *));
-    component->modules[0] = m;
-  } else {
-    component->num_modules++;
-    component->modules = (Module **)mem_realloc(
-        component->modules, sizeof(Module *) * component->num_modules);
-    component->modules[component->num_modules - 1] = m;
+  if (component.modules.size() == 0) {
+    strcpy(component.name, attr->name);
```
sangjinhan: let's use std::string in place of char*

core/modules/set_metadata.cc
```diff
@@ -53,9 +53,9 @@ class SetMetadata : public Module {
  private:
   struct snobj *AddAttrOne(struct snobj *attr);
 
-  int num_attrs_;
+  int num_attrs_ = 0;
 
-  struct Attr attrs_[MAX_ATTRS];
+  struct Attr attrs_[MAX_ATTRS] = {};
```
sangjinhan: maybe we can use std::vector? (without MAX_ATTRS limit)

sangjinhan:
Merged, thanks!

PR #65, opened 2016-10-21
Add Protobuf and gRPC

changlan:
Would like to send PR early in case this branch diverges too much.
Done:
* Add protobuf and gRPC
* Add services and messages
* Change port drivers

Remaining:
* Change modules
* Change master 
* Completely remove snobj


sangjinhan:
Merged, thanks! It seems that on some packages relying on an older version of libprotobuf seem to conflict with the new version in Ubuntu 16.04. Hm...

PR #76, opened 2016-10-25
gRPC for modules

changlan:
- Reorganize proto files
- Change BPF, Dump, ExactMatch, and FlowGen modules
- More modules to come..

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -3,10 +3,10 @@
 #include <rte_config.h>
 #include <rte_ether.h>
 
-#include "bessctl.grpc.pb.h"
 #include "message.h"
 #include "module.h"
 #include "port.h"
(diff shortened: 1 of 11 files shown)


core/bessctl.cc
```diff
@@ -937,6 +963,9 @@ class BESSControlImpl final : public BESSControl::Service {
   Status GetModuleInfo(ClientContext* context,
                        const GetModuleInfoRequest& request,
                        GetModuleInfoResponse* response) {
+    if (is_any_worker_running()) {
```
sangjinhan: get_module_info() does not require worker threads to be stopped (see sn_handlers() in snctl.cc)

changlan: Yup, fixed in the next patch!

PR #86, opened 2016-10-26
Adds benchmarking support and ports the oldtests/htable.cc benchmark …

barath:
…to use the new google benchmark routines.


sangjinhan:
Merged, thanks! There are some link errors on g++ 4.x, but I am reworking the build process in another branch so I'll fix the issue when I merge it.

PR #87, opened 2016-10-26
replace a bunch of C-style casts with STL casts

melvinw:
some weird arithmetic type conversion casts have been left alone for now


melvinw:
closing. going to resubmit as part of a more general style patch

PR #88, opened 2016-10-26
rename header guards

melvinw:


core/common.h
```diff
@@ -1,7 +1,7 @@
 /* This header file contains general (not BESS specific) C definitions */
 
-#ifndef _COMMON_H_
-#define _COMMON_H_
+#ifndef CORE_COMMON_H_
```
barath: Sorry for the nitpick -- you might prepend BESS_ to all of these, to match the header guard style from the style guide.

melvinw: fixed in f7157e9

melvinw:
going to resubmit as part of a more general style patch

PR #89, opened 2016-10-26
Reorganizes metadata routines; changes namespace.

barath:


sangjinhan:
Merged, thanks!

sangjinhan:
With the namespace now we can rename all the MT_* things to make the code less verbose, I think.

PR #95, opened 2016-10-27
Improves code coverage for bessd routines.

barath:
Adds script for generating code coverage HTML.

Moves driver initialization to PortBuilder.


sangjinhan:
Merged, thanks! (.gitignore needs to be updated)

PR #102, opened 2016-11-01
Create pcap wrapper class.

justinemarie:
Is this how we wanted it refactored? Then will work in gmock for tests.

### DIFF
--- a/core/drivers/pcap.cc
+++ b/core/drivers/pcap.cc
@@ -1,193 +1,48 @@
 #include "pcap.h"
+
+PCAPPort::~PCAPPort() {
+  if (pcap_handle_.IsInitialized()) {
+    DeInit();
+  }
+}
+
+// don't use
 struct snobj *PCAPPort::Init(struct snobj *conf) {
-  char errbuf[PCAP_ERRBUF_SIZE];
-  char *_dev = snobj_eval_str(conf, "dev");
-
-  if (!_dev) {
-    return snobj_err(EINVAL, "PCAP need to set dev option");
-  }
-
-  const std::string dev(_dev);
-
-  // non-blocking pcap
-  pcap_handle_ = pcap_open_live(dev.c_str(), PCAP_SNAPLEN, 1, -1, errbuf);
-  if (pcap_handle_ == nullptr) {
-    return snobj_err(ENODEV, "PCAP Open dev error: %s", errbuf);
-  }
-
-  int ret = pcap_setnonblock(pcap_handle_, 1, errbuf);
-  if (ret != 0) {
-    return snobj_err(ENODEV, "PCAP set to nonblock error: %s", errbuf);
-  }
-
-  LOG(INFO) << "PCAP: open dev " << dev;
-
   return nullptr;
 }
 
 pb_error_t PCAPPort::Init(const bess::protobuf::PCAPPortArg &arg) {
-  char errbuf[PCAP_ERRBUF_SIZE];
+  if (pcap_handle_.IsInitialized()) {
+    return pb_error(EINVAL, "Device already initialized.");
+  }
 
   const std::string dev = arg.dev();
-  if (dev.length() == 0) {
-    return pb_error(EINVAL, "PCAP need to set dev option");
-  }
-  // Non-blocking pcap.
-  pcap_handle_ = pcap_open_live(dev.c_str(), PCAP_SNAPLEN, 1, -1, errbuf);
-  if (pcap_handle_ == nullptr) {
-    return pb_error(ENODEV, "PCAP Open dev error: %s", errbuf);
-  }
+  pb_error_t err;
+  pcap_handle_ = PCAPHandle(dev, &err);
 
-  int ret = pcap_setnonblock(pcap_handle_, 1, errbuf);
-  if (ret != 0) {
-    return pb_error(ENODEV, "PCAP set to nonblock error: %s", errbuf);
-  }
-
-  LOG(INFO) << "PCAP: open dev " << dev;
-
-  return pb_errno(0);
+  return err;
 }
 
 void PCAPPort::DeInit() {
-  if (pcap_handle_) {
-    pcap_close(pcap_handle_);
-    pcap_handle_ = nullptr;
+  if (pcap_handle_.IsInitialized()) {
+    pcap_handle_.~PCAPHandle();
   }
 }
-
-#if 0
-static int pcap_rx_jumbo(struct rte_mempool *mb_pool,
-		struct rte_mbuf *mbuf,
-		const u_char *data,
-		uint16_t data_len)
-{
-	struct rte_mbuf *m = mbuf;
-
-	/* Copy the first segment. */
-	uint16_t len = rte_pktmbuf_tailroom(mbuf);
-
-	rte_memcpy(rte_pktmbuf_append(mbuf, len), data, len);
-	data_len -= len;
-	data += len;
-
-	while (data_len > 0) {
-		/* Allocate next mbuf and point to that. */
-		m->next = rte_pktmbuf_alloc(mb_pool);
-
-		if (unlikely(!m->next))
-			return -1;
-
-		m = m->next;
-
-		/* Headroom is needed for VPort TX */
-
-		m->pkt_len = 0;
-		m->data_len = 0;
-
-		/* Copy next segment. */
-		len = MIN(rte_pktmbuf_tailroom(m), data_len);
-		rte_memcpy(rte_pktmbuf_append(m, len), data, len);
-
-		mbuf->nb_segs++;
-		data_len -= len;
-		data += len;
-	}
-
-	return mbuf->nb_segs;
-}
-#endif
 
 int PCAPPort::RecvPackets(queue_t qid, snb_array_t pkts, int cnt) {
-  const u_char *packet;
-  struct pcap_pkthdr header;
-
-  int recv_cnt = 0;
-  struct snbuf *sbuf;
-
-  while (recv_cnt < cnt) {
-    packet = pcap_next(pcap_handle_, &header);
-    if (!packet) {
-      break;
-    }
-
-    sbuf = snb_alloc();
-    if (!sbuf) {
-      break;
-    }
-
-    if (header.caplen <= SNBUF_DATA) {
-      // pcap packet will fit in the mbuf, go ahead and copy.
-      rte_memcpy(rte_pktmbuf_append(&sbuf->mbuf, header.caplen), packet,
-                 header.caplen);
-    } else {
-/* FIXME: no support for chained mbuf for now */
-#if 0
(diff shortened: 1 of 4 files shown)


justinemarie:
Also I know nothing about C++ objects so uh, sorry if I b0rked that.

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
+#ifndef PCAP_HANDLE_H
```
barath: You might change this to:
#ifndef BESS_CORE_UTILS_PCAP_HANDLE_H_

(and the subsequent #define)

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+  // For chained mbufs, this function converts back to pcap packets with
+  // contiguous data.
+  void GatherData(unsigned char* tx_pcap_data, struct rte_mbuf* mbuf);
+
+  bool is_initialized_;
+
+  pcap_t* handle_;
+};
+
+#endif
```
barath: Just for good measure, you might change this line to say:
#endif  // BESS_CORE_UTILS_PCAP_HANDLE_H_

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <glog/logging.h>
+
+#include "../message.h"
+#include "../port.h"
+#include "../snobj.h"
+#include "../utils/pcap.h"
+
+// Wraps all accesses to libpcap; opens a device, sets it nonblocking, and then
+// lets you send/recv packets.
```
barath: I would move the comment text after the semicolon to the constructor / other methods.

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+class PCAPHandle {
+ public:
+  PCAPHandle(std::string dev, pb_error_t* error);
```
barath: Slight change:
PCAPHandle(const std::string &dev, pb_error_t *error);

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+  PCAPHandle();
+
+  ~PCAPHandle();
```
barath: virtual

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+
+  int SendPackets(snb_array_t pkts, int cnt);
```
barath: Add short method comment.

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,39 @@
...
+
+  int RecvPackets(snb_array_t pkts, int cnt);
+
+  bool IsInitialized();
```
barath: If this is just a raw getter method, rename to:
bool is_initialized() const;

core/utils/pcap_handle.cc
```diff
@@ -0,0 +1,133 @@
+#include "pcap_handle.h"
+
+PCAPHandle::PCAPHandle() {
+  is_initialized_ = 0;
+  handle_ = nullptr;
+}
```
barath: No need for this here -- can just use an initializer list for the constructor in the .h file.

core/drivers/pcap.cc
```diff
@@ -1,193 +1,48 @@
...
-  return pb_errno(0);
+  return err;
 }
 
 void PCAPPort::DeInit() {
-  if (pcap_handle_) {
-    pcap_close(pcap_handle_);
-    pcap_handle_ = nullptr;
+  if (pcap_handle_.IsInitialized()) {
+    pcap_handle_.~PCAPHandle();
```
barath: You generally don't want to explicitly call a destructor like this.  Instead, you might make pcap_handle_ a pointer, call new above when constructing it, and then call delete here to destruct it.

core/drivers/pcap.cc
```diff
@@ -1,193 +1,48 @@
...
-  if (ret != 0) {
-    return pb_error(ENODEV, "PCAP set to nonblock error: %s", errbuf);
+  if (pcap_handle_.IsInitialized()) {
+    return pb_error(EINVAL, "Device already initialized.");
   }
 
-  LOG(INFO) << "PCAP: open dev " << dev;
+  const std::string dev = arg.dev();
+  pb_error_t err;
+  pcap_handle_ = PCAPHandle(dev, &err);
```
barath: Use new instead of calling the constructor and copying.

justinemarie:
I was under the impression that Sangjin had explicitly wanted us to avoid
virtual and new (don't want to follow the pointer; instead want it declared
inline in the object).

Sangjin?

barath:
The destructor being virtual will not cause any problems.

Good point about not having a pointer to the pcap handle class.  A few options:
1) Construct the Handle in the PCAPPort constructor (which would change behavior slightly, by moving the initialization to the constructor from Init)
2) Just use a pointer and expose the underlying pcaphandle via a getter
3) Implement a move assignment operator for PCAPHandle so the ownership of the values is cleanly moved when you copy the value (i.e. it zeros out the handle for the original owner, etc.) which you then invoke with std::move() when setting the local member variable of PCAPHandle; also you can implement a Reset() method for PCAPHandle so that it can be reset without being destructed; you can then just call Reset() from the destructor.

I think 3 is probably the right approach here.

sangjinhan:
Yeah, in general we want to avoid virtual functions as much as possible for any performance-critical code. But the destructor here is not on the critical path, since the code will be called in the master thread, not on a worker. Also, keep in mind that in C++, any destructors must be virtual in virtually (pun intended) all cases.

barath:
I should have added, in addition to the move constructor, you'll want to include DISALLOW_COPY_AND_ASSIGN in pcap_handle.h

sangjinhan:
I have no comments to add, except a minor nitpick: PCAPHandle should be renamed PcapHandle, as per Google's C++ style guide. PCAPPort was some sort of exception, because the name is visible to the external world (thus requiring backward compatibility). Please reflect Barath comments to the code.

Slightly off-the-topic: maybe we can do this for the next PR;  all files in /utils directory are supposed to be independent of the rest of BESS code (common.h is perhaps the only exception). As a rule of thumb, all library classes/functions in utils must be useful and readily available on its own, for any other projects than BESS. Currently PCAPHandle has dependency on the BESS core (snbuf, module, as well as external libraries (glog, gtest, protobuf, etc.) beyond libpcap.

justinemarie:
I don't understand this:
>>Implement a move assignment operator for PCAPHandle so the ownership of
the values is cleanly moved when you copy the value (i.e. it zeros out the
handle for the original owner, etc.) which you then invoke with std::move()
when setting the local member variable of PCAPHandle;

I can do this:
>> also you can implement a Reset() method for PCAPHandle so that it can be
reset without being destructed; you can then just call Reset() from the
destructor.

justinemarie:
>
> Slightly off-the-topic: maybe we can do this for the next PR; all files in
> /utils directory are supposed to be independent of the rest of BESS code
> (common.h is perhaps the only exception). As a rule of thumb, all library
> classes/functions in utils must be useful and readily available on its own,
> for any other projects than BESS. Currently PCAPHandle has dependency on
> the BESS core (snbuf, module, as well as external libraries (glog, gtest,
> protobuf, etc.) beyond libpcap.
>
-- Is common.h where I should put a function like,
"convert_mbuf_to_char_array()"?
(Without that function I can't pull out the dependency on snbuf).

-- Should I remove the logging and error code altogether?

Justine

barath:
TLDR: you want to maintain the invariant that only one instance of an object holds a handle and that's what specifying move assignment will let you do.

The basic idea with "move" in C++11 is to capture cases where you want to explicitly say that the contents of instance foo are not just to be copied to bar but bar should take over the internal state of foo.  A good example of this is std::unique_ptr, where you want to make sure that an underlying pointer is only held by one instance of unique_ptr, but you want to be able to transfer that underlying pointer safely from one unique_ptr instance to another.  The problem is that every class will have its own way of achieving this, so they introduced a special construction "move" to achieve this.  It's used in two cases -- copy construction and assignment.  Here you're doing an assignment:

pcap_handle_ = PCAPHandle(dev, &err);

What's effectively happening is the constructor for PCAPHandle is getting called, a handle is initialized (inside the constructor), a temporary PCAPHandle object is created and then the implicitly-defined copy constructor (of PCAPHandle for pcap_handle_) is getting invoked to copy the temporary to pcap_handle_.  The implicit copy constructor just does a field-by-field copy.

This can cause a few problems:
1) The temporary is temporary, so its destructor will immediately get called.  That means that if the destructor for PCAPHandle frees the underlying handle, then pcap_handle_ will be holding an invalid handle (that was just copied to it by the implicit copy constructor).
2) If the right hand side were another PCAPHandle variable instead of a temporary, you'd have two PCAPHandles that own the same underlying pcap handle, which isn't safe.
3) it's inefficient in general, though perhaps no worse in this context.

This is why the style guide suggests eliminating implicit copy and assignment via a macro like DISALLOW_COPY_AND_ASSIGN() (in the private section of a class) -- you get rid of weird implicit unsafe / slow behavior that can happen without warning.

However in this case we want to disallow implicit copy / assignment but want to explicitly allow a move operation that we control the behavior of.  There are two types of moves we can define -- a move constructor and move assignment (due to somewhat arcane rules they will not be automatically provided by the compiler in this context).

Here because pcap_handle_ has already been constructed, you want to define a move assignment operator.  You won't need to use std::move() to force it to use your defined move code because the right hand side is a temporary, but if it were a variable then the assignment would try to use the copy assignment operator (which wouldn't exist) and it wouldn't compile; std::move() tells it to explicitly try to move in such cases.

So, wrapping up, now with a move assignment operator implemented properly, the left hand side will have sole ownership of the internal handle by the time the assignment completes and the handle held by the temporary will have been cleared, so when the temporary's destructor is called it will no longer free anything.

sangjinhan:
Note that common.h does not have any dependency either, other than standard C/C++ libraries. It does not assume anything about BESS (or its dependencies, such as DPDK). Perhaps I will move common.h to /utils directory at some point...

Regarding the mbuf <-> pcap conversion, perhaps the cleanest way for PCAPPort is to give a char * pointer to PcapHandle.

barath:
Just checked the style guide -- they suggest that if you implement a move assignment operator you should also implement a move constructor (which will behave quite similarly).

justinemarie:
It feels like that's a pretty common thing that one would want to do -- why
bake it in to PCAPPort rather than put it somewhere generally accessible?

sangjinhan:
Since rte_mbuf is DPDK-specific, so having it in common.h or utils/ will introduce a dependency to DPDK. Having it in the core code (e.g., snbuf) should be fine. Please check if snb_head_data() already does what you want.

justinemarie:
snbuf seems great, I'll look there. Thanks.

PR #103, opened 2016-11-01
rework metadata

melvinw:


core/metadata.cc
```diff
@@ -260,123 +290,75 @@ static void FillOffsetArrays() {
...
-static void AssignOffsets() {
+void Pipeline::AssignOffsets() {
   mt_offset_t offset = 0;
-  std::priority_queue<struct scope_component *,
-                      std::vector<struct scope_component *>,
-                      std::function<bool(const struct scope_component *,
-                                         const struct scope_component *)>>
+  std::priority_queue<
+      ScopeComponent *, std::vector<ScopeComponent *>,
+      std::function<bool(const ScopeComponent *, const ScopeComponent *)>>
```
barath: Minor, and perhaps unnecessary -- you might change this to the more canonical C++ stl style, which is a class that has an operator method to do the comparison.

barath: Just to clarify -- it wouldn't be an overload in the class but rather a binary function implemented as a class with its operator defined.

melvinw: Gotcha

core/metadata.h
```diff
@@ -3,73 +3,26 @@
...
-static const int kMetadataAttrMaxSize = 32; // In bytes, per attribute.
+static const int kMetadataAttrMaxSize = 32;  // In bytes, per attribute.
 
-static const int kMaxAttrsPerModule = 16; // Max number of attributes per module.
+static const int kMaxAttrsPerModule =
+    16;  // Max number of attributes per module.
 
-static const int kMetadataTotalSize = 96; // Total size of all metadata in bytes.
+static const int kMetadataTotalSize =
+    96;  // Total size of all metadata in bytes.
```
sangjinhan: This should be derived (see snbuf.h and snbuf_layout.h)

core/metadata.h
```diff
@@ -88,12 +41,7 @@ static inline int IsValidOffset(mt_offset_t offset) {
   return (offset >= 0);
 }
 
-
-enum mt_access_mode {
-  MT_READ = 0,
-  MT_WRITE,
-  MT_UPDATE
-};
+enum mt_access_mode { MT_READ = 0, MT_WRITE, MT_UPDATE };
```
sangjinhan: The type name should be changed..? (not sure)

With "enum class", we can remove the ugly MT_ prefix

core/metadata.h
sangjinhan: should be size_t..?

core/metadata.h
```diff
@@ -3,73 +3,26 @@
...
-static const int kMaxAttrsPerModule = 16; // Max number of attributes per module.
+static const int kMaxAttrsPerModule =
+    16;  // Max number of attributes per module.
 
-static const int kMetadataTotalSize = 96; // Total size of all metadata in bytes.
+static const int kMetadataTotalSize =
+    96;  // Total size of all metadata in bytes.
 
 // Normal offset values are 0 or a positive value.
 typedef int8_t mt_offset_t;
```
sangjinhan: What is the non-normal offset value here, and when is it used? (can we use unsigned?)

sangjinhan: it would be nice to check if kMetadataTotalSize <= intXX_max

core/metadata.h
```diff
@@ -102,10 +50,114 @@ struct mt_attr {
...
+  int degree_;
+};
 
-char *GetScopeAttrName(scope_id_t scope_id);
-int *GetScopeAttrSize(scope_id_t scope_id);
+class Pipeline {
+ public:
+  Pipeline() : scope_components_(), module_scopes_(){};
+
+  ~Pipeline(){};
```
sangjinhan: Should be virtual. Perhaps we can remove this line and just rely on the default destructor, which is virtual?

core/metadata.cc
```diff
@@ -15,94 +13,156 @@
...
+// Helpers -----------------------------------------------------------------
+
+static mt_offset_t ComputeNextOffset(mt_offset_t curr_offset, int8_t size) {
+  uint32_t overflow;
+  int8_t rounded_size;
+
+  rounded_size = align_ceil_pow2(size);
+
+  if (curr_offset % rounded_size != 0) {
+    curr_offset = ((curr_offset / rounded_size) + 1) * rounded_size;
```
sangjinhan: You can use align_ceil(curr_offset, rounded_size)

core/metadata.h
```diff
@@ -119,6 +119,11 @@ class Pipeline {
   // Main entry point for calculating metadata offsets.
   int ComputeMetadataOffsets();
 
+  // Registers attr and returns 0 if no attribute named attr->name with size
+  // other than attr->size has already been registered for this pipeline.
+  // Returns -errno on error.
+  int RegisterAttribute(const struct mt_attr *attr);
+
```
melvinw: DeRegisterAttribute() would be nice to have, but I'm not sure there's a clean way to do that without allowing any module to delete previously registered attributes

melvinw:
comments addressed

sangjinhan:
Merged, thanks!

PR #106, opened 2016-11-02
add metadata unit tests

melvinw:


melvinw:
Need to add some comments to the tests before merging. Submitting to see what the coverage robot says.

core/metadata_test.cc
```diff
@@ -0,0 +1,198 @@
...
+  ASSERT_EQ(0, default_pipeline.ComputeMetadataOffsets());
+
+  for (size_t i = 0; i < m0->num_attrs; i++) {
+    if (m0->attrs[i].mode == bess::metadata::AccessMode::READ)
+      continue;
+
+    bess::metadata::mt_offset_t offset = m0->attr_offsets[i];
+    for (mt_offset_t j = 0; j < m0->attrs[i].size; j++) {
+      ASSERT_FALSE(dummy_meta[offset + j]);
+      dummy_meta[offset + j] = true;
```
barath: Maybe I'm not totally getting the aim of this test -- I'm curious whether you want to be checking the actual offset results after having computed them above.

core/metadata_test.cc
```diff
@@ -0,0 +1,198 @@
...
+  ASSERT_NE(mods[1]->attr_offsets[0], mods[4]->attr_offsets[0]);
+  ASSERT_EQ(mods[0]->attr_offsets[0], mods[4]->attr_offsets[0]);
+  ASSERT_EQ(mods[3]->attr_offsets[0], mods[4]->attr_offsets[0]);
+  ASSERT_EQ(mods[5]->attr_offsets[0], mods[6]->attr_offsets[0]);
+  ASSERT_NE(mods[5]->attr_offsets[0], mods[6]->attr_offsets[1]);
+  ASSERT_EQ(mods[7]->attr_offsets[0], mods[6]->attr_offsets[0]);
+  ASSERT_NE(mods[7]->attr_offsets[0], mods[6]->attr_offsets[1]);
+  ASSERT_NE(mods[7]->attr_offsets[0], mods[8]->attr_offsets[0]);
+  ASSERT_EQ(mods[7]->attr_offsets[0], mods[9]->attr_offsets[1]);
+  ASSERT_EQ(mods[8]->attr_offsets[0], mods[9]->attr_offsets[0]);
```
melvinw: @barath re: https://github.com/NetSys/bess/pull/106/files/4990a9c7e7a1bc1e6820b69ce88c76895564db86#r86116566
or did you mean that it would be sufficient to just do something like this?

barath:
Yes, sort of the latter -- I guess I was trying to understand why you do
explicit checks only in the last test case.  But what you said makes sense.

melvinw:
Gotcha. The explicit checks definitely ought to be in the other test too. Will add

melvinw:
fixed up. ready for review

core/metadata_test.cc
```diff
@@ -0,0 +1,220 @@
...
+
+#include <gtest/gtest.h>
+
+#include <vector>
+
+// Mocking out misc things  ------------------------------------------------
+
+// -------------------------------------------------------------------------
+
+class Foo : public Module {
```
sangjinhan: I think the mock code should be not exposed to the outside world, to avoid namespace pollution.  The Google C++ style guide says all code (except #include) in *.cc should be in a namespace. Please use an unnamed namespace here.

melvinw: fixed

sangjinhan:
Merged, thanks!

PR #107, opened 2016-11-02
PCAP Wrapper Class with changes discussed previously.

justinemarie:


core/utils/pcap_handle.cc
```diff
@@ -0,0 +1,75 @@
...
+PcapHandle::PcapHandle(PcapHandle&& other) {
+  handle_ = other.handle_;
+  is_initialized_ = other.is_initialized_;
+
+  other.handle_ = nullptr;
+  other.is_initialized_ = false;
+}
+
+PcapHandle& PcapHandle::operator=(PcapHandle&& other) {
+  if (this != &other && is_initialized_) {
```
barath: I think this if statement should be broken up as follows:

if (this != &other) {
  if (is_initialized_) {
    Reset();
  }
  ...
}
return *this;

Otherwise, if this is equal to other, then down below it will get nulled out, which I don't think you want to happen.

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,52 @@
...
+#include "../common.h"
+#include "pcap.h"
+#include <pcap/pcap.h>
+#include <string>
+
+// Wraps all accesses to libpcap
+class PcapHandle {
+ public:
+  // Opens a device and sets it to nonblocking
+  PcapHandle(std::string dev);
```
barath: const std::string &dev

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,52 @@
...
+  const u_char *RecvPacket(int *caplen);
+
+  // Is false if there's no pcap binding established
+  bool is_initialized() const;
+
+ private:
+  // Only one instance of PcapHandle should exist for a given handle at a time.
+  DISALLOW_COPY_AND_ASSIGN(PcapHandle);
+
+  bool is_initialized_;
```
barath: Just wondering -- is it ever possible that is_initialized is true but handle_ is nullptr, or vice versa?  If not, then you don't need the is_initialized_ member -- whether handle_ is nullptr or not determines whether it's initialized.

justinemarie:
Try again?

core/utils/pcap_handle.cc
```diff
@@ -0,0 +1,67 @@
...
+
+void PcapHandle::Reset() {
+  if (is_initialized()) {
+    pcap_close(handle_);
+    handle_ = nullptr;
+  }
+}
+
+bool PcapHandle::is_initialized() const {
+  return (handle_ == nullptr);
```
barath: Should this be handle_ != nullptr?

core/utils/pcap_handle.h
```diff
@@ -0,0 +1,50 @@
...
+#include "../common.h"
+#include "pcap.h"
+#include <pcap/pcap.h>
+#include <string>
+
+// Wraps all accesses to libpcap
+class PcapHandle {
+ public:
+  // Opens a device and sets it to nonblocking
+  PcapHandle(const std::string dev);
```
barath: Should be a const reference rather than passing by value.

justinemarie:
Third time's a charm! :-)

sangjinhan:
Merged, thanks!

PR #108, opened 2016-11-02
Changes bessd_test to ensure that fork() behavior is self-contained, …

barath:
…helping to ensure that tests that fail propagate that error to the top level.

### DIFF
--- a/core/bessd_test.cc
+++ b/core/bessd_test.cc
@@ -2,6 +2,8 @@
 
 #include <signal.h>
 #include <sys/file.h>
+#include <sys/select.h>
+#include <unistd.h>
 
 #include <glog/logging.h>
 #include <gtest/gtest.h>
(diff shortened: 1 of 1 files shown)


core/bessd_test.cc
```diff
@@ -116,11 +117,12 @@ TEST(TryAcquirePidfileLock, AlreadyHeld) {
     ASSERT_EQ(0, flock(fd, LOCK_EX | LOCK_NB)) << "Couldn't acquire file lock for test.";
 
     // Sleep for up to 10 seconds, while holding lock.
-    sleep(10);
+    sleep(3);
```
sangjinhan: Would it be possible to remove this arbitrary (and potentially unreliable) delay, e.g., with some IPC? A few second delay makes me a little hesitant to run the ./all_test script frequently.

barath: I was thinking about that.  Sure, I can see what will work best.

barath:
I changed the behavior of bessd_test so it now uses pipes instead of sleeps to coordinate between child and parent processes in tests.  (Unfortunately DEATH tests didn't appear to work properly when invoked inside of lambdas, so I ended up not being able to use a C++11-style solution.)

sangjinhan:
Merged, thanks!

PR #109, opened 2016-11-03
pmd: Add ability to query link-status

chaitanyalala:
RFC patch

Currently there is no way to query for link
status like speed/duplex/link up/down for a physical
NIC controlled via bess. This ability is crucial to
for control plane protocols who need to take action if
link state changes.

This patch adds a ability to to access this setting via
a DPDK api and then adds plumbing all the way upto python
layer.

// Sample output
Current Link State for Physical port is {'duplex': 1, 'status': 1,
'autoneg': 0, 'speed': 10000}
// End

Signed-off-by: Chaitanya Lala <clala@arista.com>

### DIFF
--- a/bessctl/conf/samples/vport.bess
+++ b/bessctl/conf/samples/vport.bess
@@ -6,3 +6,8 @@ p = PMDPort(port_id=0)
 
 PortInc(port=p) -> PortOut(port=v)
 PortInc(port=v) -> PortOut(port=p)
+
+link_state = bess.get_port_link(p.name)
+# http://dpdk.org/doc/api/rte__ethdev_8h_source.html
+# Values for link_state
+print "Current Link State for Physical port is %s" % link_state
--- a/bessctl/port.py
+++ b/bessctl/port.py
@@ -23,3 +23,5 @@ class Port(object):
 
     def get_port_stats(self):
         return self.bess.get_port_stats(self.name)
+    def get_port_link(self):
+        return self.bess.get_port_link(self.name)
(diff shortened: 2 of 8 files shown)


sangjinhan:
One clarification question: what does autoneg exactly mean in the struct? At a first glance it looks like "whether we enabled the auto negotiation feature" (so, **configuration**). If so, the nature of this field is somewhat different from others -- link status, duplex, speed -- since they are more like **status**, not configuration.

If its value never changes unless we do so, shall we remove the autoneg field from the struct?

chaitanyalala:
What it means is "did the port achieve current speed/duplex/link by auto
negotiating or by fix setting". It is always a part of these things. I
tried to keep the same structure as is used by dpdk. Even standard Linux
ethtool displays. IMHO it should not be removed.

chaitanyalala:
To illustrate my point, please look at what standard Linux ethtool report...

# ethtool eth0
Settings for eth0:
        Supported ports: [ TP ]
        Supported link modes:   10baseT/Half 10baseT/Full
                                100baseT/Half 100baseT/Full
                                1000baseT/Full
        Supports auto-negotiation: Yes
        Advertised link modes:  10baseT/Half 10baseT/Full
                                100baseT/Half 100baseT/Full
                                1000baseT/Full
        Advertised auto-negotiation: Yes <------- This is what the autoneg bit in struct means.
        Speed: 100Mb/s
        Duplex: Full
        Port: Twisted Pair
        PHYAD: 1
        Transceiver: internal
        Auto-negotiation: on
        Supports Wake-on: d
        Wake-on: d
        Link detected: yes

sangjinhan:
Oh I see, thanks for the comment. I will make sure equivalent functionality exist in the upcoming c++ version of BESS as well.

PR #110, opened 2016-11-03
Add basic unit tests and add initialization checks for PcapHandle.

justinemarie:


core/utils/pcap_handle.cc
```diff
@@ -52,14 +52,18 @@ bool PcapHandle::is_initialized() const {
 }
 
 int PcapHandle::SendPacket(const u_char* packet, int len) {
-  if (len <= PCAP_SNAPLEN) {
+  if (is_initialized() && len <= PCAP_SNAPLEN) {
```
barath: Minor question -- I think this is fine here, but perhaps in the higher performance drivers we should expect the caller to check if it's initialized before making any send/receive calls, so that this check doesn't have to happen per packet.

core/utils/pcap_handle_test.cc
```diff
@@ -0,0 +1,67 @@
...
+
+// Check that it knows it's initialized if it has a pcap handle.
+TEST_F(PcapHandleFixtureTest, IsInitialized) {
+  ASSERT_TRUE(pcap_with_fake_handle.is_initialized());
+}
+
+// Move operator.
+TEST_F(PcapHandleFixtureTest, MoveOperator) {
+  ASSERT_TRUE(pcap_with_fake_handle.is_initialized());
+  PcapHandle MoveTo = std::move(pcap_with_fake_handle);
```
barath: Unless I'm reading this wrong, this will actually invoke the move constructor, not the move assignment operator because the object is being created on this line.  However if you were to separate the lines as:

PcapHandle MoveTo();
MoveTo = std::move(pcap_with_fake_handle);

then it will invoke the move assignment operator.

The way I think about it is this:
Does the object on the left already exist? Yes: it will use an assignment operator, No: a constructor.
Does the object on the right already exist? Yes: it will not move unless told to explicitly, No: it will try to use a move if possible

core/utils/pcap_handle_test.cc
```diff
@@ -0,0 +1,67 @@
...
+  pcap_with_fake_handle = std::move(MoveTo);
+  ASSERT_FALSE(MoveTo.is_initialized());
+}
+
+// Test the move constructor.
+TEST_F(PcapHandleFixtureTest, MoveConstructor) {
+  PcapHandle MoveTo(std::move(pcap_with_fake_handle));
+  ASSERT_FALSE(pcap_with_fake_handle.is_initialized());
+  ASSERT_TRUE(MoveTo.is_initialized());
+  pcap_with_fake_handle = std::move(MoveTo);
```
barath: I believe this will invoke the move assignment operator.  (Which is fine -- just thought I'd point that out in case that wasn't intended.)

sangjinhan:
Merged, thanks! Added an additional commit (dacf62c) to address minor issues, including Barath's comment.

PR #111, opened 2016-11-03
add some module tests

melvinw:
+ add example test for modules going forward
+ add metadata getter/setter template tests


melvinw:
waiting for bessctl to converge before merging this.

PR #112, opened 2016-11-04
Protobuf polymorphism

changlan:
- Minor naming changes in proto files
- Use google.protobuf.Any types for arguments of module init function and commands 
- Remove BESS core dependency on module-specific proto files


justinemarie:
Whoaaaaa that's a lot of files!

sangjinhan:
Merged, thanks!

PR #114, opened 2016-11-07
Replace old log_*() calls with Google glog

sangjinhan:
(along with other style fixes)


core/utils/format.cc
```diff
@@ -0,0 +1,34 @@
...
+  va_end(ap);
+  return s;
+}
+
+std::string FormatVarg(const char *fmt, va_list ap) {
+  const int init_bufsize = 128;
+  std::unique_ptr<char[]> buf;
+
+  buf.reset(new char[init_bufsize]);
+  int ret = vsnprintf(&buf[0], init_bufsize, fmt, ap);
```
barath: Just curious -- why not vasprintf()?  Should simplify the buffer sizing aspect.

barath: (It's true that vasprintf() is a GNU/Linux/BSD function, not standard C, so that could be a reason not to use it.)

sangjinhan: I had no particular reason (just copy-and-pasted Chang's one from message.h/cc). I updated the code with vasprintf().

melvinw:
Looks good. +1

PR #115, opened 2016-11-07
reintroduce code coverage script

melvinw:
It's useful for checking the effects of new tests before submitting pull
requests.


sangjinhan:
- .gitignore needs to include the generated file and directory.
- Maybe you can run all_test, instead of iterating over all individual files?

sangjinhan:
Thanks, merged with some additional minor changes.

melvinw:
was 027f31c supposed to get chopped off? It adds some convenience stuff

PR #116, opened 2016-11-07
Explicit types for init and cmd args

changlan:
Previously we used google::protobuf::Any as the argument type for module Init functions and Command functions. Inside those functions we convert google::protobuf::Any into appropriate message types using clumsy code snippets in every function:

```
pb_error_t PMDPort::Init(const google::protobuf::Any &arg_) {
  bess::pb::PMDPortArg arg;		
  arg_.UnpackTo(&arg);
  ...
```

This PR enables those functions to use module specific argument type, while keeping modules message types separate from BESS core.


changlan:
Working on the same mechanism for ports.

sangjinhan:
Let me know whenever it's ready

changlan:
Yep, it is finished.

sangjinhan:
Thanks!

PR #117, opened 2016-11-07
Do not create .make.dep file

sangjinhan:
.make.dep contains Makefile dependency information for all source code files. Whenever a single file is  modified, the entire .make.dep is updated. This patch introduces individual dependency files for each source code file in .deps/, without having the central .make.dep file.

Other changes:
- Makefile does not automatically enforce parallel build any longer. You have to explicitly specify `make -j` if parallel build is desired.
- A bug that does not show message with `VERBOSE=1 make` was fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ core/*.gcda
 core/*.gcno
 core/*/*.gcda
 core/*/*.gcno
+coverage.info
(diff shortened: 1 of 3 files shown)


sangjinhan:
Something went wrong. Making a new PR.

PR #123, opened 2016-11-08
Use -Og flag when built for coverage test

sangjinhan:
`-Ofast` option (the default) may show surprising coverage results due to aggressive optimization. Instead, we use `-Og` when the program is built with `COVERAGE=1`.

As a side effect of -Og, there are warnings and test failures that do not appear with `-Ofast`. Some commits fix these issues as well.


core/utils/time_test.cc
```diff
@@ -8,6 +8,12 @@ TEST(RdtscTest, NonDecreasing) {
   ASSERT_LE(a, b) << "Time stamp counter should not decrease.";
 }
 
+TEST(TscToUs, Frequency) {
+  ASSERT_NE(0, tsc_hz) << "tsc_hz has not been initialized";
```
barath: Will tsc_hz be 0 if TscHzSetter() hasn't been called?  Or will it be some undefined value?

barath: Ok, good, just checking.

PR #128, opened 2016-11-11
Introduce gate hooks

melvinw:
It would be helpful to have a mechanism by which developers can debug pipelines of modules. Currently this is accomplished with #defines and by conditionally tacking extra fields on the end of `struct gate`. This patch replaces the existing process with a more general mechanism, gate hooks.


core/bessctl.cc
```diff
@@ -53,12 +54,12 @@ static inline Status return_with_errno(T* response, int code) {
 }
 
 static int collect_igates(Module* m, GetModuleInfoResponse* response) {
-  for (int i = 0; i < m->igates.curr_size; i++) {
-    if (!is_active_gate(&m->igates, i))
+  for (size_t i = 0; i < m->igates.size(); i++) {
```
barath: C++11 style iterator?

core/module.h
```diff
@@ -64,6 +61,45 @@ static inline module_init_func_t MODULE_INIT_FUNC(
...
+  virtual void ProcessBatch(const struct pkt_batch *){};
+
+ private:
+  DISALLOW_COPY_AND_ASSIGN(GateHook);
+
+  const std::string &name_;
+
+  const uint16_t priority_;
+
+ protected:
```
barath: Does protected go above private?

core/module.h
```diff
@@ -64,6 +61,45 @@ static inline module_init_func_t MODULE_INIT_FUNC(
...
+  struct gate *gate_;
+};
+
+inline bool GateHookComp(const GateHook *lhs, const GateHook *rhs) {
+  return (lhs->priority() < rhs->priority());
+}
+
+const std::vector<GateHook *> kNoHooks = {};
```
barath: Might add a comment explaining this.

core/metadata.cc
```diff
@@ -261,8 +263,8 @@ void Pipeline::IdentifyScopeComponent(Module *m, const struct Attribute *attr) {
   /* cycle detection */
   module_scopes_[m] = static_cast<int>(scope_components_.size());
 
-  for (int i = 0; i < m->ogates.curr_size; i++) {
-    ogate = m->ogates.arr[i];
+  for (size_t i = 0; i < m->ogates.size(); i++) {
```
barath: C++11 style iteration?

core/metadata.cc
```diff
@@ -226,8 +227,8 @@ int Pipeline::TraverseDownstream(Module *m, const struct Attribute *attr) {
     return -1;
   }
 
-  for (int i = 0; i < m->ogates.curr_size; i++) {
-    ogate = m->ogates.arr[i];
+  for (size_t i = 0; i < m->ogates.size(); i++) {
```
barath: C++11 style iteration?

core/metadata.cc
```diff
@@ -207,8 +208,8 @@ int Pipeline::TraverseDownstream(Module *m, const struct Attribute *attr) {
     AddModuleToComponent(m, found_attr);
     found_attr->scope_id = scope_components_.size();
 
-    for (int i = 0; i < m->ogates.curr_size; i++) {
-      ogate = m->ogates.arr[i];
+    for (size_t i = 0; i < m->ogates.size(); i++) {
```
barath: C++11 style iteration.

core/metadata.cc
```diff
@@ -175,16 +176,16 @@ void Pipeline::TraverseUpstream(Module *m, const struct Attribute *attr) {
   }
   module_scopes_[m] = static_cast<int>(scope_components_.size());
 
-  for (int i = 0; i < m->igates.curr_size; i++) {
-    struct gate *g = m->igates.arr[i];
+  for (size_t i = 0; i < m->igates.size(); i++) {
```
barath: C++11 style iteration.

core/module.cc
```diff
@@ -562,55 +542,38 @@ int Module::EnableTcpDump(const char *fifo, gate_idx_t ogate) {
...
-    if (ret < 0) {
-      if (errno == EPIPE) {
-        DLOG(WARNING) << "Broken pipe: stopping tcpdump";
-        gate->tcpdump = 0;
-        gate->fifo_fd = 0;
-        close(fd);
-      }
-      return;
+  struct gate *gate = ogates[ogate];
+  for (size_t i = 0; i < gate->hooks.size(); i++) {
```
barath: Use auto i = gate->hooks.begin() instead of index to iterate?

core/module.h
melvinw: Not sure. Will check

sangjinhan: DISALLOW_COPY_AND_ASSIGN should go to the end of the private section

bessctl/commands.py
```diff
@@ -194,6 +194,19 @@ def get_var_attrs(cli, var_token, partial_word):
             var_desc = 'configuration filename'
             var_candidates = complete_filename(partial_word)
 
+        if var_token == '[DIRECTION]':
+            var_type = 'name'
+            var_desc = 'gate direction discriminator (default "out")'
+            var_candidates = ['in', 'out']
+
+        if var_token == '[PRIORITY]':
```
sangjinhan: It might be conceptually simpler each hook type has its own pre-defined priority value, rather than user specifies it for each hook individually. One possible (common?) way to do this is:

- Suppose there are three hook types: x, y, z (x has the highest priority, z is the lowest)
- Module M has those hooks enabled for both its input gate and output gate
- Then the order of execution should be: x - y - z - M - z - y - x
  - It is something like destructors are called in the reverse order of constructors for a derived class in C++.

bessctl/commands.py
```diff
@@ -1334,6 +1358,33 @@ def tcpdump_module(cli, module_name, ogate, opts):
         except:
             pass
 
+@cmd('track MODULE [DIRECTION] [GATE] [PRIORITY]',
```
melvinw: Ah, this is cleaner than what I had in mind for the global enable/disable case. Will do.

melvinw: I've been assuming gate 0 only, but all output gates does seem more intuitive/consistent

sangjinhan: Let's not bother with tcpdump's default gate number for this PR.

melvinw:
Pushed a bunch of changes, but I'm not quite done. Still need to properly order hooks on igates/ogates.

melvinw:
If you merge #130 before this, I can refactor gates as part of this patch

sangjinhan:
Yup, merged #130 first. Please remove the label whenever it's ready

melvinw:
something went terribly wrong. opening a new one

PR #129, opened 2016-11-12
Remove unused protobuf message types

changlan:


core/modules/ip_encap.cc
```diff
@@ -27,7 +27,7 @@ struct snobj *IPEncap::Init(struct snobj *arg [[maybe_unused]]) {
   return nullptr;
 };
 
-pb_error_t IPEncap::InitPb(const bess::pb::IPEncapArg &arg [[maybe_unused]]) {
+pb_error_t IPEncap::InitPb(const bess::pb::EmptyArg &) {
```
changlan: Not sure if I get it right, but if we add IPEncapArg, then the message fields that Init() expects to get becomes different. EmptyArg does not contain those fields so Init() will not be able to accept an EmptyArg message.

sangjinhan: Oh I was assuming the case where the new IPEncapArg does not have any "required" fields.

changlan:
Found other identical message types. Will resubmit.

PR #130, opened 2016-11-12
Run modules without recursion for better testability.

justinemarie:
I also ran clang-format resulting in some lines-that-look-edited that really weren't.

### DIFF
--- a/core/gate.h
+++ b/core/gate.h
@@ -1,5 +1,5 @@
-#ifndef BESS_CORE_GATE_H
-#define BESS_CORE_GATE_H
+#ifndef BESS_GATE_H_
+#define BESS_GATE_H_
 
 #include "utils/cdlist.h"
 
--- a/core/gate.h
+++ b/core/gate.h
@@ -1,0 +1,62 @@
+#ifndef BESS_CORE_GATE_H
+#define BESS_CORE_GATE_H
+
+#include "utils/cdlist.h"
+
+class Module;
+
+typedef uint16_t gate_idx_t;
+
+#define TRACK_GATES 1
+#define TCPDUMP_GATES 1
+
+#define INVALID_GATE UINT16_MAX
+
+/* A module may have up to MAX_GATES input/output gates (separately). */
+#define MAX_GATES 8192
+#define DROP_GATE MAX_GATES
+static_assert(MAX_GATES < INVALID_GATE, "invalid macro value");
+static_assert(DROP_GATE <= MAX_GATES, "invalid macro value");
+
+struct gate {
+  /* immutable values */
+  Module *m;           /* the module this gate belongs to */
+  gate_idx_t gate_idx; /* input/output gate index of itself */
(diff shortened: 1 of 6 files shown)


core/gate.h
```diff
@@ -0,0 +1,62 @@
+#ifndef BESS_CORE_GATE_H
```
barath: Add a trailing _ to the header guard.

core/task.cc
```diff
@@ -7,21 +7,43 @@
...
-  return t->m->RunTask(t->arg);
+  struct task_result ret = t->m->RunTask(t->arg);
+
+  // Depth first goes through all pending modules and services
+  while (ctx.gates_pending()) {
+    struct gate_task task = ctx.pop_ogate_and_packets();
+    gate *ogate = task.gate;
+    struct pkt_batch *next_packets = &(task.batch);
+
+#if TRACK_GATES
```
barath: Could you add a comment saying what this does?

core/task.cc
```diff
@@ -7,21 +7,43 @@
...
+
+  return ret;
 }
 
 struct task *task_create(Module *m, void *arg) {
   struct task *t;
 
   t = (struct task *)mem_alloc(sizeof(*t));
-  if (!t) return nullptr;
+  if (!t)
```
barath: Add braces (here and for other ifs)?

core/task.cc
```diff
@@ -112,9 +136,11 @@ void assign_default_tc(int wid, struct task *t) {
 static int get_next_wid(int *wid) {
   static int rr_next = 0;
 
-  if (num_workers == 0) return -1;
+  if (num_workers == 0)
+    return -1;
 
-  while (!is_worker_active(rr_next)) rr_next = (rr_next + 1) % MAX_WORKERS;
+  while (!is_worker_active(rr_next))
```
barath: Add braces?

core/worker.h
```diff
@@ -112,6 +120,28 @@ class Worker {
     return igate_stack_[stack_depth_];
   }
 
+  // Store gate+packets into tasks for worker to service.
```
barath: Maybe mention in the comment "Returns true upon success."

core/task.cc
melvinw: No real need. This gets dropped by #128, which adds documentation somewhere else (unless we care about the commit history)

justinemarie:
Which is not really needed? TRACK GATWS?

melvinw:
Nope, TRACK_GATES is fine. I meant the comment Barath was asking for

core/gate.h
sangjinhan: Also, all other header files in the core directory does not have "CORE" in the header guard. Not sure whether we should, but let's keep them consistent at least for now.

justinemarie:
Okay, added requested changes.

sangjinhan:
Merged, thanks!

PR #131, opened 2016-11-12
Refactor gates and introduce gate hooks

melvinw:
It would be helpful to have a mechanism by which developers can debug pipelines of modules. Currently this is accomplished with #defines and by conditionally tacking extra fields on the end of struct gate. This patch replaces the existing process with a more general mechanism, gate hooks.


sangjinhan:
Merged, thanks!

PR #132, opened 2016-11-15
Minor changes in logging behavior

sangjinhan:
This PR includes the following minor changes:

- Use LOG(FATAL) only for totally unexpected errors.
  - e.g., If another BESS process is running, the error log message should be ERROR, not FATAL.
- Disable glog's default failure handler (stack backtrace), since it interferes with BESS's own fault handler, which prints out more detailed information
- Redirect all outputs to stdout/stderr/cout/cerr to Google glog.
  - If running in foreground mode, glog will print out log messages (depending on the -stderrthreshold option of glog)
- When log messages are printed to stderr (in foreground mode), colored output is enabled by default.
- Restore the "crash log" feature: bessctl shows the stack backtrace if it detects abnormal termination of the process.
- SIGUSR1 signal can be used to print out the current call stack of master/worker threads, without terminating the process.

### DIFF
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -89,7 +89,7 @@ namespace bessd {
 
 void ProcessCommandLineArgs() {
   if (FLAGS_t) {
-    bess::debug::dump_types();
+    bess::debug::DumpTypes();
     exit(EXIT_SUCCESS);
   }
 
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -95,7 +95,6 @@ void ProcessCommandLineArgs() {
 
   if (FLAGS_f) {
     google::LogToStderr();
-    LOG(INFO) << "Launching BESS daemon in process mode...";
     if (!FLAGS_s) {
       VLOG(1) << "TC statistics output is disabled (add -s option?)";
     }
@@ -178,7 +177,7 @@ std::tuple<bool, pid_t> TryAcquirePidfileLock(int fd) {
   return std::make_tuple(lockacquired, pid);
 }
 
-void CheckUniqueInstance(const std::string &pidfile_path) {
+int CheckUniqueInstance(const std::string &pidfile_path) {
   static const int kMaxPidfileLockTrials = 5;
 
   int fd = open(pidfile_path.c_str(), O_RDWR | O_CREAT, 0644);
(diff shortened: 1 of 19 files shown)


core/bessd.cc
```diff
@@ -10,34 +10,102 @@
...
+namespace {
+
+class StreambufLogger : public std::streambuf {
+ public:
+  StreambufLogger(std::ostream &stream, google::LogSeverity severity)
+      : stream_(stream), log_level_(severity) {
+    org_streambuf_ = stream_.rdbuf(this);
+  };
+
+  ~StreambufLogger() { stream_.rdbuf(org_streambuf_); }
```
barath: virtual?

core/bessd.cc
```diff
@@ -10,34 +10,102 @@
...
+// - In process mode (foreground; -f option):
+//   - via Google glog (recommended)
+//     - LOG(*) -> standard error (colored, if applicable)
+//   - via libc/libstdc++
+//     - stdout/cout -> standard output
+//     - stderr/cerr -> standard error (colored, currently always)
+
+namespace {
+
+class StreambufLogger : public std::streambuf {
```
barath: Maybe add class/method comments?

core/bessd.cc
```diff
@@ -10,34 +10,102 @@
...
+    return traits_type::not_eof(v);
+  }
+
+ private:
+  std::ostream &stream_;
+  std::streambuf *org_streambuf_;
+  google::LogSeverity log_level_;
+
+  void WriteToGlog(const char_type *s, std::streamsize count) {
+    if (count <= 0)
```
barath: Braces around if statements bodies.

core/bessd.cc
```diff
@@ -124,54 +192,107 @@ void CheckUniqueInstance(const std::string &pidfile_path) {
...
+  stderr = fopencookie(NULL, "w", stderr_funcs);
+  setvbuf(stderr, NULL, _IOLBF, 0);
+
+  // Redirect stdout output to LOG(INFO) and stderr output to LOG(WARNING)
+	static StreambufLogger stdout_buf(std::cout, google::GLOG_INFO);
+	static StreambufLogger stderr_buf(std::cerr, google::GLOG_WARNING);
+
+  // For whatever reason if fd happens to be assigned 0, 1, or 2, do not close
+  // it since it now points to our custom handler
+  if (fd > 2)
```
barath: Braces.

sangjinhan: Yup, will push changes soon.

PR #134, opened 2016-11-15
Fix dangling pointer bug of Pipeline::attrs_

sangjinhan:
- attrs_ used to maintain a map of string->Attribute*, where the
  attribute pointer is provided by the caller of RegisterAttribute().
  - However the caller may call the function with a stack-allocated
    Attribute instance, then the pointer would be immediately invalid.
  - This issue caused process crash with sample/exactmatch.bess,
    due to memory corruption
  - This issue caused AddMetadataAttr failure with
    samples/wildcardmatch.bess, due to a garbase value of the size of
    already registered attribute with the same name.
- Now Pipeline keeps track or how many modules are actually using
  each attribute.
  - If no modules are using an attribute any longer, we allow modules
    to register the attribute with the same name, possibly with
    a different size.
  - To implement this, deregister logic was implemented, which is
    triggered when a module is destroyed.

### DIFF
--- a/core/metadata.cc
+++ b/core/metadata.cc
@@ -466,6 +466,8 @@ int Pipeline::RegisterAttribute(const std::string &attr_name, size_t size) {
     count++;
     return 0;
   } else {
+    LOG(ERROR) << "Attribute '" << attr_name << "' has size mismatch: registered("
+               << registered_size << ") vs new(" << size << ")";
     return -EINVAL;
   }
 }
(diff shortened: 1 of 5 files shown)


core/metadata.cc
```diff
@@ -452,16 +452,41 @@ int Pipeline::ComputeMetadataOffsets() {
...
-  return (it->second->size == attr->size) ? 0 : -EEXIST;
+
+  size_t registered_size = std::get<0>(it->second);
+  int &count = std::get<1>(it->second);
+
+  if (registered_size == size) {
+    count++;
+    return 0;
+  } else {
+    return -EINVAL;
```
barath: Add a log message here (just so we know that the caller asked for a mismatched size)?

PR #135, opened 2016-11-16
Fix bug when gates are appended out of order

sangjinhan:
In the current code, for example, if output gates are added with indices 0, 1, 2, and 5, then the size of ogates will be 4 (not 6), and ogates[5] (which is not a valid memory area of ogates) will be written.

You can test the issue with samples/exactmatch.bess and samples/wildcardmatch.bess


melvinw:
Thanks for catching/fixing this!

PR #136, opened 2016-11-16
Add module benchmark

sangjinhan:
#130 introduced significant performance degradation: e.g., the throughput of samples/s2s.bess dropped about 35%. Before working on optimization, this PR adds benchmark tests regarding invocation of module code.

### DIFF
--- a/core/worker.cc
+++ b/core/worker.cc
@@ -20,7 +20,9 @@
 int num_workers = 0;
 std::thread worker_threads[MAX_WORKERS];
 Worker *volatile workers[MAX_WORKERS];
-thread_local Worker ctx;
+
+// See worker.h
+__thread Worker ctx;
 
 struct thread_arg {
   int wid;
--- a/core/worker.h
+++ b/core/worker.h
@@ -2,8 +2,10 @@
 #define BESS_WORKER_H_
 
 #include <cstdint>
-#include <glog/logging.h>
 #include <thread>
+#include <type_traits>
+
+#include <glog/logging.h>
 
 #include "gate.h"
 #include "pktbatch.h"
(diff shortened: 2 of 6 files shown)


core/worker.cc
```diff
@@ -20,7 +20,9 @@
 int num_workers = 0;
 std::thread worker_threads[MAX_WORKERS];
 Worker *volatile workers[MAX_WORKERS];
-thread_local Worker ctx;
+
+// See worker.h
+__thread Worker ctx;
```
melvinw: Why this change?

core/worker.h
```diff
@@ -45,24 +45,6 @@ struct gate_task {
...
-        silent_drops_(),
-        current_tsc_(),
-        current_ns_(),
-        igate_stack_(),
-        stack_depth_(),
-        pending_gates_(),
-        splits_() {}
-
-  virtual ~Worker() {}
-
```
melvinw: Ah. Nevermind

core/worker.cc
sangjinhan: worker.h has some brief comments on it. One concern is __thread is a gcc extension, but seems universally supported by other compilers as well.
```
// NOTE: Do not use "thread_local" here. It requires a function call every time
// it is accessed. Use __thread instead, which incurs minimal runtime overhead.
// Worker must be "trivial" type.
```

sangjinhan:
Build fails since it lacks a change I was about to push. Will hold for now.

core/worker.h
```diff
@@ -167,25 +135,26 @@ class Worker {
...
 };
 
 extern int num_workers;
 extern std::thread worker_threads[MAX_WORKERS];
 extern Worker *volatile workers[MAX_WORKERS];
-extern thread_local Worker ctx;
+
+// NOTE: Do not use "thread_local" here. It requires a function call every time
+// it is accessed. Use __thread instead, which incurs minimal runtime overhead.
+// Worker must be "trivial" type.
```
melvinw: might be worthwhile to static assert on std::is_trivial (http://en.cppreference.com/w/cpp/types/is_trivial) here

sangjinhan: Yup, will do

PR #137, opened 2016-11-16
struct snbuf -> class Packet

melvinw:
This is the first of several patches that aim to break the dependency on rte_mbuf.

### DIFF
--- a/core/debug.cc
+++ b/core/debug.cc
@@ -21,7 +21,7 @@
 #include <rte_version.h>
 
 #include "module.h"
-#include "snbuf.h"
+#include "packet.h"
 #include "snobj.h"
 #include "tc.h"
 #include "utils/htable.h"
@@ -293,7 +293,7 @@ static std::string DumpStack() {
       stack << print_code(symbols[i], (i == skips) ? 3 : 0);
     }
 
-    free(symbols); // required by backtrace_symbols()
+    free(symbols);  // required by backtrace_symbols()
   } else
     stack << "ERROR: backtrace_symbols() failed\n";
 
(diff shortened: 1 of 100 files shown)


core/debug.cc
```diff
@@ -342,16 +342,16 @@ static void TrapHandler(int sig_num, siginfo_t *info, void *ucontext) {
...
 #endif
 
   if (is_fatal) {
     oops << "A critical error has occured. Aborting..." << std::endl;
   } else
 
-  oops << "Signal: " << sig_num << " (" << strsignal(sig_num)
-       << "), si_code: " << info->si_code << " ("
-       << si_code_to_str(sig_num, info->si_code) << ")" << std::endl;
+    oops << "Signal: " << sig_num << " (" << strsignal(sig_num)
```
sangjinhan: Oh sorry. the "else" above should be removed

melvinw: fixed

core/packet.cc
```diff
@@ -0,0 +1,226 @@
...
+
+      break;
+    }
+  }
+
+  return ret;
+}
+#endif
+
+void Packet::Dump(FILE *file) {
```
sangjinhan: In the next PR, could you make it return string?

core/packet.h
```diff
@@ -0,0 +1,274 @@
...
+struct rte_mempool *get_pframe_pool();
+struct rte_mempool *get_pframe_pool_socket(int socket);
+
+void init_mempool(void);
+void close_mempool(void);
+
+// For the layout of snbuf, see snbuf_layout.h
+class Packet {
+ public:
+  struct rte_mbuf mbuf() {
```
sangjinhan: It returns the entire copy of rte_mbuf, probably this is not what you really want...

core/packet.h
```diff
@@ -0,0 +1,274 @@
...
+
+  phys_addr_t dma_addr() { return seg_dma_addr(&mbuf_); }
+
+  // TODO: stream operator
+
+  void Dump(FILE *file);
+
+  static Packet *from_paddr(phys_addr_t paddr);
+
+  static inline int mt_offset_to_databuf_offset(
```
sangjinhan: If you are defining functions in the class declaration, you don't need to specify inline (because it will be, even if you don't want)

core/packet.h
```diff
@@ -0,0 +1,274 @@
...
+}
+
+struct rte_mempool *get_pframe_pool();
+struct rte_mempool *get_pframe_pool_socket(int socket);
+
+void init_mempool(void);
+void close_mempool(void);
+
+// For the layout of snbuf, see snbuf_layout.h
+class Packet {
```
sangjinhan: Consider explicitly deleting constructor/destructor, since we don't want them to be used in any case.

core/packet.h
melvinw: Doh. Will fix

sangjinhan:
Merged, thanks! A kitten photo will be rewarded for every merged PR, as our new tradition.
img

PR #138, opened 2016-11-16
Specify module params only if they are not default

sangjinhan:
Added default parameters to Module: no command, one input gate, and one output gate. This configuration works for most common modules. If derived modules use the default parameters they do not need to overload them, allowing the code short and tidy.

Will merge #136 once this one is merged.


melvinw:
Thanks!

Issue #140, opened 2016-11-18
Incorrect offset assignment for metadata attributes


sangjinhan:
samples/vxlan.bess fails (ping fails), due to incorrect offset assignment. I am not sure this is the only bug with samples/vxlan.bess, though.

In the chain of `... -> ip_encap0 -> setattr1 -> ether_encap0 -> ...` (see below),
* `ip_encap0` sets the offset of `ether_type` to 10
* `setattr1` sets the offset of `ether_dst` (6-byte long) to 8, which overwrites `ether_type`
* `ether_encap0` reads the incorrect `ether_type` from offset 10.

Perhaps the bug is in Pipeline::AssignOffsets()? 

```
localhost:10514 $ show module
  bpf::BPF ()
    Input gates:
          0: batches N/A packets N/A port_inc0:0 ->
    Output gates:
          1: batches 41               packets 42               -> 0:vxlan_decap0
    Dump:
      None
  ether_encap0::EtherEncap ()
    Per-packet metadata fields:
      ether_src: read   6 bytes at offset 0
      ether_dst: read   6 bytes at offset 8
     ether_type: read   2 bytes at offset 10
    Input gates:
          0: batches N/A packets N/A setattr1:0 ->
    Output gates:
          0: batches 17               packets 18               -> 0:port_out1
    Dump:
      None
  ip_encap0::IPEncap ()
    Per-packet metadata fields:
         ip_src: read   4 bytes at offset 0
         ip_dst: read   4 bytes at offset 4
       ip_proto: read   1 bytes at offset 8
     ip_nexthop: write  4 bytes (no downstream reader)
     ether_type: write  2 bytes at offset 10
    Input gates:
          0: batches N/A packets N/A vxlan_encap0:0 ->
    Output gates:
          0: batches 17               packets 18               -> 0:setattr1
    Dump:
      None
  port_inc0::PortInc (vport1/VPort)
    Output gates:
          0: batches 70               packets 75               -> 0:bpf
    Dump:
      None
  port_inc1::PortInc (vport0/VPort)
    Output gates:
          0: batches 17               packets 18               -> 0:setattr0
    Dump:
      None
  port_out0::PortOut (vport0/VPort)
    Input gates:
          0: batches N/A packets N/A vxlan_decap0:0 ->
    Dump:
      None
  port_out1::PortOut (vport1/VPort)
    Input gates:
          0: batches N/A packets N/A ether_encap0:0 ->
    Dump:
      None
[...]

PR #141, opened 2016-11-20
fix metadata attribute offset calculation bug

melvinw:
Scope components weren't being sorted correctly, causing AssignOffsets()
to make whacky decisions. This patch fixes DegreeComp() to prevent this
behavior and adds a test to catch it, should it be broken in the future.

This patch fixes issue #140


sangjinhan:
Merged (manually), thanks!
img

PR #142, opened 2016-11-23
declare all modules as final

melvinw:
As pointed out by @barath , this might enable optimizations that could lead to a performance gain.

### DIFF
--- a/core/modules/bpf.h
+++ b/core/modules/bpf.h
@@ -17,7 +17,7 @@ struct filter {
   char *exp;        /* original filter expression string */
 };
 
-class BPF : public Module {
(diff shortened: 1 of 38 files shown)


barath:
Thanks for this.

sangjinhan:
Perhaps we can do the same thing with drivers?

barath:
Melvin and I chatted about doing the same with drivers.  I think it probably won't help (given how the virtual functions for Port are called), but we might as well because it won't hurt.

barath:
Looking at it a bit more, it may help for pmd, so might be worth doing overall.

sangjinhan:
img

PR #145, opened 2016-11-28
fix UnixSocketPort SendPackets bug

melvinw:
The msghdr supplied to sendmsg was not getting initialized properly,
casuing sends to fail. Segemented packets were also handled incorrectly.
This patch fixes these issues.


sangjinhan:
Merged, thanks!
img

PR #146, opened 2016-11-28
Clean up snobj

changlan:


core/module.h
```diff
@@ -50,20 +50,8 @@ class Module;
...
-
-  // if non-zero, workers don't need to be paused in order to
-  // run this command
-  int mt_safe;
-};
-
-template <typename T>
-using Commands = std::vector<struct Command<T>>;
-
 // TODO: Change type name to Command when removing snobj
```
sangjinhan: The time has come :-)

core/module.h
```diff
@@ -72,16 +60,15 @@ struct PbCommand {
   int mt_safe;
```
sangjinhan: In the new bessctl.cc, there is no logic to check if a command is multi-thread safe.

core/module.h
```diff
@@ -182,9 +171,9 @@ class Module {
         ogates() {}
   virtual ~Module() {}
 
-  pb_error_t Init(const google::protobuf::Any &arg);
+  pb_error_t RunInit(const google::protobuf::Any &arg);
```
sangjinhan: I think this one should be moved to the private section, since:
- A derived class does not need to override this function, but the comment above says something different
- If bessctl.cc is the only user (besides module_test.cc), it should be friended.
The name RunInit is too generic and it does not say anything about the difference between RunInit and Init. How about InitWithGenericArg or something like that?

sangjinhan:
There are some documents mentioning `snobj.md`. They also need to be updated.

core/module.h
changlan: Just wanted to confirm, the change seems to be introduced in snctl.cc (https://github.com/NetSys/bess/blame/d9070a8a6aea1be54f0f38a75a902cc4416c6c98/core/snctl.cc#L1295). Should I refactor RunCommand()?

core/module.h
changlan: Renaming this function makes sense. I am not sure whether this function should be private:
- It is meant to be used by external users, whereas private member functions are mostly for internal use.
- There are other public member functions that a derived class does not override. I can move this function to a different public section.
- Google C++ guide does not encourage the usage of friend classes.

barath:
Thanks!

barath:
photo 188

PR #147, opened 2016-11-28
add zero copy vport test

melvinw:


core/dpdk.cc
```diff
@@ -25,13 +25,16 @@ static int get_numa_count() {
   int cnt;
 
   fp = fopen("/sys/devices/system/node/possible", "r");
-  if (!fp) goto fail;
+  if (!fp)
```
barath: Braces (here and below).

PR #149, opened 2016-11-30
Change build.py to accept lib locations

apanda:
This changes build.py so that one can specify where Google Benchmark library is stored.

This change only adds a flag for one library, but it can easily be extended to others. I mostly did this because I am personally a bit weirded out by cloning the repository and running `make install`. I think it might be good to have similar stuff for gRPC (which needs to be built manually at the moment), protoc, and a few others. I can do that later, mostly wanted to test the waters to see if this was something others were OK with.

By the way, during development I noticed that `build.py` does not check `protoc` version, so it will get pretty far with old protobuf installations, but then fail in the Python bit.

### DIFF
--- a/build.py
+++ b/build.py
@@ -166,19 +166,20 @@ def check_mlx():
         set_config(DPDK_FINAL_CONFIG, 'CONFIG_RTE_LIBRTE_MLX4_PMD', 'n')
         set_config(DPDK_FINAL_CONFIG, 'CONFIG_RTE_LIBRTE_MLX5_PMD', 'n')
 
-def generate_extra_mk():
+def generate_dpdk_extra_mk():
     global extra_libs
-    global cxx_flags
-    global ld_flags
 
-    with open('core/extra.mk', 'w') as fp:
+    with open('core/extra.dpdk.mk', 'w') as fp:
         fp.write('LIBS += %s\n' % \
                 ' '.join(map(lambda lib: '-l' + lib, extra_libs)))
-        fp.write('CFLAGS += %s\n' % \
-                ' '.join(cxx_flags))
+
+def generate_extra_mk():
+    global cxx_flags
+    global ld_flags
+    with open('core/extra.mk', 'w') as fp:
         fp.write('CXXFLAGS += %s\n' % \
                 ' '.join(cxx_flags))
-        fp.write('LD_FLAGS += %s\n' % \
+        fp.write('LDFLAGS += %s\n' % \
                 ' '.join(cxx_flags))
 
 def download_dpdk():
@@ -201,7 +202,7 @@ def configure_dpdk():
 
         check_mlx()
 
-        generate_extra_mk()
+        generate_dpdk_extra_mk()
 
         cmd('cp -f %s %s' % (DPDK_FINAL_CONFIG, DPDK_ORIG_CONFIG))
         cmd('make -C %s config T=%s' % (DPDK_DIR, DPDK_TARGET))
@@ -239,6 +240,8 @@ def build_bess():
 
     if not os.path.exists('%s/build' % DPDK_DIR):
         build_dpdk()
+
+    generate_extra_mk()
 
     print 'Generating protobuf codes for libbess-python...'
     cmd('protoc protobuf/*.proto \
@@ -293,8 +296,8 @@ def print_usage(parser):
 
 def update_benchmark_path(path):
     print 'Specified benchmark path %s' % path
-    cxx_flags.extend(['-I', '%s/include'%(path)])
-    ld_flags.extend(['-L', '%s/lib'%(path)])
+    cxx_flags.extend(['-I%s/include'%(path)])
+    ld_flags.extend(['-L%s/lib'%(path)])
 
 
 def main():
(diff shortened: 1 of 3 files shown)


apanda:
By the way, the Travis job is almost sure going to fail. This is because https://github.com/NetSys/bess/blob/c%2B%2B/.travis.yml#L37 seems to refer to a file `core/all_tests`, which does not seem to exist.

build.py
```diff
@@ -238,7 +241,9 @@ def build_bess():
     print 'Building BESS daemon...'
     cmd('bin/bessctl daemon stop 2> /dev/null || true')
     cmd('rm -f core/bessd')     # force relink as DPDK might have been rebuilt
-    cmd('make -C core')
+    if cxx_flags or ld_flags:
+        cmd('make -C core CXXFLAGS=%s CFLAGS=%s LDFLAGS=%s' %\
```
sangjinhan: One concern with this is that once `build.py` is run with custom flags, subsequent builds without `build.py` (i.e., running `make` in the `core` directory) will fail. This could be problematic since during development phase developers should be able to use the `Makefile` without without dealing with `build.py`. Somehow the script (which is currently like a poor man's autoconf) should remember special flags that were used during the initial build.

The `core/extra.mk` file is supposed to be used for this purpose, although the current usage is limited for additional libraries required by DPDK build.

build.py
```diff
@@ -97,8 +100,8 @@ def check_c_lib(lib):
         with open(test_c_file, 'w') as fp:
             fp.write(textwrap.dedent(src))
 
-        return cmd_success('gcc %s -l%s -o %s' % \
-                (test_c_file, lib, test_e_file))
+        return cmd_success('gcc %s -l%s %s %s -o %s' % \
+                (test_c_file, lib, cxx_flags, ld_flags, test_e_file))
```
sangjinhan: If `cxx_flags` and `ld_flags` are lists, they should be converted to strings with `'  '.join()`?

build.py
```diff
@@ -282,30 +287,38 @@ def print_usage():
...
+    print 'Specified benchmark path %s' % path
+    cxx_flags.extend(['-I', '%s/include'%(path)])
+
+
 def main():
     os.chdir(BESS_DIR)
+    parser = argparse.ArgumentParser(description = 'Build Bess')
+    parser.add_argument('action', metavar='action', nargs='?', default='all', \
+            help='Action is one of all, dpdk, bess, kmod, clean, dist_clean, help')
+    parser.add_argument('--with-benchmark', dest='benchmark_path', nargs=1, help='Location of benchmark library')
```
sangjinhan: Rather than introducing new options, one may just run `CXXFLAGS=-I/opt/benchmark/include build.py` on the shell? (or `export` the flag so that it can persist for subsequent builds in the `core` directory)

apanda:
@sangjinhan Fixed those comments, the flag vs env var thing is up to you. Whichever is fine.

build.py
```diff
@@ -165,10 +168,18 @@ def check_mlx():
 
 def generate_extra_mk():
```
sangjinhan: I am a bit concerned that `generate_extra_mk()` is called only during configure_dpdk(). If DPDK is already built (e.g., the docker image for `container_build.py` comes with pre-built DPDK) the extra.mk file will never be updated.

One potential solution to address this issue is that we separate `core/extra_dpdk.mk` from `core/extra.mk`, both of which are included by `core/Makefile` (.gitignore should include the both, of course), so that `core/extra.mk` can be updated by `build_bess()`.

build.py
```diff
@@ -165,10 +168,18 @@ def check_mlx():
...
     global extra_libs
+    global cxx_flags
+    global ld_flags
 
     with open('core/extra.mk', 'w') as fp:
-        fp.write('LIBS += %s ' % \
+        fp.write('LIBS += %s\n' % \
                 ' '.join(map(lambda lib: '-l' + lib, extra_libs)))
+        fp.write('CFLAGS += %s\n' % \
```
sangjinhan: Maybe we can skip this one, since i) `CFLAGS` is never used and ii) in general there is no guarantee that flags for C++ will also work for C.

build.py
```diff
@@ -276,37 +287,45 @@ def do_dist_clean():
...
-    print >> sys.stderr, \
-            'Usage: %s [all|dpdk|bess|kmod|clean|dist_clean|help]' % \
-            sys.argv[0]
+def print_usage(parser):
+    parser.print_help(file=sys.stderr)
     sys.exit(2)
 
+def update_benchmark_path(path):
+    print 'Specified benchmark path %s' % path
+    cxx_flags.extend(['-I', '%s/include'%(path)])
```
sangjinhan: just a minor nit: let's not add a whitespace after `-I` and `-L`, to be consistent with the existing `Makefile` style

sangjinhan: i.e., `cxx_flags.append('-I%s/include' % path)`

apanda:
@sangjinhan Addressed those. Sorry about the number of back and forths.

sangjinhan:
Merged, thanks!
`core/all_test` is a byproduct made during the build process.
img

barath:
Just talking to Panda -- was it intentional that this was closed and not merged?  (Was it getting merged in via some other branch?)

sangjinhan:
Sorry forgot to push. Merged via the Web UI.

PR #150, opened 2016-11-30
replace weird memsets

melvinw:


barath:
@changlan Would it be possible to run your clang checks against this?

changlan:
clang also warns about memmove (like this: https://github.com/NetSys/bess/blob/c%2B%2B/core/modules/wildcard_match.cc#L338), which also overwrites vtable.

melvinw:
kk. Is that all? Could you post the output here?

changlan:
Yes, I think so. The message is

```
modules/wildcard_match.cc:340:13: error: destination for this 'memmove' call is a pointer to class containing a dynamic
      class 'HTable<wm_hkey_t, WmData, &WildcardMatch::wm_keycmp, &WildcardMatch::wm_hash>'; vtable pointer will be
      overwritten [-Werror,-Wdynamic-class-memaccess]
    memmove(&tuples_[idx], &tuples_[idx + 1],
    ~~~~~~~ ^
modules/wildcard_match.cc:340:13: note: explicitly cast the pointer to silence this warning
    memmove(&tuples_[idx], &tuples_[idx + 1],
            ^
            (void*)
```

melvinw:
Thanks. Will push a fix

core/modules/wildcard_match.cc
```diff
@@ -335,7 +335,10 @@ int WildcardMatch::DelEntry(struct WmTuple *tuple, wm_hkey_t *key) {
     tuple->ht.Close();
 
     num_tuples_--;
-    memmove(&tuples_[idx], &tuples_[idx + 1],
+
+    // shift remaing tuples to the left
+    memmove(reinterpret_cast<void*>(&tuples_[idx]),
```
sangjinhan: Perhaps we can update tuples_ (and fields_ as well) with vectors? It should make the code much simpler (although there might be a small penalty with element dereference on the datapath)

barath: Agreed -- Melvin and I were discussing that and I think there are several places where such a change can be made.  (The references shouldn't cost any more once optimized, I think.)

core/worker.cc
```diff
@@ -185,7 +185,7 @@ void *Worker::Run(void *_arg) {
   rte_thread_set_affinity(&set);
 
   /* just in case */
-  memset(this, 0, sizeof(*this));  // FIXME
+  Clear();
```
barath: I'm not sure we can use assert() because it will disappear with NDEBUG.

melvinw: A `check()` should work just as well

sangjinhan: Should be fine to disappear in the debug mode, since it indeed is an assertion.

barath: Sounds good.

core/modules/wildcard_match.cc
melvinw: Kind of tangential, but it would be nice if we had a way to measure the impact of these sorts of changes. Individual module benchmarks seems kind of tedious, so maybe just for those that ought to be fast-- or this might also just be something we punt to integration testing.

core/utils/htable.cc
```diff
@@ -351,7 +351,18 @@ int HTableBase::Init(size_t key_size, size_t value_size) {
 void HTableBase::Close() {
   mem_free(buckets_);
   mem_free(entries_);
-  memset(this, 0, sizeof(*this));
+  key_size_ = 0;
```
melvinw: kk

core/modules/wildcard_match.cc
sangjinhan: Perhaps we can devise a common framework for the performance of every module?
* We define a few input sets (packet batches), each of which represents a combination of various packets
* Benchmark is done with every combination of input set and module (thus producing a table)
  * If the kNumIGates of a module is zero we can skip the module.
  * We create a dummy module, and connect the module under test to the dummy (via each of kNumOGates output gates).
* The benchmark code directly call module's ProcessBatch() directly.
* It can be done with google benchmark or our custom code. The latter case is interesting in that:
  * We can measure the latency at the cycle granularity with `rdtscp`. This is beneficial since the wall-clock time used by gbenchmark is not constant)
  * All other overheads (google benchmark, packet alloc/free, etc.) can be excluded.
  * min/median latency for each call can be calculated, which will be much more consistent across runs than average, without noise.

sangjinhan: At some point we should rework /util/histogram.h, for the following desired properties:
* it should not internally impose any specific time units (e.g., nanoseconds or cycles).
  * actually it doesn't need to be 'time' in the first place.
* logarithmic binning, since latency tends to be long-tailed.
* not really required at the moment, but ideally it can be templatized to support floating point numbers as well.

melvinw: Good idea. We would need a way for module writers to somehow set up the appropriate module state, though. I'll write up a todo.

barath:
img_0965

PR #151, opened 2016-12-01
Build with make CXX=clang++

changlan:
- Fix build issues with CXX=clang++
- Fix a few memory leaks detected by LLVM sanitizer

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -54,7 +54,7 @@ CXXFLAGS += -std=gnu++11 -g3 -ggdb3 -Ofast -march=native \
             -Werror -isystem $(DPDK_INC_DIR) -D_GNU_SOURCE \
             -Wall -Wextra -Wcast-align
 
-PERMISSIVE = -Wno-missing-field-initializers -Wno-unused-parameter # FIXME
+PERMISSIVE = -Wno-unused-parameter
 
 # -Wshadow should not be used for g++ 4.x, as it has too many false positives
 CXXVERSION_5_OR_HIGHER := $(shell expr $(CXXVERSION) \>= 50000)
--- a/core/Makefile
+++ b/core/Makefile
@@ -48,7 +48,7 @@ else
 endif
 
 CXXFLAGS += -std=gnu++11 -g3 -ggdb3 -Ofast -march=native \
-	    -Werror -isystem $(DPDK_INC_DIR) -D_GNU_SOURCE \
+			-Werror -isystem $(DPDK_INC_DIR) -D_GNU_SOURCE \
 			-Wall -Wextra -Wcast-align
 
 PERMISSIVE = -Wno-missing-field-initializers -Wno-unused-parameter # FIXME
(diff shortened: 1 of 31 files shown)


core/gate_test.cc
```diff
@@ -41,7 +46,9 @@ class IOGateTest : public ::testing::Test {
 
 TEST_F(GateTest, AddExistingHookFails) {
   ASSERT_EQ(0, g->AddHook(new TrackGate()));
-  ASSERT_EQ(EEXIST, g->AddHook(new TrackGate()));
+  GateHook *hook = new TrackGate();
+  ASSERT_EQ(EEXIST, g->AddHook(hook));
+  delete hook;
```
sangjinhan: This will cause `g` to hold a freed pointer to `hook`...?

core/utils/htable.cc
```diff
@@ -376,6 +369,7 @@ void *HTableBase::GetHash(uint32_t pri, const void *key) const {
 
   /* check primary bucket */
   ret = get_from_bucket(pri, pri, key);
+
```
sangjinhan: Surely it's a realm of personal preference, but can we not have a blank line between an assignment and its failure check? The Guide recommends minimizing the use of vertical space... (although I don't always agree with the principle)

core/Makefile
```diff
@@ -164,7 +172,13 @@ $(eval $(call BUILD, \
 	CXX, \
 	%.pb.o, \
 	%.pb.cc $(DEPDIR)/$$@.d, \
-	$$(CXX) -o $$@ -c $$< $$(PROTOCXXFLAGS) $$(DEPFLAGS)))
+	$$(CXX) -o $$@ -c $$< $$(CXXFLAGS) $$(PERMISSIVE) $$(DEPFLAGS)))
+
+$(eval $(call BUILD, \
+	CXX, \
+	bessctl.o, \
```
sangjinhan: Why does this file require special treatment? If the issue is not straightforward to fix, please leave a brief comment in the Makefile.

core/Makefile
```diff
@@ -12,15 +12,19 @@ DEPFLAGS = -MT $@ -MMD -MP -MF $(DEPDIR)/$(@:.o=.d)
...
+$(shell mkdir -p $(DEPDIR)/hooks >/dev/null)
 
 # e.g., 4.9.3 -> 40903
 CXXVERSION := $(shell $(CXX) -dumpversion | sed -e 's/\.\([0-9][0-9]\)/\1/g' -e 's/\.\([0-9]\)/0\1/g' -e 's/^[0-9]\{3,4\}$$/&00/')
-
 CXXVERSION_4_8_OR_HIGHER := $(shell expr $(CXXVERSION) \>= 40800)
-ifeq "$(CXXVERSION_4_8_OR_HIGHER)" "0"
-  $(error g++ 4.8 or higher is required)
+
+IS_GCC := $(shell expr "$(shell $(CXX) --version)" : "g++")
```
sangjinhan: Minor nit: perhaps we can use something like `CXXCOMPILER := $(word 1, $(shell $(CXX) --version))`, so that the value can be either "g++" or "clang"? The name `IS_GCC` implies it is boolean, i.e., 0 or 1, while it's not. Also, while it's very unlikely, in a future version `clang++ --version` may print out "clan**g++**" somewhere in the message 😛

core/Makefile
```diff
@@ -12,15 +12,19 @@ DEPFLAGS = -MT $@ -MMD -MP -MF $(DEPDIR)/$(@:.o=.d)
...
+
+ifeq "$(IS_GCC)" "3"
+		ifeq "$(CXXVERSION_4_8_OR_HIGHER)" "0"
```
sangjinhan: For Makefile the tab-width should be 8, thus a single hard TAB character should suffice here

core/gate_test.cc
changlan: No. AddHook() will return EEXIST without storing hook.

changlan:
Comments addressed.

melvinw:
kitty
Thanks!

PR #156, opened 2016-12-07
Build bessd in parallel

sangjinhan:
Since the core/Makefile now does not do parallel build by default, we
add -j option to the build.py script


changlan:
Thanks!

image

PR #157, opened 2016-12-07
Add Clang builds to Travis CI

changlan:
- Add Clang builds to Travis CI
- PEP-8 formatting

### DIFF
--- a/.travis/rebuild_image.sh
+++ b/.travis/rebuild_image.sh
@@ -4,7 +4,7 @@ set -e
 
 REPOSITORY=${REPOSITORY:-nefelinetworks/bess_build}
 VERSION=${VERSION:-`date +%y%m%d`}
-docker build --build-arg BESS_DPDK_BRANCH=${BESS_DPDK_BRANCH:-develop} -t $REPOSITORY:latest -t $REPOSITORY:$VERSION .
+docker build --build-arg BESS_DPDK_BRANCH=${BESS_DPDK_BRANCH:-c++} -t $REPOSITORY:latest -t $REPOSITORY:$VERSION .
 
 echo Build succeeded: $REPOSITORY:$VERSION
 echo Build succeeded: $REPOSITORY:latest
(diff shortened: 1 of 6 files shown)


.travis/rebuild_image.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 REPOSITORY=${REPOSITORY:-nefelinetworks/bess_build}
 VERSION=${VERSION:-`date +%y%m%d`}
-docker build --build-arg BESS_DPDK_BRANCH=${BESS_DPDK_BRANCH:-develop} -t $REPOSITORY:latest -t $REPOSITORY:$VERSION .
+docker build --build-arg BESS_DPDK_BRANCH=${BESS_DPDK_BRANCH:-c++} -t $REPOSITORY:latest -t $REPOSITORY:$VERSION .
```
justinemarie: Aren't we moving c++ to develop soon? Why c++ now?

build.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2.7
+#!/usr/bin/env python
```
justinemarie: Is this recommended? I know that between versions of python there are often compatibility issues...

justinemarie:
Most of these changes are due to formatting script...

.travis/rebuild_image.sh
changlan: This script broke due to a previous change to the develop branch. We should keep it working.
Also: perhaps we won't be renaming c++ to develop.

build.py
changlan: - In general I think we should support Python 2.x (which we do). I don't think we used anything 2.7-specific so it'd be better to keep it general.

- python2.7 is a Linux distro-specific name.

changlan:
Yup, those PEP-8 formatting changes are intentional. :)

melvinw:
LGTM

melvinw:
cat
Thanks!

PR #158, opened 2016-12-07
call modules recursively

justinemarie:
This switches back to calling modules recursively to improve throughput.

### DIFF
--- a/core/module.h
+++ b/core/module.h
@@ -12,7 +12,7 @@
 #include "task.h"
 
 static inline void set_cmd_response_error(pb_cmd_response_t *response,
-                                          const pb_error_t &error) {
+    const pb_error_t &error) {
   response->mutable_error()->CopyFrom(error);
 }
 #define MODULE_NAME_LEN 128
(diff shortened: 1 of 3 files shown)


core/module.h
```diff
@@ -12,7 +12,7 @@
 #include "task.h"
 
 static inline void set_cmd_response_error(pb_cmd_response_t *response,
-                                          const pb_error_t &error) {
+    const pb_error_t &error) {
```
changlan: Nit: Unnecessary indentation changes to the file (along with other lines below)

justinemarie: I ran clang-format -- is that no longer recommended?

changlan: That doesn't seem like the clang-format output. Perhaps this file is missed? (just guessing)

justinemarie:
Looks like that quirk is gone now that I actually clang-formatted

barath:
Dumb question -- where are the stack-allocated packet batches in this PR?

justinemarie:
They are allocated in RunTask or in ProcessBatch (in the modules themselves).

barath:
Are no changes needed in RunTask or ProcessBatch then to revert to that behavior?

justinemarie:
That's correct.
* The current code in the C++ branch allows them to pass along a pointer to their stack-allocated PacketBatch to RunChooseModule, and then RunChooseModule copies it in to the heap and returns. So they were always stack-allocating and then we just copied them.
* The new-new version that was never committed to the C++ allocated everything from the mempool in the heap, and that required changes to all the modules. But I never made a pull request for that version since it still wasn't as fast as the recursive version.

justinemarie:
What is this failure?

justinemarie:
image

melvinw:
cat
Thanks!

PR #161, opened 2016-12-08
A set of fixes in porting from C to C++

apanda:
I can break this into smaller pull requests if that helps, but the code changes should be quite minimal. The main changes of interest are:

- PMD ports were unusable with PCI specs, this change reallows that. I don't think the solution is particularly elegant, but I had limited options because of choices made by protobuf.

- The rewrite module reliably segfaulted for me, fixed that by changing memcpy used.

- Measure did not forward most packets, so a minor fix for that module.

- The histogram bits in utils were unusable.

### DIFF
--- a/core/modules/measure.cc
+++ b/core/modules/measure.cc
@@ -38,50 +38,48 @@ void Measure::ProcessBatch(bess::PacketBatch *batch) {
     start_time_ = get_time();
   }
 
-  if (static_cast<int>(HISTO_TIME_TO_SEC(time - start_time_)) < warmup_) {
+  if (static_cast<int>(HISTO_TIME_TO_SEC(time - start_time_)) >= warmup_) {
     RunNextModule(batch);
-    return;
-  }
+    pkt_cnt_ += batch->cnt();
 
-  pkt_cnt_ += batch->cnt();
+    for (int i = 0; i < batch->cnt(); i++) {
+      uint64_t pkt_time;
+      if (get_measure_packet(batch->pkts()[i], &pkt_time)) {
+        uint64_t diff;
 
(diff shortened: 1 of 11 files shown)


core/modules/measure.cc
```diff
@@ -38,50 +38,48 @@ void Measure::ProcessBatch(bess::PacketBatch *batch) {
...
+        bytes_cnt_ += batch->pkts()[i]->total_len();
+        total_latency_ += diff;
 
-      record_latency(&hist_, diff);
+        record_latency(&hist_, diff);
+      }
     }
+    RunNextModule(batch);
   }
-}
```
melvinw: this should come back. travis probably already yelled about this

core/modules/measure.cc
```diff
@@ -38,50 +38,48 @@ void Measure::ProcessBatch(bess::PacketBatch *batch) {
...
-      bytes_cnt_ += batch->pkts()[i]->total_len();
-      total_latency_ += diff;
+        bytes_cnt_ += batch->pkts()[i]->total_len();
+        total_latency_ += diff;
 
-      record_latency(&hist_, diff);
+        record_latency(&hist_, diff);
+      }
     }
+    RunNextModule(batch);
```
melvinw: Why call `RunNextModule` twice? Is this behavior defined?

apanda: Oops, it was a merge conflict, sorry I did not build before submitting (serves me right). Let me fix it and update it shortly.

apanda:
Fixed those comments, sorry.

melvinw:
Thanks! Looks good to me otherwise. Will approve after CI passes

melvinw:
Thanks! 
cat

PR #163, opened 2016-12-08
Optimizes round robin traffic class and adds a round robin benchmark.

barath:


core/traffic_class.cc
```diff
@@ -200,15 +198,17 @@ void RoundRobinTrafficClass::FinishAndAccountTowardsRoot(
...
-    if (children_.empty()) {
-      blocked_ = true;
-    }
   } else {
-    children_.push_back(child);
+    ++next_child_;
+  }
+
+  // Wrap around for round robin.
+  if (next_child_ >= children_.size()) {
```
melvinw: is this faster than using `%`?

melvinw:
looks good! just one small comment

core/traffic_class.cc
barath: I believe so, but I didn't benchmark it.  Let me do so and will report back.

melvinw:
thanks! 
cat

PR #164, opened 2016-12-08
Removes cdlist and changes flowgen to use std::deque instead.

barath:
Also removes minheap which was already unused.


melvinw:
looks good! just waiting on CI

melvinw:
Thanks! 
cat2

PR #167, opened 2016-12-09
Finish refactoring Module

melvinw:


core/module.h
```diff
@@ -233,6 +232,26 @@ class Module {
     return attrs_;
   }
 
+  const std::vector<Task *> tasks() const {
```
barath: Mark accessors as inline.

barath:
Would you be able to rebase?

melvinw:
Done

barath:
You're right -- my mistake.  I forgot this was a .h file.

melvinw:
rolled that change back

barath:
image

core/module.h
```diff
@@ -234,6 +233,28 @@ class Module {
...
+    return tasks_;
+  }
+
+  void set_attr_offset(size_t idx, bess::metadata::mt_offset_t offset) {
+    if (idx < bess::metadata::kMaxAttrsPerModule) {
+      attr_offsets_[idx] = offset;
+    }
+  }
+
+  const bess::metadata::mt_offset_t *all_attr_offsets() const {
```
sangjinhan: If it is used in the form of `all_attr_offsets()[i])` in most cases, how about using `mt_offset_t attr_offset(size_t idx)` instead?

melvinw: Ah, that's cleaner. Will submit another pr

PR #168, opened 2016-12-09
Canonicalize C++ code as per style guide

changlan:
- Fix warnings from cpplint that are not false positives. 
- Run clang-format

### DIFF
--- a/core/debug.cc
+++ b/core/debug.cc
@@ -6,6 +6,10 @@
 #include <ucontext.h>
 #include <unistd.h>
 
+#include <glog/logging.h>
+#include <rte_config.h>
+#include <rte_version.h>
+
 #include <cassert>
 #include <csignal>
 #include <cstdint>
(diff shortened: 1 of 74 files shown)


core/module.h
```diff
@@ -291,7 +292,7 @@ inline void Module::RunChooseModule(gate_idx_t ogate_idx,
   }
 
   ctx.set_current_igate(ogate->igate_idx());
-  ((Module *)ogate->arg())->ProcessBatch(batch);
+  (reinterpret_cast<Module *>(ogate->arg()))->ProcessBatch(batch);
```
barath: static_cast instead?

core/module_test.cc
```diff
@@ -155,7 +155,7 @@ TEST_F(ModuleTester, RunCommand) {
     response = m->RunCommand("foo", arg);
     EXPECT_EQ(0, response.error().err());
   }
-  EXPECT_EQ(10, ((AcmeModule *)m)->n);
+  EXPECT_EQ(10, (reinterpret_cast<AcmeModule *>(m))->n);
```
barath: static_cast instead?

core/debug.cc
```diff
@@ -6,6 +6,10 @@
 #include <ucontext.h>
 #include <unistd.h>
 
+#include <glog/logging.h>
```
changlan: Looks like it isn't. Functions in debug.h still get called.

barath:
image

PR #169, opened 2016-12-10
bessctl: remove total from monitor tc

melvinw:
It doesn't make sense and is already reported by the root node


barath:
<pre>
|\---/|
| o_o |
 \_^_/
</pre>

PR #170, opened 2016-12-10
Modifies license copyright.

barath:


LICENSE
```diff
@@ -1,6 +1,7 @@
 BESS is distributed under the following BSD3 license:
 
-Copyright (c) 2014-2016, Regents of the University of California and Nefeli Networks, Inc.  All rights reserved.
+Copyright (c) 2014-2016, Regents of the University of California.  All rights reserved.
```
melvinw: There's an extra space between "." and "All"

LICENSE
```diff
@@ -1,6 +1,7 @@
...
+Copyright (c) 2016, Nefeli Networks, Inc.  All rights reserved.
```
melvinw: Same here

LICENSE
barath: Will update.  My habit has always been two spaces after periods, but I know that's less common these days:

https://en.wikipedia.org/wiki/History_of_sentence_spacing#French_and_English_spacing

barath:
Got feedback from Kevin, made an additional edit.

melvinw:
Thanks! 
moneycat

PR #171, opened 2016-12-12
module: add indexed attr_offsets_ accessor

melvinw:


core/module.h
```diff
@@ -243,6 +243,11 @@ class Module {
     }
   }
 
+  bess::metadata::mt_offset_t attr_offset(size_t idx) const {
+    CHECK_LT(idx, bess::metadata::kMaxAttrsPerModule);
```
barath: Maybe use a DCHECK so there isn't the performance hit in production?

barath:
photo 197

PR #174, opened 2016-12-12
bessctl/scripts: fix queue

melvinw:


changlan:
Thanks!
image

sangjinhan:
We need to discuss whether to expose leaf TCs to end users. The old script was short and simple, but now it has become somewhat verbose...

melvinw:
@barath and I were talking about this earlier. The verbosity should go away once we combine LeafTC and Task

PR #175, opened 2016-12-13
fix divide by zero bug causing bpf.bess to crash

justinemarie:
Summary says it all -- quick fix to the bpf.bess crash bug.


sangjinhan:
img

PR #178, opened 2016-12-13
modules/measure: add median, 99% latency to output

melvinw:
This is just quick patch for the current milestone. Ideally this would incorporate a rewrite of utils/histogram.


sangjinhan:
img

PR #179, opened 2016-12-13
scripts: fix vxlan.bess

changlan:
- bessctl: Implicitly convert to name when use Port/Module as parameter
- module/set_metadata: Accept both value_hex and value_int
- bessctl: Refuse to run most commands when daemon is not connected

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -29,6 +29,23 @@ class ConfError(Exception):
     pass
 
 
+@staticmethod
+def _choose_arg(arg, kwargs):
+    if kwargs:
+        if arg:
+            raise TypeError('You cannot specify both arg and keyword args')
+
+        for key in kwargs:
+            if isinstance(kwargs[key], (Module, Port)):
+                kwargs[key] = kwargs[key].name
+
+        return kwargs
+
+    if isinstance(arg, (Module, Port)):
+        return arg.name
+    else:
+        return arg
+
 def __bess_env__(key, default=None):
     try:
         return os.environ[key]
@@ -593,7 +610,7 @@ def _do_run_file(cli, conf_file):
             raise cli.InternalError('Invalid driver name: %s' % name)
 
         new_globals[name] = type(str(name), (Port,),
-                                 {'bess': cli.bess})
+                                 {'bess': cli.bess, 'choose_arg': _choose_arg})
 
     # Add BESS module classes
     for name in class_names:
@@ -601,7 +618,7 @@ def _do_run_file(cli, conf_file):
             raise cli.InternalError('Invalid module class name: %s' % name)
 
         new_globals[name] = type(str(name), (Module,),
-                                 {'bess': cli.bess})
+                                 {'bess': cli.bess, 'choose_arg': _choose_arg})
 
     code = compile(xformed, conf_file, 'exec')
 
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -513,6 +513,10 @@ def daemon_start(cli, opts):
         _do_start(cli, opts)
 
 
+def is_bess_connected(cli):
+    return cli.bess.is_connected()
+
+
 def is_pipeline_empty(cli):
     workers = cli.bess.list_workers().workers_status
     modules = cli.bess.list_modules().modules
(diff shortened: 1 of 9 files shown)


bessctl/conf/samples/vxlan.bess
```diff
@@ -36,33 +36,34 @@ Container Bob  |    (regular Ethernet port)    |
...
+import struct
 
-def aton(ip):
-    return socket.inet_aton(ip)
+def atoh(ip):
+    return struct.unpack("!L", socket.inet_aton(ip))[0]
 
-os.system('docker pull ubuntu > /dev/null')
-os.system('docker run -i -d --net=none --name=vxlan_test ubuntu /bin/bash > /dev/null')
+os.system('sudo docker pull ubuntu > /dev/null')
```
sangjinhan: You don't need root privilege to run Docker containers, if the user belongs to `docker` group.

bessctl/conf/samples/vxlan.bess
```diff
@@ -36,33 +36,34 @@ Container Bob  |    (regular Ethernet port)    |
...
+os.system('sudo docker run -i -d --net=none --name=vxlan_test ubuntu /bin/bash > /dev/null')
 
 # Alice live outside the container, wanting to talk to Bob in the container
 v_bob = VPort(ifname='eth_bob', docker='vxlan_test', ip_addrs=['172.16.100.33/28'])
 v_alice = VPort(ifname='eth_alice', ip_addrs=['10.0.10.2/24'], mac_addr='02:0a:0b:0c:0d:0e')
 
 # Alice -> Bob
-PortInc(port=v_alice) -> bpf::BPF():1 -> VXLANDecap() -> PortOut(port=v_bob)
+PortInc(port=v_alice.name) -> bpf::BPF():1 -> VXLANDecap() -> PortOut(port=v_bob.name)
```
sangjinhan: Originally when `Module` or `Port` instances appear as a parameter, they used to be implicitly converted to its name (string). Please see https://github.com/NetSys/bess/blob/c-legacy/bessctl/commands.py#L514

sangjinhan: (so that other existing scripts can work without modification)

bessctl/conf/samples/vxlan.bess
```diff
@@ -36,33 +36,34 @@ Container Bob  |    (regular Ethernet port)    |
...
-             {'name': 'tun_ip_dst', 'size': 4, 'value': aton('10.0.10.2')},
+            [{'name': 'tun_ip_src', 'size': 4, 'value': atoh('10.0.10.1')},
+             {'name': 'tun_ip_dst', 'size': 4, 'value': atoh('10.0.10.2')},
              {'name': 'tun_id', 'size': 4, 'value': 999}]) \
     -> VXLANEncap(dstport=4789) \
     -> IPEncap() \
     -> SetMetadata(attrs=
-            [{'name': 'ether_src', 'size': 6, 'value': '\x02\x01\x02\x03\x04\x05'},
-             {'name': 'ether_dst', 'size': 6, 'value': '\x02\x0a\x0b\x0c\x0d\x0e'}]) \
+            [{'name': 'ether_src', 'size': 6, 'value': 0x050403020102},
```
sangjinhan: The constant number 0x050403020102 is implicitly assuming a little-endian system. It would be nice to keep the code endian-neutral...

I think `value` used to accept both integer and string. If this is not feasible with PB, maybe we can have separate fields: `value_hex` (`value_str`?) and `value_int`.

bessctl/commands.py
```diff
@@ -530,6 +534,10 @@ def _do_reset(cli):
 
 @cmd('daemon reset', 'Remove all ports and modules in the pipeline')
 def daemon_reset(cli):
+    if not is_bess_connected(cli):
```
changlan: Yes, I think so. I couldn't think of a way to implement that without intricate the code. Can think about it in a next PR.

sangjinhan: Sure, no problem. Thanks!

sangjinhan:
When making a next PR, could you rename value_hex to value_bin? ;)

🐈 🐈 🐈 🐈 🐈 🐈 🐈 🐈

PR #182, opened 2016-12-14
bessctl: fix show tc

melvinw:
This patch fixes `show tc` to correctly report which tcs belong to which workers. It also changes `show tc` to report tcs for all workers when no worker is specified.


barath:
Could you add a small unit test in traffic_class_test for Size()?  (And optionally Traverse()?)

melvinw:
Sure thing

barath:
image

PR #183, opened 2016-12-14
traffic_class: free leaf tc tasks when resetting

melvinw:


core/traffic_class.cc
```diff
@@ -322,9 +322,10 @@ bool TrafficClassBuilder::ClearAll() {
 
     if (c->policy_ == POLICY_LEAF) {
       LeafTrafficClass *l = static_cast<LeafTrafficClass *>(c);
-      if (!l->tasks_.empty()) {
-        return false;
+      for (const auto task : l->tasks_) {
```
melvinw: fixed

barath:
Just a minor note -- I agree about the use of const auto in the loop (that the explicit type is better in general) but didn't comment on it because we are likely to change the type soon...

barath:
image

PR #185, opened 2016-12-15
Sanity check script

justinemarie:
The idea would be to run something like this in Travis so we don't have any more edits that completely bork modules and configurations end to end.

### DIFF
--- a/sanity_check.sh
+++ b/sanity_check.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+BESSCTL='./bessctl/bessctl'
+SCRIPTS='./bessctl/conf/samples'
+OUTFILE='sanity_check.out'
+
+TESTS='exactmatch.bess
+	flowgen.bess
+	generic_encap.bess
+	hash_lb.bess
+	igate.bess
+	iplookup.bess
+	l2_forward.bess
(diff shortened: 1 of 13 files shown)


justinemarie:
Wait, where are these other commits from y'all coming from? I thought I hadn't pulled from anywhere other than develop...

changlan:
If the script is for Travis only, would it be better to put that inside .travis directory?

sanity_check.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+BESSCTL='./bessctl/bessctl'
+SCRIPTS='./bessctl/conf/samples'
+OUTFILE='sanity_check.out'
+
+TESTS='exactmatch.bess
```
melvinw: why not run everything in bessctl/conf/samples that doesn't require bringing up physical ports?

melvinw: (if these are the those scripts ignore me)

melvinw:
Merged manually. Thanks! 
pup

sangjinhan:
No dawgs!
giphy

justinemarie:
How do we go about integrating this into Travis?

PR #194, opened 2016-12-16
bessctl: Throw exception when daemon is not connected

changlan:
Previously command.py checks whether daemon is connected before running every command. I move this logic to bess.py and throw an exception when daemon is not connected. This approach aligns with other error handling in bessctl and reduces code verbosity.


sangjinhan:
Excellent. Thanks!
5052346178_562cea8216

PR #195, opened 2016-12-16
warn when spawning workers on the master core

melvinw:


core/dpdk.cc
```diff
@@ -155,4 +156,9 @@ static void init_eal(const char *prog_name, int mb_per_socket,
 void init_dpdk(const ::std::string &prog_name, int mb_per_socket,
                int multi_instance, bool no_huge) {
   init_eal(prog_name.c_str(), mb_per_socket, multi_instance, no_huge);
+  if (FLAGS_c == rte_lcore_index(rte_get_master_lcore())) {
+    LOG(WARNING) << "Specified default worker core (core " << FLAGS_c
```
barath: I wonder whether this is strong enough -- should we disallow it entirely?  @sangjinhan?

core/worker.cc
```diff
@@ -255,6 +255,11 @@ void *run_worker(void *_arg) {
 }
 
 void launch_worker(int wid, int core) {
+  if (core == rte_lcore_index(rte_get_master_lcore())) {
+    LOG(WARNING) << "Launching a worker on core (core " << core
```
barath: Same here.

sangjinhan:
The issue is a little bit complicated with grpc. In the old C code, there was one "control" (master) thread and N worker threads, and the only thing we needed to make sure that the master thread does run on one of the cores worker are running.

With grpc, there can be many control threads [1] that are pooled and currently we don't have any control (how many threads, what thread to use for a command, etc.) over them. 

[1] By the way, it is a bug to have more than one control thread. BESS code is not thread safe between control threads.

An ideal solution would be:
1. When started, the bessd process unpins itself from all "isolated cpus" (/proc/cmdline) if any, so that all non-control and non-worker threads be spawned on remaining cores, and
2. We limit the number of grpc control threads to 1, and
3. Whenever a worker thread is created/updated, we unpin the control thread from the worker core. (e.g., see reset_core_affinity in bessctl.cc)

PR #196, opened 2016-12-18
Sanity check working!!

justinemarie:
We finally have a set of sanity check, end-to-end scripts that work in Travis :-)


barath:
Great.  I'm seeing failures with gcc 6 -- were those there before?

justinemarie:
fml. debugging again :(

PR #202, opened 2016-12-23
modules: Add ACL module

changlan:
- Almost identical to the one in NetBricks
- `established` keyword is not yet implemented

### DIFF
--- a/core/modules/acl.cc
+++ b/core/modules/acl.cc
@@ -2,7 +2,6 @@
 
 #include <rte_ether.h>
 #include <rte_ip.h>
-#include <rte_lpm.h>
 #include <rte_udp.h>
 
 #include <string>
@@ -13,12 +12,13 @@ const Commands ACL::cmds = {
 
 pb_error_t ACL::Init(const bess::pb::ACLArg &arg) {
   for (const auto &rule : arg.rules()) {
-    ACLRule new_rule = {.src_ip = CIDRNetwork(rule.src_ip()),
-                        .dst_ip = CIDRNetwork(rule.dst_ip()),
-                        .src_port = static_cast<uint16_t>(rule.src_port()),
-                        .dst_port = static_cast<uint16_t>(rule.dst_port()),
-                        .established = rule.established(),
-                        .drop = rule.drop()};
+    ACLRule new_rule = {
+        .src_ip = CIDRNetwork(rule.src_ip()),
+        .dst_ip = CIDRNetwork(rule.dst_ip()),
+        .src_port = htons(static_cast<uint16_t>(rule.src_port())),
+        .dst_port = htons(static_cast<uint16_t>(rule.dst_port())),
+        .established = rule.established(),
+        .drop = rule.drop()};
     rules_.push_back(new_rule);
   }
   return pb_errno(0);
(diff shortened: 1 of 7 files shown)


core/modules/acl.cc
```diff
@@ -0,0 +1,65 @@
...
+
+void ACL::ProcessBatch(bess::PacketBatch *batch) {
+  gate_idx_t out_gates[bess::PacketBatch::kMaxBurst];
+  int cnt = batch->cnt();
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+
+    struct ether_hdr *eth = pkt->head_data<struct ether_hdr *>();
+    struct ipv4_hdr *ip = reinterpret_cast<struct ipv4_hdr *>(eth + 1);
+    struct udp_hdr *udp = reinterpret_cast<struct udp_hdr *>(ip + 1);
```
barath: Do we need a check for options here?  If there are options, then ip+1 won't point at the UDP header.

core/utils/ether.h
```diff
@@ -27,7 +27,7 @@ struct [[gnu::packed]] EthHeader {
 
     void Randomize();
 
-    bool operator == (const Address &o) const {
+    bool operator==(const Address &o) const {
```
barath: Given that it's a small value, would it be faster to treat it as an int and compare the ints?

core/utils/ether.h
```diff
@@ -36,7 +36,7 @@ struct [[gnu::packed]] EthHeader {
       return true;
     }
 
-    bool operator !=(const Address &o) const {
+    bool operator!=(const Address &o) const {
```
barath: Why is != needed?

core/utils/ip.h
```diff
@@ -0,0 +1,24 @@
...
+namespace bess {
+namespace utils {
+
+typedef uint32_t IPAddress;
+
+struct CIDRNetwork {
+  CIDRNetwork() = default;
+  explicit CIDRNetwork(const std::string& cidr);
+
+  bool Match(const IPAddress& ip) const { return (addr & mask) == (ip & mask); }
```
barath: Method comment.

core/utils/ip.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <string>
+
+namespace bess {
+namespace utils {
+
+typedef uint32_t IPAddress;
+
+struct CIDRNetwork {
+  CIDRNetwork() = default;
```
barath: Comments for constructors.

core/modules/acl.cc
```diff
@@ -0,0 +1,65 @@
...
+
+    IPAddress src_ip = ntohl(ip->src_addr);
```
barath: Would it be faster overall to keep everything network byte ordered?  Then these conversions don't have to happen.

core/utils/ip_test.cc
```diff
@@ -0,0 +1,37 @@
+#include "ip.h"
+
+#include <gtest/gtest.h>
+
+namespace {
+
+using bess::utils::CIDRNetwork;
+
+TEST(IPTest, CIDRInStr) {
```
barath: Test function comments.

protobuf/module_msg.proto
```diff
@@ -131,6 +131,18 @@ message WildcardMatchCommandSetDefaultGateArg {
   uint64 gate = 1;
 }
 
+message ACLArg {
```
barath: Protobuf message/field comments.

changlan:
Fixed. There weren't changes to `ether.h` in this PR except formatting.

protobuf/module_msg.proto
```diff
@@ -133,12 +133,12 @@ message WildcardMatchCommandSetDefaultGateArg {
 
 message ACLArg {
   message Rule {
-    string src_ip = 1;
-    string dst_ip = 2;
-    uint32 src_port = 3;
-    uint32 dst_port = 4;
-    bool established = 5;
-    bool drop = 6;
+    string src_ip = 1;    // Source IP: "192.168.0.1". Wildcard if "".
```
barath: Is the given IP an example?

barath:
Just made this...

out

PR #205, opened 2016-12-29
Remove memcpy_sloppy

apanda:
As seen in 623771a9 memcpy_sloppy can cause some issues, and it is only used in two modules with uncertain benefits when
compared to rte_memcpy or other memcpy. This commit eliminates the function and all callers.


barath:
image

PR #207, opened 2016-12-30
Disable -Ofast for g++-6 and enable sanity_check for travis.

justinemarie:
So, turns out g++6 + O3/Ofast makes weird and untraceable things happen that I can't figure out. Things crash in BESS when they're enabled (particularly queue.bess) so I changed the Makefile to use -O2 if CXX is set to g++-6. Now that this works, Travis can run sanity_check.sh and all tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ before_install:
 
 install:
   - sudo apt-get install -y python
+  - pip install grpcio scapy
   - ln -s /build/dpdk-16.07 deps
   - docker pull $IMAGE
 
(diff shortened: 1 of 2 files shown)


sangjinhan:
We should fix issues when they are discovered, not hide them. Heisenbugs that appear with different compiler optimization levels are a great indicator that there are some obvious problems in our code. Working around with lower optimization levels would be acceptable only when we confirm that there are -O3 bugs in g++ 6, but it is very unlikely.

Please submit an issue with the crash log messages when the test fails.

Issue #208, opened 2016-12-30
queue.bess can't run with g++6 and -Ofast/-O3 are enabled.


justinemarie:
When you run queue.bess with g++6 and -Ofast it crashes.

Here's what you see in GDB:
Thread 20 "bessd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffedb27700 (LWP 41252)]
0x000000000062990e in __llring_mp_do_enqueue (behavior=LLRING_QUEUE_VARIABLE, n=32, obj_table=<optimized out>, r=0x7fffe001a5f0) at modules/../kmod/llring.h:433
433	modules/../kmod/llring.h: No such file or directory.

And the backtrace:
#0  0x000000000062990e in __llring_mp_do_enqueue (behavior=LLRING_QUEUE_VARIABLE, n=32, obj_table=<optimized out>, r=0x7fffe001a5f0) at modules/../kmod/llring.h:433
#1  llring_mp_enqueue_burst (n=32, obj_table=<optimized out>, r=0x7fffe001a5f0) at modules/../kmod/llring.h:1009
#2  Queue::ProcessBatch (this=<optimized out>, batch=0x7fffed916770) at modules/queue.cc:109
#3  0x000000000063a6c6 in Module::RunChooseModule (batch=0x7fffed916770, ogate_idx=0, this=<optimized out>) at modules/../module.h:321
#4  Module::RunNextModule (batch=0x7fffed916770, this=<optimized out>) at modules/../module.h:325
#5  Source::RunTask (this=<optimized out>) at modules/source.cc:83
#6  0x00000000005fa085 in Task::Scheduled (this=<optimized out>) at task.cc:36
#7  0x00000000005f8c0e in bess::LeafTrafficClass::RunTasks (this=0x7fffe0026620) at traffic_class.h:440
#8  bess::Scheduler::ScheduleOnce (this=<optimized out>) at scheduler.h:69
#9  bess::Scheduler::ScheduleLoop (this=0x7fffe4210c10) at scheduler.cc:30
#10 0x00000000005fd9e6 in Worker::Run (this=0x7fffed917340, _arg=_arg@entry=0x7ffff64a84e0) at worker.cc:241
#11 0x00000000005fe063 in run_worker (_arg=0x7ffff64a84e0) at worker.cc:254
#12 0x0000000000b00faf in execute_native_thread_routine ()
[...]

PR #209, opened 2016-12-30
Fix memory alignment in Queue module

sangjinhan:
This PR addresses #208.

### DIFF
--- a/core/mem_alloc.cc
+++ b/core/mem_alloc.cc
@@ -24,7 +24,9 @@ void *mem_alloc_ex(size_t size, size_t align, int) {
 
   ret = posix_memalign(&ptr, align, size);
   if (ret)
-    return NULL;
+    return nullptr;
+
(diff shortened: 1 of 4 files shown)


justinemarie:
Can you add sanity_check.sh to the .travis.yml file so Travis will show that this actually fixes the bug?

justinemarie:
(You will also need to add 'pip install grpcio scapy' to the travis file)

justinemarie:
Also, woohoo! Thanks @sangjinhan and @apanda!

sangjinhan:
@justinemarie I think you can download the PR branch to your local repository, like `git fetch origin fix_alignment` (not sure about the exact command)

justinemarie:
I did (and it passed! yay!). I can make another PR to make sure that script gets called in Travis if you think that's better?

justinemarie:
Obligatory cats...
image

PR #210, opened 2016-12-30
Add sanity check to Travis.

justinemarie:
It should pass now :-)


sangjinhan:
The PR includes non-related commits. Please rebase.

Perhaps sanity_check.out needs to be added to .gitignore..?

PR #213, opened 2016-12-31
Minor updates to container build

sangjinhan:


.travis/Dockerfile
```diff
@@ -26,10 +26,10 @@ RUN cd /build && \
 	git clone -b $(curl -s -L http://grpc.io/release) https://github.com/grpc/grpc && \
 	cd /build/grpc && \
 	git submodule update --init && \
-	make -j && \
+	make -j4 && \
```
barath: Quick question on this -- have you run into VM hanging issues with higher levels of parallelism in builds?  I seem to get hangs when going above j2 (it's nondeterministic).

sangjinhan: I think if a VM has 4GB of memory and 2GB out of 4 is dedicated to hugepages, the actual amount of available memory could be very small. Not sure how big a typical g++ process would take, but excessive swapping might happen even with 2 processes.

sangjinhan: I wouldn't worry about this Dockerfile though, since we can assume the Docker image would be built (by us) on a reasonably beefy machine.

barath:
img_0262

PR #216, opened 2017-01-03
Minor updates to gRPC client and server

sangjinhan:
It addresses the following issues in the current code:

- 145f434: if "daemon start" fails for whatever reason, bessctl gets stuck in an infinite loop
- 050e2ed: SetupControl returns before the listening socket is ready. As a result, bessctl's "daemon start" and bessd may race.
- 95c0343: when bessctl makes an initial connection to bessd, a low-level socket is used to probe its availability, not with grpc
- 1e09387: even if grpcio python module is not available, no explicit error message is shown


changlan:
Excellent!
image

PR #218, opened 2017-01-06
Add -fPIC to DPDK cflags

apanda:
The Google testing infrastructure (and NetBricks when that becomes useful) link to some of the static libraries in
.so. Without passing in -fPIC to DPDK this causes build problems on some machines (I had it on a machine with gcc
version 5.4.0 20160609 (Debian 5.4.0-4) LD 2.26 and Linux 4.6.0), which is fixed by passing in -fPIC. Additionally this
should have no impact on the binaries.


sangjinhan:
Thanks!
i13901531291

PR #219, opened 2017-01-06
Adds a utility class for reconstructing a TCP flow.

barath:


core/utils/tcp_flow_reconstruct.h
```diff
@@ -0,0 +1,114 @@
...
+  // the packet is a SYN then we use the SYN to set the initial sequence number
+  // offset.
+  //
+  // Assumes that the IP header has no options and that the first call is the
+  // SYN.  Behavior is undefined if the first packet is not the SYN or the
+  // packet is no a TCP packet.
+  void InsertPacket(Packet *p) {
+    struct ether_hdr *eth = p->head_data<struct ether_hdr *>();
+    struct ipv4_hdr *ip = (struct ipv4_hdr *)(eth + 1);
+    struct tcphdr *tcp = (struct tcphdr *)(ip + 1);
```
changlan: What if there are IP options?

barath: Do you want support for it?  Right now the comments explicitly say that it is not supported.

core/utils/tcp_flow_reconstruct.h
```diff
@@ -0,0 +1,114 @@
...
+
+  // Adds the data of the given packet based upon its TCP sequence number.  If
+  // the packet is a SYN then we use the SYN to set the initial sequence number
+  // offset.
+  //
+  // Assumes that the IP header has no options and that the first call is the
+  // SYN.  Behavior is undefined if the first packet is not the SYN or the
+  // packet is no a TCP packet.
+  void InsertPacket(Packet *p) {
+    struct ether_hdr *eth = p->head_data<struct ether_hdr *>();
```
sangjinhan: Consider using `struct EthHeader (from utils/ether.h)`. You may want to define structs for IP and TCP headers as well.

core/utils/tcp_flow_reconstruct.h
changlan: I see. If it is not too difficult, it would be better to support it though, since it should be a generic utility class (rather than solely for the milestone demo).

core/utils/tcp_flow_reconstruct.h
barath: Ok.  We should change other uses in the BESS codebase to use this header as well, then.  struct ether_hdr from DPDK is used in other modules (e.g., ip lookup).

core/utils/tcp_flow_reconstruct.h
barath: Ok, sure.

nrdmao: Generally it is best to use the IP header length in the IP header to find the start of the transport header.

barath: Yes, changing this shortly.

changlan:
This PR disappeared?

barath:
Not sure what happened.  I will open a new one shortly with a workaround
for the gtest / gcc bug so that travis stops complaining.

PR #220, opened 2017-01-07
Add NAT and IPSwap

changlan:


bessctl/conf/samples/nat.bess
```diff
@@ -0,0 +1,28 @@
+import scapy.all as scapy
```
barath: Would you be able to add comments both to the head of this script and to a few of the commands that are not commented explaining their intent / function?

core/modules/ipswap.h
```diff
@@ -0,0 +1,11 @@
+#ifndef BESS_MODULES_IPSWAP_H_
+#define BESS_MODULES_IPSWAP_H_
+
+#include "../module.h"
+
+class IPSwap final : public Module {
```
barath: Could you add a class comment explaining what this module is supposed to do?

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+
+typedef std::pair<CIDRNetwork, CIDRNetwork> NATRule;
+
+template <class T>
+inline void hash_combine(std::size_t &seed, const T &v) {
+  std::hash<T> hasher;
+  seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+}
+
+class NAT final : public Module {
```
barath: Could you add a class comment?

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+ public:
+  struct Flow {
```
barath: Is there a reason struct Flow is declared inside the NAT class?  Also, Since it has a non-trivial method, perhaps it should be declared as a Class and add a comment?

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+    uint8_t proto;
+    IPAddress src_ip;
+    uint16_t src_port;
+    IPAddress dst_ip;
+    uint16_t dst_port;
+    uint16_t icmp_ident;
+
+    Flow reverse_flow() const {
```
barath: Method comment?  Also, you might rename this as ReverseFlow()

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+
+class NAT final : public Module {
+ public:
+  struct Flow {
+    uint8_t proto;
+    IPAddress src_ip;
+    uint16_t src_port;
+    IPAddress dst_ip;
+    uint16_t dst_port;
+    uint16_t icmp_ident;
```
barath: Could you add a comment saying what icmp_ident is?

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+      hash_combine(seed, f.src_ip);
+      hash_combine(seed, f.src_port);
+      hash_combine(seed, f.dst_ip);
+      hash_combine(seed, f.dst_port);
+      hash_combine(seed, f.icmp_ident);
+      return seed;
+    }
+  };
+
+  struct FlowRecord {
```
barath: Comment.

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+    uint16_t port;
+    Flow internal_flow;
+    Flow external_flow;
+    uint64_t time;
+    FlowRecord() : port(0), time(0) {}
```
barath: Since you have a constructor, perhaps rename as a class?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+  struct udp_hdr *udp = reinterpret_cast<struct udp_hdr *>(l4);
+  struct icmp_hdr *icmp = reinterpret_cast<struct icmp_hdr *>(l4);
+  NAT::Flow flow;
+
+  flow.proto = ip->next_proto_id;
+  flow.src_ip = ip->src_addr;
+  flow.dst_ip = ip->dst_addr;
+
+  switch (ip->next_proto_id) {
+    case 0x06:
```
barath: Could you make an enum for these values used in these two switch statements?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+  return pb_cmd_response_t();
+}
+
+inline static void compute_cksum(struct ipv4_hdr *ip, void *l4) {
+  struct tcp_hdr *tcp = reinterpret_cast<struct tcp_hdr *>(l4);
+  struct udp_hdr *udp = reinterpret_cast<struct udp_hdr *>(l4);
+  struct icmp_hdr *icmp = reinterpret_cast<struct icmp_hdr *>(l4);
+
+  ip->hdr_checksum = 0;
+  switch (ip->next_proto_id) {
```
barath: Could you make an enum for these case values?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+  available_ports_.resize(MAX_PORT - MIN_PORT + 1);
+  std::iota(available_ports_.begin(), available_ports_.end(), MIN_PORT);
+  std::random_shuffle(available_ports_.begin(), available_ports_.end());
+
+  flow_vec_.resize(MAX_PORT - MIN_PORT + 1);
+
+  return pb_cmd_response_t();
+}
+
+inline static void compute_cksum(struct ipv4_hdr *ip, void *l4) {
```
barath: Function comment?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+          VLOG(1) << "Unknown icmp_type: " << icmp->icmp_type;
+      }
+      break;
+    default:
+      VLOG(1) << "Unknown next_proto_id: " << ip->next_proto_id;
+  }
+  return flow;
+}
+
+inline static void stamp_flow(struct ipv4_hdr *ip, void *l4,
```
barath: Function comment?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+                              const NAT::Flow &flow) {
+  struct udp_hdr *udp = reinterpret_cast<struct udp_hdr *>(l4);
+  struct icmp_hdr *icmp = reinterpret_cast<struct icmp_hdr *>(l4);
+
+  ip->src_addr = flow.src_ip;
+  ip->dst_addr = flow.dst_ip;
+
+  switch (flow.proto) {
+    case 0x06:
```
barath: Enum for case values.

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+      icmp->icmp_cksum = 0;
+      icmp->icmp_cksum = rte_ipv4_udptcp_cksum(ip, icmp);
+      break;
+    default:
+      VLOG(1) << "Unknown next_proto_id: " << ip->next_proto_id;
+  }
+  ip->hdr_checksum = rte_ipv4_cksum(ip);
+}
+
+inline static NAT::Flow parse_flow(struct ipv4_hdr *ip, void *l4) {
```
barath: Function comment?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+  int cnt = batch->cnt();
+
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+
+    // By default, pass packet
+    out_gates[i] = incoming_gate;
+
+    struct ether_hdr *eth = pkt->head_data<struct ether_hdr *>();
+    struct ipv4_hdr *ip = reinterpret_cast<struct ipv4_hdr *>(eth + 1);
```
barath: If you would like, you can now use ether.h, ip.h declarations.

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+    int ip_bytes = (ip->version_ihl & 0xf) << 2;
+    struct udp_hdr *udp = reinterpret_cast<struct udp_hdr *>(
```
barath: We don't yet have a udp header of our own, but you could add one if you'd like.

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+        reinterpret_cast<uint8_t *>(ip) + ip_bytes);
+
+    // L4 protocol must be TCP, UDP, or ICMP
```
barath: It says TCP, UDP, or ICMP in the comment, but the code appears to only handle UDP since it assumes that it's a UDP header.  Do you plan to make it general for TCP/ICMP or am I misreading it?

core/modules/nat.cc
```diff
@@ -0,0 +1,239 @@
...
+    const auto rule_it = std::find_if(
+        rules_.begin(), rules_.end(),
+        [&ip](const NATRule &rule) { return rule.first.Match(ip->src_addr); });
+    if (rule_it == rules_.end()) {
+      // No rules found for this source IP address, drop.
+      out_gates[i] = DROP_GATE;
+      continue;
+    }
+
+    uint16_t new_port = available_ports_.back();
```
barath: Might add a TODO here about port randomization.

core/modules/nat.h
```diff
@@ -0,0 +1,98 @@
...
+
+ private:
+  IPAddress RandomIP(const CIDRNetwork &net) {
+    uint32_t min = ntohl(net.addr & net.mask);
+    uint32_t max = ntohl(net.addr | (~net.mask));
+    return htonl(rng_.GetRange(max - min) + min);
+  }
+
+  std::vector<NATRule> rules_;
+  std::vector<uint16_t> available_ports_;
```
barath: Just to clarify -- aren't the available ports not shared by the entire NAT by instead per-external-IP?  If that's the case, should this be a map from external IP to available ports for that IP?

core/modules/nat.cc
changlan: It is already randomized.

core/modules/nat.cc
changlan: It is already handling TCP/ICMP.

core/modules/nat.cc
barath: It is randomized at the start, I agree, but if I'm reading it right, do the ports get randomized upon re-use?

core/modules/nat.cc
barath: Ah, ok -- just to be clear.  For TCP you are relying on the fact that you only look at the first 32 bits, which are the same for both TCP and UDP?  You might add a comment clarifying that if so.

changlan:
Convince me why it is needed?

barath:
See REQ-15 of the RFC:

https://tools.ietf.org/html/rfc6888

But if you think it's not needed, then that's fine, but we will need to
change the description in the description doc.

changlan:
I agree that "A CGN's port allocation scheme SHOULD make it hard for attackers to guess port numbers", but my real point is, if the initial allocation is random, and the sequence of reused ports is (almost) random, the resulting allocation should be (almost) random. So it is not clear to me that why we need to do extra randomization upon reuse.

barath:
Ok, that's fine.  Let me know about the other changes and I can approve after.

core/modules/nat.cc
changlan: Actually -- no, it is just a pointer to the L4 header. None of the code below actually use the struct. I updated the code to use `void *` instead.

changlan:
Done.

barath:
img_0020

PR #221, opened 2017-01-07
Adds a utility class for reconstructing a TCP flow.

barath:


core/utils/ip.h
```diff
@@ -1,8 +1,72 @@
...
+
+/*
+ * Definitions for internet protocol version 4.
+ * Per RFC 791, September 1981.
+ */
+
+/*
+ * Structure of an internet header, naked of options.
+ */
+struct [[gnu::packed]] ip
```
changlan: `Ipv4Header`? (to be consistent with `EthHeader`)

core/utils/ip.h
```diff
@@ -1,8 +1,72 @@
...
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+    unsigned int ip_hl:4;		/* header length */
+    unsigned int ip_v:4;		/* version */
+#endif
+#if __BYTE_ORDER == __BIG_ENDIAN
+    unsigned int ip_v:4;		/* version */
+    unsigned int ip_hl:4;		/* header length */
+#endif
+    u_int8_t ip_tos;			/* type of service */
+    u_short ip_len;			/* total length */
```
changlan: This might be my personal preference -- I don't like the BSD-style definition very much, for the following reasons:

- It is 21st century :)
- We should use int type names with explicit length (e.g. `uint8_t`, `uint16_t`), not `u_short`, `unsigned int`, etc.
- `struct in_addr` is unnecessary.
- The prefix `ip_*` in field names are redundant.

The style in `rte_ip.h` (http://dpdk.org/doc/api/rte__ip_8h_source.html) looks much better.

core/utils/tcp.h
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+#ifndef BESS_UTILS_TCP_H_
+
+typedef	u_int32_t tcp_seq;
+/*
+ * TCP header.
+ * Per RFC 793, September, 1981.
+ */
+struct [[gnu::packed]] tcphdr
```
changlan: TcpHeader

core/utils/tcp.h
```diff
@@ -0,0 +1,106 @@
...
+  {
```
changlan: Same comment about the style.

barath:
These aren't just BSD-style -- they are BSD.  The headers are intentionally
unmodified BSD headers, and I only pulled them out so that we can mark them
as packed, to avoid some gcc alignment complaints.  I marked at the top of
the headers their origin.

The DPDK headers have other issues (they don't properly mark the subfields
within single-byte fields).

changlan:
Yes, I knew it was pulled from BSD, so my comments weren't directed to your style. But the fact that the code is from BSD doesn't mean the code is not questionable or modifiable. I don't see why they should be kept intentionally unmodified. My point was that, now that they are part of BESS codebase, and given the issues I listed, we should make it conform to our code style.

sangjinhan:
A 3rd-party module may want to use other sources, rte_*.h, /usr/include/netinet/*.h, etc., for protocol header definitions. To minimize possibility of namespace collision between the well-known types (in C with all lowercase names) and ours, I think using CamelCase is a good idea here.

barath:
Already changed...hit reload :)

changlan:
Merged, thanks!

image

PR #222, opened 2017-01-07
Fix "AssertionError: Output gate is already bound"

sangjinhan:
If a sugared bessctl script contains a module with multiple
gate numbers, e.g., `foo -> bar:2 -> 1:bar:3 -> baz`
(bar is bound with an output gate twice, 2 and 3),
"Output/Input gate is already bound" assertion failure happens.
This patch addresses the issue by treating bar:2 and 2:bar:3
as different Python object.

### DIFF
--- a/bessctl/module.py
+++ b/bessctl/module.py
@@ -1,5 +1,5 @@
+import copy
 import types
-
 
 def _callback_factory(self, cmd, arg_type):
     return lambda mod, arg=None, **kwargs: \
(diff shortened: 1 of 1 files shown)


bessctl/module.py
```diff
@@ -44,10 +44,12 @@ def __mul__(self, ogate):
             assert False, 'Gate ID must be an integer'
 
         if self.ogate is not None:
+            print self, self.ogate, ogate
```
changlan: Quick question: is this print statement intended?

bessctl/module.py
```diff
@@ -1,6 +1,6 @@
+import copy
 import types
 
-
```
changlan: Really minor, feel free to ignore: PEP-8 suggests two blank lines

bessctl/module.py
sangjinhan: Updated :D

changlan:
I fixed the issue since it's been pending for a while :D

image

PR #225, opened 2017-01-10
Fix bugs in .travis.yml

changlan:
- Upload codecov report when DEBUG=1
- Run unit test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
 script:
   - ./container_build.py bess
   - ./container_build.py kmod_buildtest
-  - cd core; ./all_test; cd ..
+  - ./core/all_test
   - ./sanity_check.sh
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
 script:
   - ./container_build.py bess
   - ./container_build.py kmod_buildtest
-  - ./core/all_test
+  - cd core && ./all_test && cd ..
   - ./sanity_check.sh
 
 after_success:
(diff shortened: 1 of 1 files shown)


justinemarie:
I'm a bit confused:  How did the repo end up in this state? Did we merge a commit that didn't pass Travis?

justinemarie:
Also, shouldn't these commits be squashed into one?

changlan:
Just a side note: You can squash and merge on GitHub if you really care.

justinemarie:
Ah, cool -- I forgot that they'd added that feature. Still, what's the backstory on the bug?

justinemarie:
I figured it out -- Barath's patch failed in all_test, but passed travis because there are multiple commands on one line. See:

justine@Justines-MacBook-Pro:/$ cd ..; `exit 2`; cd .
justine@Justines-MacBook-Pro:/$ echo $?
0

So the fix is both to add ./all_test and make it 
cd .. && ./all_test && cd . 
That way is the test fails travis will notice.

justinemarie:
@barath -- why do we need the cd rather than doing core/all_test?

barath:
Good catch!

As for why we need cd -- it's because we need to access tracefiles for the
tcp reconstruct unit test and those are specified with a relative path, and
the relative path that we run things from is different in different
environments.  I didn't see an easy way to specify an absolute path that
would work well both locally and in travis, so changing the PWD via cd
seemed easiest.

justinemarie:
Ah okay, so then we can't revert to ./all_tests, but we should do the && solution?

barath:
Sounds good.

justinemarie:
I restarted the travis job? Looks like a passing error during apt-get update.

barath:
img_0792

PR #226, opened 2017-01-11
nat: avoid unecessary gc attempts

melvinw:
I think we'll see a non-negligible performance hit as the number of flows entering the NAT increases beyond `MAX_PORT - MIN_PORT` by GCing the current way. We can avoid doing unnecessary work by calculating the earliest we can expect there to be a free port the first time we enter the GC section of the code. Then future attempts to GC will only happen that point or later, when we know we'll find something.

### DIFF
--- a/core/modules/nat.cc
+++ b/core/modules/nat.cc
@@ -2,6 +2,7 @@
 
 #include <rte_ip.h>
 
+#include <algorithm>
 #include <numeric>
 #include <string>
 
@@ -215,6 +216,7 @@ void NAT::ProcessBatch(bess::PacketBatch *batch) {
           flow_hash_.erase(hash_it);
           flow_hash_.erase(hash_it->second.internal_flow);
         }
+        next_expiry_ = now;
       }
     }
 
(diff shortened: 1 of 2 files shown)


core/modules/nat.cc
```diff
@@ -234,13 +237,15 @@ void NAT::ProcessBatch(bess::PacketBatch *batch) {
...
-    if (available_ports_.empty()) {
+    if (available_ports_.empty() && now >= next_expiry_) {
       for (auto &record : flow_vec_) {
         if (record.time != 0 && now - record.time >= TIME_OUT) {
           available_ports_.push_back(record.port);
           record.time = 0;
           flow_hash_.erase(record.internal_flow);
           flow_hash_.erase(record.external_flow.ReverseFlow());
+        } else if (record.time != 0) {
+          next_expiry_ = std::min(next_expiry_, record.time + TIME_OUT);
```
barath: If I'm reading this right, won't the old next_expiry_ be the min always?  Do we need to set it to int max at the beginning of this if?

melvinw: Ah, you're right. Good catch

changlan:
Merged, thanks!
image

PR #227, opened 2017-01-13
Add UrlFilter module

changlan:


core/modules/url_filter.cc
```diff
@@ -0,0 +1,279 @@
...
+    Flow flow;
+    flow.src_ip = ip->src;
+    flow.dst_ip = ip->dst;
+    flow.src_port = tcp->src_port;
+    flow.src_port = tcp->dst_port;
+
+    // Check if the flow is already blocked
+    std::unordered_map<Flow, uint64_t, FlowHash>::iterator it;
+    if ((it = blocked_flows_.find(flow)) != blocked_flows_.end()) {
+      uint64_t now = tsc_to_ns(rdtsc());
```
sangjinhan: `rdtsc()` is an expensive operation. You can use `ctx.current_ns()`, which is refreshed for every scheduling round, in place of `tsc_to_ns(rdtsc())`

core/modules/url_filter.cc
```diff
@@ -0,0 +1,279 @@
...
+
+#include "../utils/ether.h"
+#include "../utils/http_parser.h"
+#include "../utils/ip.h"
+
+using bess::utils::EthHeader;
+using bess::utils::Ipv4Header;
+using bess::utils::TcpHeader;
+
+const uint64_t TIME_OUT = 12e10;
```
sangjinhan: `TIME_OUT_NS` might be better?
Also, if you intended 2 minutes, 120e9 would be a better representation

changlan:
Sounds good. Will change.

core/modules/url_filter.cc
sangjinhan: ... or 120 * 1000 * 1000 * 1000. 120e9 is a floating point number so it may be slightly off after type casting.

changlan: Got the "integer overflow in expression" error. Will use 10000000000 instead..

changlan: Oh I could do `120L * 1000 * 1000 * 1000`. Nevermind.

sangjinhan:
Merged, thanks!
i15573941050

PR #228, opened 2017-01-13
CLI.err() should not terminate process immediately

sangjinhan:
Instead, the current command need be gracefully terminated with further
actions properly performed (additional error messages, exception handling, etc.)


changlan:
Good catch!
image

justinemarie:
Wait wait wait -- this needs to return an error code! If we exit due to error we need to return an error code, otherwise our testing scripts break.

PR #229, opened 2017-01-13
Do not recalculate checksum in IPSwap module

sangjinhan:
Swapping IP src/dst addresses and L4 port numbers affects neither IP
nor L4 checksums, since all changes does not cross 2-byte alignment
boundary and checksum calculation is commutative.

With the `samples/nat.bess` example, it reduced the cost of IPSwap from 7.8% to 3.8% on my machine.


changlan:
Merged, thanks!
image

PR #231, opened 2017-01-13
Minor changes to NAT

changlan:
- Replace `rdtsc()` with `ctx.current_ns()`
- Use `emplace()` instead of `insert()`
- Drop packets if there is no available port after GC


sangjinhan:
Thanks!
i13788116986

Issue #230, opened 2017-01-13
Change to error handling breaks testing.


justinemarie:
https://github.com/NetSys/bess/commit/5b2e0ce6a1ee480b272c18653e4f4773254f1b18

This commit breaks our testing scripts, because bessctl will no longer return an error code when it exits. This either needs to be patched or rolled back.

justinemarie:
Thanks for the patch, Sangjin!

PR #232, opened 2017-01-15
If bessctl loop ended with an error, sys.exit(1)

sangjinhan:
This patch addresses Issue #230.
Also, added ferr output stream, which was separated from fout


justinemarie:
This is great! Thank you!

justinemarie:
image
This is a photo of my sister, with our family cat, dressed as the cat for Halloween.

PR #236, opened 2017-01-17
Fix overflow PacketBatch and dangling packets

changlan:


justinemarie:
I've read Chang's patch and Chang's read mine -- but can we get a third set of eyes?

sangjinhan:
Why send RST first, then 403? 403 should arrive earlier to the source.

changlan:
Good catch!

sangjinhan:
What I usually do is:
```
loop {
  ...
  batch.add(pkt);
  if (batch.full()) {
    RunNextModule(batch);
    batch.clear();
  }
  ...
}

But this is just a style issue as performance doesn't really matter, especially given that URL filtering action would occur not very frequently.

**Comment by justinemarie** [comment]
@sangjinhan Wouldn't that cause higher latency variability?

**Comment by changlan** [comment]
Good point. I'd also prefer this style when I only need to add one packet in each iteration. When we need to inject multiple packets (as in this example), the code would become more verbose:

```
loop {
  ...
  batch.add(pkt);
  if (batch.full()) {
    RunNextModule(batch);
    batch.clear();
  }
  batch.add(pkt2);
  if (batch.full()) {
    RunNextModule(batch);
    batch.clear();
  }
  batch.add(pkt3);
  if (batch.full()) {
    RunNextModule(batch);
    batch.clear();
  }
  ...
}
```

But thanks for pointing out the idiomatic usage though!

sangjinhan:
Thanks!
tumblr_lg4lgyblpl1qfyzelo1_1280

PR #239, opened 2017-01-17
Changes hash function for NAT flows.

barath:


core/modules/nat.h
```diff
@@ -17,21 +17,16 @@ using bess::utils::CIDRNetwork;
...
 
-template <class T>
-inline void hash_combine(std::size_t &seed, const T &v) {
-  std::hash<T> hasher;
-  seed ^= hasher(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
-}
-
 // 5 tuple for TCP/UDP packets with an additional icmp_ident for ICMP query pkts
-class Flow {
+class [[gnu::packed]] Flow {
```
sangjinhan: This class doesn't need to be packed (It affects performance, since the compiler cannot make any assumption of the starting address of the object). If you just want to limit the size, just reorder fields in an aligned manner:
```
IPAddress src_ip;		    IPAddress src_ip;
IPAddress dst_ip;		    IPAddress dst_ip;
uint16_t src_port;		    uint16_t src_port;
uint16_t dst_port;		    uint16_t dst_port;
uint16_t icmp_ident;  // identifier of ICMP query
uint8_t proto;
```

barath:
Changed.  It appears to have caused no difference in performance.

changlan:
image

PR #241, opened 2017-01-18
Use CRC for flow hashing

changlan:
This leads to very minor performance improvement as shown by the microbenchmark.

### DIFF
--- a/core/modules/url_filter.h
+++ b/core/modules/url_filter.h
@@ -26,7 +26,7 @@ class Flow {
   uint16_t dst_port;
   uint32_t padding;
 
-  Flow() : padding(0) {}
+  Flow() : src_ip(0), dst_ip(0), src_port(0), dst_port(0), padding(0) {}
 
   bool operator==(const Flow &other) const {
     return src_ip == other.src_ip && src_port == other.src_port &&
--- a/core/modules/url_filter.h
+++ b/core/modules/url_filter.h
@@ -1,6 +1,9 @@
 #ifndef BESS_MODULES_URL_FILTER_H_
 #define BESS_MODULES_URL_FILTER_H_
 
+#include <rte_config.h>
+#include <rte_hash_crc.h>
+
 #include <map>
 #include <string>
 #include <utility>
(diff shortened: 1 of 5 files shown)


core/modules/url_filter.h
```diff
@@ -21,6 +24,9 @@ class Flow {
   uint32_t dst_ip;
   uint16_t src_port;
   uint16_t dst_port;
+  uint32_t padding;
+
+  Flow() : padding(0) {}
```
barath: I'm not sure you should add this constructor -- padding should be getting value-initialized automatically if you don't have a constructor.

sangjinhan: I might be wrong, but I think the default constructor for individual member fields with a primitive type is no-op... thus without an explicit class constructor the padding won't be zero-initialized by simply doing `Flow f;`.

barath: I think this example shows the case we're dealing with here, which would mean the other values are not getting zeroed:

http://en.cppreference.com/w/cpp/language/default_initialization


`    const T2 t2;      // const class, calls the user-provided default ctor
                      // t2.mem is default-initialized (to indeterminate value)`

barath: I was curious what Stroustroup says, so I just looked in his book, and he says, among other things:

"If you don't like the complications, just initialize consistently."

changlan: If I am not misunderstanding, you were making two different points here:
1. Remove the constructor because members should get zero-initialized: I think Sangjin is right that this is not true, so it is necessary to initialize them in ctor.
2. Initialize all members consistently in the ctor: This is fine to me. Also, this contradicts with the first point -- just to point it out.

barath: You're right, I think I was assuming value initialization (braces) and it's probably not being used that way elsewhere in the code.  So I was confused.

Perhaps we can settle on option 2) in general -- all classes should have constructors that explicitly list every variable in the initialize list -- either with a value or just () depending on whether the default is ok.

PR #244, opened 2017-01-19
Adds STL's unordered map to the hash table benchmark comparison.

barath:


changlan:
I'm curious what's the result?

barath:
Non-inlined HTable is about the same speed as unordered_map and about 4-5x as fast when inlined.  So I am looking at switching NAT to use HTable.

sangjinhan:
i13750240890

PR #246, opened 2017-01-19
Fix incorrect port statistics for i40e interfaces

sangjinhan:
`show port` / `monitor port` did not show statistics counters for i40e PMD devices. There were two issues:

* i40e doesn't support per-queue stats, but only port-wide stats. The PMD driver only used per-queue stats.
* While we do not need to accumulate counters with PMD drivers (since they have their own), it was doing that (thus wasting CPU cycles). This is regression of C++ porting.


justinemarie:
image

PR #247, opened 2017-01-19
Show fraction part for pkts/sched and cycles/pkt

sangjinhan:
Regression of C++ porting. They are useful for performance monitoring...


changlan:
image

PR #249, opened 2017-01-19
Add a sample script for UrlFilter and fix up pktgen_http

changlan:


justinemarie:
@changlan Can the urlfilter keep up with that rate specified in the sample script? the number of flows looks too high.

changlan:
Will test with current bug fixes. What rate did you use for benchmarking?

sangjinhan:
tumblr_m1icqnuro41qenqklo1_1280

PR #254, opened 2017-01-19
Update packetgen to balance load when in multicore and to rewrite both

justinemarie:
src and dst IPs.


sangjinhan:
The new script generates `NUM_FLOWS^2` flows, not just NUM_FLOWS...

PR #255, opened 2017-01-19
Cleans up a flow cache entry upon receipt of a FIN.

barath:


melvinw:
Thanks! 
wtfcat

core/modules/url_filter.cc
```diff
@@ -277,6 +277,10 @@ void UrlFilter::ProcessBatch(bess::PacketBatch *batch) {
           eth->dst_addr, eth->src_addr, ip->dst, ip->src, tcp->dst_port,
           tcp->src_port, tcp->ack_num + strlen(HTTP_403_BODY), tcp->seq_num));
     }
+
+    if (tcp->flags & TCP_FLAG_FIN) {
+      flow_cache_.erase(flow);
```
changlan: `flow_cache_.erase(flow)` may have been called once already if `matched` is `true`. I will submit another patch.

barath: That's ok -- erase is then a nop.

PR #257, opened 2017-01-19
Disallow negative buf_offset

changlan:
- Disallow negative buf_offset
- Misc clang-format changes

This PR should fix the crashing behavior, but ultimately FlowGen should generate TCP flows with correct sequence numbers.


sangjinhan:
tumblr_m2sp5tde6f1qzex9io1_1280

PR #258, opened 2017-01-19
Use RunChooseModule in NAT

changlan:
`RunChooseModule` should be faster than `RunSplit`.


sangjinhan:
tumblr_mbvnf7cykj1qddbvio1_1280

PR #259, opened 2017-01-20
Adds explicit construction and destruction to Trie Nodes to avoid uni…

barath:
…nitialized

infinitely-recursing destruction in some cases.


sangjinhan:
Merged, thanks!
tumblr_m2ngf2e1e31qzyqubo1_1280

PR #260, opened 2017-01-20
Self rating flowgen.

justinemarie:
For benchmarking -- this change allows your script to re-route return traffic back in to the flowgen so that it can find out the throughput of the device under test. It then self-scales the sending rate to find how much the device can handle.

### DIFF
--- a/core/modules/flowgen.cc
+++ b/core/modules/flowgen.cc
@@ -11,15 +11,18 @@
 
 #define RETRY_NS 1000000ul /* 1 ms */
 
+const uint64_t BENCHMARK_UPDATE_INTERVAL_S = 10L;
+const uint64_t BENCHMARK_UPDATE_INTERVAL_NS = (BENCHMARK_UPDATE_INTERVAL_S * 1000 * 1000 * 1000);
+
 typedef std::pair<uint64_t, struct flow *> Event;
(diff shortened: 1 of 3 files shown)


justinemarie:
Fixed the bug making Travis crash -- failed to mark an overwritten function w/ override.

justinemarie:
Beep beep?

core/modules/flowgen.cc
```diff
@@ -400,6 +414,38 @@ struct task_result FlowGen::RunTask(void *) {
...
+
+        // be a little aggressive -- try to increase.
+        total_pps_ = 1.1 * interval_pps;
+
+        // At low line rates traffic becomes burstier; harder to measure.
+        if (total_pps_ <= 1e5) {
+          total_pps_ = 1e5;
+        }
+
+        // flow_rate_ = ratio * flow_rate_; //Scale flows proportionally
```
justinemarie: delete

core/modules/flowgen.h
```diff
@@ -8,6 +8,8 @@
 #include "../module_msg.pb.h"
 #include "../utils/random.h"
 
+//30 seconds * ms * micros * nanos
```
justinemarie: delete

PR #261, opened 2017-01-20
Fix Update module updating at an incorrect offset

sangjinhan:
it should update packets at `offset`, not `offset * sizeof(uint64_t)`


sangjinhan:
offset may be a negative number, while Packet::head_data() only takes an offset in uint16_t.

sangjinhan:
:cat:

PR #262, opened 2017-01-20
Use HTable in NAT

changlan:


changlan:
Do not merge until the performance improvement is proved.

barath:
img_0051

PR #263, opened 2017-01-20
Make flowgen support sequence numbers in TCP.

justinemarie:


core/modules/flowgen.cc
```diff
@@ -340,7 +346,7 @@ void FlowGen::GeneratePackets(bess::PacketBatch *batch) {
     uint64_t t;
     struct flow *f;
     bess::Packet *pkt;
-
+
```
melvinw: nite: unnecessary indentation

core/modules/flowgen.cc
```diff
@@ -85,6 +86,7 @@ inline struct flow *FlowGen::ScheduleFlow(uint64_t time_ns) {
   flows_free_.pop_front();
 
   f->first = 1;
+  f->next_seq_no = 12345;
```
melvinw: ...why this value?

core/modules/flowgen.cc
```diff
@@ -145,6 +147,7 @@ void FlowGen::PopulateInitialFlows() {
 
       /* overwrite with a emulated pre-existing flow */
       f->first = 0;
+      f->next_seq_no = 56789;
```
melvinw: same here

core/modules/flowgen.cc
justinemarie: It's recognizable for debugging purposes.

core/modules/flowgen.cc
justinemarie: Also same -- these packets here are part of the "initial" backfilled flows.

core/modules/flowgen.cc
melvinw: kk

PR #266, opened 2017-01-20
Align struct Flow with 16-byte boundary

sangjinhan:
30% performance boost was observed with samples/nat.bess


sangjinhan:
Another 30% boost with simplified Flow comparator....

changlan:
Merged, thanks!

image

PR #268, opened 2017-01-24
Add comment about ansible usage

justinemarie:
Add a quick comment about how to use ansible -- become-pass is needed to address the glitch Sangjin saw.

### DIFF
--- a/ansible-playbook.yml
+++ b/ansible-playbook.yml
@@ -2,7 +2,7 @@
 ## Note: This has only been tested on *fresh* Ubuntu 16.04 installations.
 ## Usage:
 ##   sudo apt-get install ansible
-##   ansible-playbook -K ansible-playbook.yml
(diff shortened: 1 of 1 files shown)


justinemarie:
@sangjinhan This last commit fixes the bug you found -- pip does not require sudo and so should not use become:true; if you try to give it a password it hangs because it's never prompting for sudo...

sangjinhan:
Let's add `sudo -H` to force pip to install packages to the global repository for all users.

sangjinhan:
Oh what I meant was `sudo -H pip install ...`, not the entire script. (=_=)

justinemarie:
@sangjinhan

justinemarie:
Can someone please merge this? The undergrads will want to check this out :-)

melvinw:
Thanks! 
catsun

PR #271, opened 2017-01-27
Resizing and specifying IP/portranges now works.

justinemarie:
THREE PRS IN ONE DAY HOLLA 

So here are two more changes to Flowgen:
(1) SYN and FIN packets are now payload-less, so as not to trigger much angst in our NFs.

(2) (And this will need some explaining) -> it now chooses a new <IP Src, IP Dst, Src Port, Dst Port> for each flow. The user can specify a "range" for these values that are an increment off the base packet_template. So for example, if the template says SRC IP = 10.0.0.1, and my Src Port Range is 4, It will generate flows with SRC IP \in {10.0.0.1, 10.0.0.2, 10.0.0.3, 10.0.0.4}.

Here's why this needs to be in Flowgen: with the other packetgen, we just do random updates to e.g, update the source IP or change the ports. But that doesn't retain the flow semantics -- doing it randomly breaks it -- everything from a flow has the same header, sequence numbers line up, etc. The rewriting needs to be consistent. 

Doing this with a "range" value is what the WindRiver packetgen does (one of the sample DPDK apps), so I literally just copied their idea.

### DIFF
--- a/core/modules/flowgen.cc
+++ b/core/modules/flowgen.cc
@@ -1,11 +1,14 @@
 #include "flowgen.h"
 
+#include <arpa/inet.h>
 #include <cmath>
 #include <functional>
 
-#include <arpa/inet.h>
 #include "../mem_alloc.h"
+#include "../utils/ether.h"
 #include "../utils/format.h"
+#include "../utils/ip.h"
+#include "../utils/tcp.h"
 #include "../utils/time.h"
 
 #define MAX_TEMPLATE_SIZE 1536
@@ -17,7 +20,7 @@ typedef std::priority_queue<Event, std::vector<Event>,
                             std::function<bool(Event, Event)>>
     EventQueue;
 
-//Priority queue must be a *min* heap -> next upcoming event first.
+// Priority queue must be a *min* heap -> next upcoming event first.
 bool EventLess(const Event &a, const Event &b) {
   return a.first > b.first;
 }
@@ -87,7 +90,10 @@ inline struct flow *FlowGen::ScheduleFlow(uint64_t time_ns) {
 
   f->first = 1;
   f->next_seq_no = 12345;
-  f->flow_id = rng_.Get();
+  f->src_ip = htonl(ip_src_base_ + rng_.GetRange(ip_src_range_));
+  f->dst_ip = htonl(ip_dst_base_ + rng_.GetRange(ip_dst_range_));
+  f->src_port = htons(port_src_base_ + rng_.GetRange(port_src_range_));
+  f->dst_port = htons(port_src_base_ + rng_.GetRange(port_dst_range_));
 
   /* compensate the fraction part by adding [0.0, 1.0) */
   f->packets_left = NewFlowPkts() + rng_.GetReal();
@@ -166,6 +172,10 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
 
   memset(templ_, 0, MAX_TEMPLATE_SIZE);
   memcpy(templ_, arg.template_().c_str(), template_size_);
+  pb_error_t err = UpdateBaseAddresses();
+  if(err.err() != 0){
+    return err;
+  }
 
   total_pps_ = arg.pps();
   if (std::isnan(total_pps_) || total_pps_ < 0.0) {
@@ -204,6 +214,23 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
 
   if (arg.quick_rampup()) {
     quick_rampup_ = 1;
+  }
+
+  ip_src_range_ = arg.ip_src_range();
+  ip_dst_range_ = arg.ip_dst_range();
+
+  if (arg.port_src_range() > 65535 || arg.port_dst_range() > 65535) {
+    return pb_error(EINVAL, "portrang must be e <= 65535");
+  }
+
+  port_src_range_ = (uint16_t)arg.port_src_range();
+  port_dst_range_ = (uint16_t)arg.port_dst_range();
+
+  if (ip_src_range_ == 0 && ip_dst_range_ == 0 && port_src_range_ == 0 &&
+      port_dst_range_ == 0) {
+    /*randomize ports anyway*/
+    port_dst_range_ = 20000;
+    port_src_range_ = 20000;
   }
 
   return pb_errno(0);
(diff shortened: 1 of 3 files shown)


justinemarie:
Also also, as soon as this and the other Flowgen PR are merged, it's two seconds to get Barath that code I promised him for the tests he needs to run (so please look over this soon :-)

justinemarie:
Oh, and I'll also add -- I'm not too concerned about the performance impact of the updated ports/ips because creating packets is *already* expensive and this is small change on top of that (and for big packetgen tests we run the flowgen multicore anyway).

core/modules/flowgen.cc
```diff
@@ -206,6 +206,23 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
     quick_rampup_ = 1;
   }
 
+  ip_src_range_ = arg.ip_src_range();
+  ip_dst_range_ = arg.ip_dst_range();
+  port_src_range_ = (uint16_t) arg.port_src_range();
+  port_dst_range_ = (uint16_t) arg.port_dst_range();
+
+  if(ip_src_range_ == 0 &&
```
barath: Just a general comment -- run clang-format.

core/modules/flowgen.cc
```diff
@@ -327,13 +344,38 @@ bess::Packet *FlowGen::FillPacket(struct flow *f) {
   if (f->packets_left <= 1)
     tcp_flags |= 0x01; /* FIN */
 
-  *(reinterpret_cast<uint32_t *>(p + 14 + /* IP dst */ 16)) = f->flow_id;
+  if(ip_src_range_ != 0){
```
barath: You might instead use the headers in utils/ip.h and utils/tcp.h and then you can access these various fields as a struct rather than manually accessing them. (Unless I'm not understanding what you're trying to do.)

core/modules/flowgen.cc
```diff
@@ -206,6 +209,18 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
     quick_rampup_ = 1;
   }
 
+  ip_src_range_ = arg.ip_src_range();
+  ip_dst_range_ = arg.ip_dst_range();
+  port_src_range_ = (uint16_t)arg.port_src_range();
```
sangjinhan: Port types should be uint16 in .proto file in the first place, rather than doing type casting here.

protobuf/module_msg.proto
```diff
@@ -185,6 +185,10 @@ message FlowGenArg {
   string arrival = 5;
   string duration = 6;
   bool quick_rampup = 7;
+  uint32 ip_src_range = 8;
+  uint32 ip_dst_range = 9;
+  uint32 port_src_range = 10;
```
sangjinhan: uint16

core/modules/flowgen.cc
```diff
@@ -327,16 +342,48 @@ bess::Packet *FlowGen::FillPacket(struct flow *f) {
...
+  }
+
+  if (port_src_range_ != 0) {
+    uint16_t *tcp_src = &(tcpheader->src_port);
+    *tcp_src = htons(ntohs(*tcp_src) + (f->flow_id % port_src_range_));
+  }
+
+  if (port_dst_range_ != 0) {
+    uint16_t *tcp_dst = &(tcpheader->dst_port);
+    *tcp_dst = htons(ntohs(*tcp_dst) + (f->flow_id % port_dst_range_));
```
sangjinhan: One concern here is the generated 4-tuple won't have enough entropy, as all field values increase in lock step. Suppose that both port_src_range and port_dst_range are 100. One might expect there will be 10000 flows generated, but the current code would generates only 100 flows (with the same offset for src/dst ports).

This behavior is somewhat counterintuitive, since if port_src_range == 99 and port_dst_range == 100, the code would do generate 9900 flows.

core/modules/flowgen.cc
```diff
@@ -327,16 +342,48 @@ bess::Packet *FlowGen::FillPacket(struct flow *f) {
...
+  }
+
+  tcpheader->flags = tcp_flags;
+  tcpheader->seq_num = htonl(f->next_seq_no);
 
   f->next_seq_no += f->first ? 1 : size - (14 + 20 + 20); /* eth + ip + tcp*/
 
+  if (f->first || f->packets_left <= 1) {  // syn or fin
+    pkt->set_total_len(60);                /*eth + ip + tcp*/
```
sangjinhan: Since `set_total_len()`/`set_data_len()` is already used above, all packets with SYN/FIN flags will invoke these functions twice (waste of cycles). We can defer the code on line 335-336 until this point.

core/modules/flowgen.cc
justinemarie: Protobuf doesn't support shorts -- only 32-bit and 64-bit ints.

core/modules/flowgen.cc
justinemarie: I had considered this but I decided it wasn't too confusing -- but I guess maybe it is counterintuitive and that's bad. Do you know of a fast/easy way to make it randomize the space a bit better? Hash the counter and then pull the offset values from different indices in the hash output?

core/modules/flowgen.cc
```diff
@@ -3,10 +3,13 @@
...
 #include <functional>
 
-#include <arpa/inet.h>
 #include "../mem_alloc.h"
+#include "../utils/ether.h"
 #include "../utils/format.h"
+#include "../utils/ip.h"
+#include "../utils/tcp.h"
 #include "../utils/time.h"
+#include <arpa/inet.h>
```
justinemarie: Fixed

core/modules/flowgen.cc
justinemarie: fixed

protobuf/module_msg.proto
justinemarie: protobuf doesn't support uint16

core/modules/flowgen.cc
sangjinhan: Oh that's unfortunate... Then perhaps we can do range check? One may innocently specify 65536 to intend "all ports", but it will result in `port_src_range == 0` due to narrow casting.

justinemarie: Ok, I will add a range check

core/modules/flowgen.cc
sangjinhan: My suggestion is:
* Remove `flow_id` from the struct. It's only used for IP/port manipulation. Instead we add a 4-tuple in `flow`.
* The 4-tuple is set in `ScheduleFlow()`, thus only once for each flow, not every packet. Since integer modulo operation `%` is hella expensive we shouldn't recompute IPs/ports for every FillPacket() operation.
* In `ScheduleFlow()`, you can use a new random number for each field to update, rather than using only one random number for all fields. Note there is `Random::GetRange()`, which is an optimized version of `rand() % range` you can use.

justinemarie: Ok!

justinemarie: Wait -- what about collisions? With the rotating counter you can't have two flows with the same 5-tuple at the same time (so long as you have more possible flow IDs than possible concurrent active flows). If we randomize, we wind up with birthday paradox...

justinemarie: (This matters because flow reconstruction)

sangjinhan: `flow_id` used to be a randomly generated 32-bit integer so collision is not a new issue here. You may want to generate (src IP, dst IP, src port, dst port) permutations in lexicographic order to minimize the chance of collision.

justinemarie:
Assuming this passes Travis, can someone take a look and make more comments / go ahead and merge? That push updates the last of Sangjin's comments.

justinemarie:
(Please do wait for Travis to finish before pushing).

sangjinhan:
LGTM (other than there should be a separator line between the system header <arpa/inet.h> and the following standard C++ headers... 😄)

sangjinhan:
Does GetRange() work fine with 0 range...?

justinemarie:
Yes, I ran it with the test script which defaults it to zero.

melvinw:
Thanks! 
v7tto2w542cy

PR #273, opened 2017-01-27
Splits the constructor and destructor for TCP flow reconstruction.

barath:
Splits the constructor and destructor for TCP flow reconstruction to get codecov to report on it.


barath:
Codecov appears to be having difficulty picking up coverage changes, and this didn't do the trick.  Will look into this...

PR #277, opened 2017-02-03
Fix memory hogging bug in flowgen

sangjinhan:
- When the event queue (priority_queue) is empty, top() returns a garbage value. It results in an infinite loop, pushing a (garbage) flow to the pool for each iteration.
- Another issue addressed: the pool of free flow structs is sized 120% of the number of average concurrent connections. Since now Update() command can change the load level at runtime, the pool size should not be hard-limited. Now the pool grows dynamically as needed.

### DIFF
--- a/core/modules/flowgen.cc
+++ b/core/modules/flowgen.cc
@@ -84,12 +84,16 @@ inline uint64_t FlowGen::NextFlowArrival() {
 }
 
 inline struct flow *FlowGen::ScheduleFlow(uint64_t time_ns) {
-  if (flows_free_.empty()) {
-    return nullptr;
-  }
+  struct flow *f;
 
-  struct flow *f = flows_free_.front();
-  flows_free_.pop_front();
+  if (flows_free_.empty()) {
+    f = new flow();
+    if (!f)
+      return nullptr;
+  } else {
+    f = flows_free_.top();
+    flows_free_.pop();
+  }
 
   f->first = 1;
   f->next_seq_no = 12345;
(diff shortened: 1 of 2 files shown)


core/modules/flowgen.cc
```diff
@@ -84,13 +84,17 @@ inline uint64_t FlowGen::NextFlowArrival() {
 }
 
 inline struct flow *FlowGen::ScheduleFlow(uint64_t time_ns) {
+  struct flow *f;
+
   if (flows_free_.empty()) {
-    return nullptr;
+    f = new flow();
+    if (!f)
+      return nullptr;
```
barath: Isn't this redundant?  That is, might as well just write:
return new flow()

core/modules/flowgen.h
```diff
@@ -89,7 +89,7 @@ class FlowGen final : public Module {
   int allocated_flows_;
   uint64_t generated_flows_;
   struct flow *flows_;
-  std::deque<struct flow *> flows_free_;
+  std::stack<struct flow *> flows_free_;
```
barath: Could you add a member variable comment for this (and if you feel like it, also for the other members)?

core/modules/flowgen.cc
sangjinhan: If allocation succeeded, the control flow should fall through.

Actually `new` won't return a nullptr but an exception upon a failure, so the if statement is not really necessary.

barath:
img_0730

PR #278, opened 2017-02-04
Replace HTable with CuckooMap

changlan:
- Replace HTable with CuckooMap
- Remove HTable
- Add iterators, `begin`, `end`.
- Remove `*_bench`'s dependency on `*.o` in Makefile


changlan:
Make cannot build target 'utils/cuckoo_map_bench'. Odd..

sangjinhan:
The current Makefile requires you have `utils/cuckoo_map.cc`, I suppose?

barath:
Could you add a usage comment to the .h file (before the class declaration), briefly giving an example of how to use it (any maybe noting any differences from similar STL containers in its behavior / API / guarantees)?

sangjinhan:
How does it compare with HTable, when it comes to `samples/exactmatch` and `samples/wildcardmatch` examples?

changlan:
Overheads shown by `perf top`:
 
`ExactMatch::ProcessBatch`: 42% -> 44%

`WildcardMatch::LookupEntry`: 53% -> 49%
`WildcardMatch::ProcessBatch`: 9% -> 14%

sangjinhan:
What perf shows is relative percent overhead, which is not directly proportional to per-packet cost. With samples/wildcardmatch, the throughput is 13.16 Mpps (new) vs 16.75 Mpps old). Let's think about what we can do narrow the gap...

core/modules/exact_match.cc
```diff
@@ -84,33 +84,23 @@ pb_error_t ExactMatch::Init(const bess::pb::ExactMatchArg &arg) {
...
 
-  default_gate = ACCESS_ONCE(default_gate_);
-
   for (int i = 0; i < cnt; i++) {
-    memset(&keys[i][key_size - 8], 0, sizeof(uint64_t));
+    char *p = reinterpret_cast<char *>(&keys[i]);
+    memset(p + total_key_size_ - 8, 0, sizeof(uint64_t));
   }
 
+  default_gate = ACCESS_ONCE(default_gate_);
```
justinemarie: Don't mind me... what does ACCESS ONCE mean?

sangjinhan: default_gate_ might be updated during ProcessBatch() by a gRPC thread (with .mt_safe=1 command). ACCESS_ONCE ensures that the variable is read only once at the beginning of ProcessBatch(), so that the value does not change in the middle (since it is stored in the local variable).

ACCESS_ONCE is a macro that temporarily makes a variable `volatile`; compiler won't try to optimize its access. Otherwise, default_gate_ can be read from memory multiple times, potentially with different values.

sangjinhan:
@changlan I fixed the following correctness issues (and did a bit of microoptimization to compensate the overhead introduced by the patches). Please review (and merge if no other issues are found).

* When the bucket table expands, already hashed items must be relocated
* Find() can fail if multiple keys share the same hash value.
* WildcardMatch may use keys with garbage padding bytes, if total_key_size_ is not a multiple of 8

changlan:
Wonderful. Thanks Sangjin!
image

PR #279, opened 2017-02-04
Do not run google benchmarks on CI

changlan:
... in order to reduce the feedback time. Running benchmarks on CI is not useful as the performance numbers are not very precise.


barath:
Just as a general (future) thing to be aware of: we should ensure that any code paths that are executed in benchmarks are also covered in unit tests, so that the elimination of benchmarks in Travis won't leave us unaware of any bugs.

sangjinhan:
20170121000845_6pthvylu_1a1b2f7f371f64c2ad2970668f39e072_2h1hsxfkbc64kbvq5lli5l33h12

PR #280, opened 2017-02-06
Fix Makefile issues

sangjinhan:
- Use canonical filename for protobuf targets
  - protoc does not handle bar/../foo.proto correctly. This often causes build failures with parallel build (`make -j`)
- Allow *_bench.o for header-only code (#278)
- Clean up Makefile style issues
  - Indentation should be hard tabs, which are 8-chars wide
  - Remove trailing whitespace
  - Prefer simply expanded variables := to =
  - Use non-nested conditional statements
  - Use $$ only when it is necessary
- Rename CORE_OBJS -> LIB_OBJS


barath:
img_1379

PR #281, opened 2017-02-07
Fix printing error in testing script.

justinemarie:
Josh found a bug in the testing script only triggered when there are errors :-)


barath:
img_1875

PR #282, opened 2017-02-07
allow argument updates for certain traffic classes

melvinw:
This may be useful for certain applications. Right now it only supports the rate limit TC.

### DIFF
--- a/core/traffic_class.h
+++ b/core/traffic_class.h
@@ -283,7 +283,7 @@ class WeightedFairTrafficClass final : public TrafficClass {
   void FinishAndAccountTowardsRoot(Scheduler *sched, TrafficClass *child,
                                    resource_arr_t usage, uint64_t tsc) override;
 
-  inline resource_t resource() const { return resource_; }
+  resource_t resource() const { return resource_; }
 
   const extended_priority_queue<ChildData> &children() const {
     return children_;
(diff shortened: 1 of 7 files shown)


core/traffic_class.h
```diff
@@ -376,16 +376,36 @@ class RateLimitTrafficClass final : public TrafficClass {
...
+  // Convert resource units to work units
+  static uint64_t to_work_units(uint64_t x) {
+    return (x << (USAGE_AMPLIFIER_POW - 4)) / (tsc_hz >> 4);
+  }
+
+  // Convert work units to resource units
+  static uint64_t to_resource_units(uint64_t x) {
+    return (x * (tsc_hz >> 4)) >> (USAGE_AMPLIFIER_POW - 4);
+  }
+
```
barath: Could you add a really simple unit test that checks that x == to_resource_units(to_work_units(x))?

melvinw: Yup

sangjinhan: Because of the nature of the operation, `x == to_resource_units(to_work_units(x))` cannot be guaranteed. Even with Melvin's +1 hack does not always work correctly (tried millions of times):

```
traffic_class_test.cc:593: Failure
Value of: RateLimitTrafficClass::to_resource_units( RateLimitTrafficClass::to_work_units(val))
  Actual: 2692831867
Expected: val
Which is: 2692831866
```

`to_resource_units()` is only used by bessctl to retrieve a resource limit value in bits/pkts/cycles per sec, which was originally given by the user. We can simply remove the function, and store the original value (`limit_org_`) given by the user, besides the converted one (`limit_`).

The comment below seems incorrect. Should be `In work units / CPU cycle` ...?
```
uint64_t limit_;      // In bits/pkts/cycles per sec. (0 if unlimited).
```

barath: Yeah, Melvin and I discussed it.  It's truncation, and the +1 hack is not a good option.  We agreed the best option is to store the original value.

melvinw:
All fixed up

core/traffic_class.h
```diff
@@ -399,9 +430,11 @@ class RateLimitTrafficClass final : public TrafficClass {
   // prof->limit < 2^36 (~64 Tbps)
   // 2^24 < tsc_hz < 2^34 (16 Mhz - 16 GHz)
   // tb->limit < 2^36
-  uint64_t limit_;      // In bits/pkts/cycles per sec. (0 if unlimited).
-  uint64_t max_burst_;  // In work units.
-  uint64_t tokens_;     // In work units.
+  uint64_t limit_;          // In work units per cycle (0 if unlimited).
+  uint64_t limit_arg_;      // In work units per cycle (0 if unlimited).
+  uint64_t max_burst_;      // In resource units per second.
```
barath: I think you meant to swap these -- the ones with arg_ are in resource units.

melvinw: Doh. Will fix

sangjinhan:
gzxdcfe6hj5mpfn502qp0hnso1_500

PR #284, opened 2017-02-08
Multi-Level Feedback Queue Module

TheJStone:
/*
   The Multi-Level Feedback Queue module queues from one input gate to one output gate without any packet modification. Queue is set up as having specified number of priority levels where each Flow will be assigned based on its priority. The priority for each flow is calculated as a function of Flow throughput and average load of the Module. After receiving a batch to process, in which each packet is assigned to a flow based on their 5 tuple, The module will find the highest priority level and split the current return batch size between the all of this levels flows evenly. The module will then query those flows for packets based on this evenly split batch size. The module will then return this new packet batch.

    based on this idea:
      https://en.wikipedia.org/wiki/Multilevel_feedback_queue
  * EXPECTS: Input packets in any format
  
  * MODIFICATIONS: None
  
  * INPUT GATES: 1
  
  * OUTPUT GATES: 1
  
  * PARAMETERS:
      * Batch size: the size of the output batches
      * Number of levels: Number of levels in the Multi-Level Flow Queue
      * Initial Load: the predicted average load of the module as a initial base to calculate priorities
      * Max Flow Queue Size: the maximum size that any Flows queue can get before the module will start dropping the flows packets
  * COMMANDS
      * update Batch Size: can be done live
      * update Number of levels: cannot not be done live
      * update Max Flow Queue Size: can be fone live


core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  return ips && ports && id1.protocol == id2.protocol;
+}
+bool operator<(const MLFQueue::Q_Id &id1, const MLFQueue::Q_Id &id2)  {
+  bool ips = id1.src_ip < id2.src_ip && id1.dst_ip < id2.dst_ip;
+  bool ports = id1.src_port < id2.src_port && id1.dst_port < id2.dst_port;
+  return ips && ports && id1.protocol < id2.protocol;
+}
+
+int RoundToPowerTwo(int val) {
+ return pow(2, ceil(log2(val)));
```
barath: This will probably be slow, so you don't want to be executing it on the fast path.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+    {"set_num_lvls", "MlfqLvlArg", MODULE_CMD_FUNC(&MLFQueue::CommandNumPriorityLevels), 0},
+    {"set_batch_Size", "MlfqBatchArg", MODULE_CMD_FUNC(&MLFQueue::CommandBatchSize), 0},
+    {"set_max_flow_queue_size", "MlfqMaxFlowQueueSizeArg",
+          MODULE_CMD_FUNC(&MLFQueue::CommandMaxFlowQueueSize), 0}};
+
+pb_error_t MLFQueue::Init(const bess::pb::MlfqArg &arg) {
+  pb_error_t err;
+
+  if (arg.num_lvls() != 0) {
+    err = SetNumPriorityLevels(arg.num_lvls());
```
barath: Better to avoid duplicated code like this.  Instead, you could have (suitably modified for each instance, and for the pending change of constant name):

auto levels = arg.num_lvls() ? arg_num_lvls() : DEFAULT_NUM_LVLS;

Then you don't need the if statement or duplicated code.  Same for the other args.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+    load_avg_ = INITIAL_LOAD;
+  }
+  ready_flows_ = 0;
+  return pb_errno(0);
+}
+
+void MLFQueue::DeInit() {
+  bess::Packet *pkt;
+
+  for (std::map<Q_Id, Flow>::iterator it=flows_.begin(); it!=flows_.end(); ) {
```
barath: Unless I'm missing it, I don't see where you advance the iterator here.

Also, prefer auto for iterator types to reduce clutter.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+}
+
+pb_cmd_response_t MLFQueue::CommandMaxFlowQueueSize(const bess::pb::MlfqMaxFlowQueueSizeArg &arg) {
+  pb_cmd_response_t response;
+  set_cmd_response_error(&response, SetMaxFlowQueueSize(arg.max_queue_size()));
+  return response;
+}
+
+void MLFQueue::ProcessBatch(bess::PacketBatch *batch) {
+  google::SetLogDestination(google::GLOG_INFO,"/vagrant/batch.log");
```
barath: This is fine for testing, but needs to be removed before merging the PR.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  LOG(INFO)<< std::to_string(batch->cnt());
+  //insert packets in the batch into their corresponding flows
+  for (int i = 0; i < batch->cnt(); i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    Q_Id id = MLFQueue::GetId(pkt);//doesn't handle fragmented packets
```
barath: Instead, you could write this as a TODO, e.g.:

// TODO(joshua): Add support for fragmented packets.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+    std::map<Q_Id, Flow>::iterator it = flows_.find(id);
```
barath: Prefer auto for iterators.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+pb_cmd_response_t MLFQueue::CommandMaxFlowQueueSize(const bess::pb::MlfqMaxFlowQueueSizeArg &arg) {
+  pb_cmd_response_t response;
+  set_cmd_response_error(&response, SetMaxFlowQueueSize(arg.max_queue_size()));
+  return response;
+}
+
+void MLFQueue::ProcessBatch(bess::PacketBatch *batch) {
+  google::SetLogDestination(google::GLOG_INFO,"/vagrant/batch.log");
+  LOG(INFO)<< std::to_string(batch->cnt());
+  //insert packets in the batch into their corresponding flows
```
barath: Follow style guide for comments.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+    Q_Id id = MLFQueue::GetId(pkt);//doesn't handle fragmented packets
+    std::map<Q_Id, Flow>::iterator it = flows_.find(id);
+
+    //if the Flow doesn't exist create one
+    //and add the packet to the new Flow
+    if (it == flows_.end()) {
+      Flow f = {(float) max_lvl_, 0, 0, 0};
+      int slots = RoundToPowerTwo(FLOW_QUEUE_FACTOR*batch_size_);
+      int err = MLFQueue::AddQueue(&f, slots);
+      assert(err == 0);
```
barath: Is this assertion a substitute for an error check rather than checking an invariant?  If so, you should add a TODO saying that you need to do proper error handling here.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  for (int i = 0; i < batch->cnt(); i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    Q_Id id = MLFQueue::GetId(pkt);//doesn't handle fragmented packets
+    std::map<Q_Id, Flow>::iterator it = flows_.find(id);
+
+    //if the Flow doesn't exist create one
+    //and add the packet to the new Flow
+    if (it == flows_.end()) {
+      Flow f = {(float) max_lvl_, 0, 0, 0};
+      int slots = RoundToPowerTwo(FLOW_QUEUE_FACTOR*batch_size_);
```
barath: Probably too slow to be called on the fast path.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+      int err = MLFQueue::AddQueue(&f, slots);
+      assert(err == 0);
+      assert(f->queue);
+      err += 1;
```
barath: Use ++ if it's just 1.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+}
+
+//obtain the next batch of packets from the highest priority level's Flows.
+bess::PacketBatch MLFQueue::GetNextBatch() {
+  bess::PacketBatch batch;
+  batch.clear();
+
+  //the highest priority level with flows.
+  int lvl = FindTopLevel();
+  if(levels_[lvl].size() == 0) return batch;
```
barath: Put on a separate line and use braces.  (This is a BESS-specific decision we made above the style guide.)

Prefer .empty()

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  RunNextModule(&next_batch);
+}
+
+//obtain the next batch of packets from the highest priority level's Flows.
+bess::PacketBatch MLFQueue::GetNextBatch() {
+  bess::PacketBatch batch;
+  batch.clear();
+
+  //the highest priority level with flows.
+  int lvl = FindTopLevel();
```
barath: Prefer non-abbreviated variable names.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  //and update their priority location
+  UpdateAllFlows(false, lvl);
+  return batch;
+}
+
+//dequeues from the specified slots from the flow f into the Packetbatch
+int MLFQueue::AddtoBatch(bess::PacketBatch *batch, Flow* f, int slots) {
+  int prev = batch->cnt();
+  int cnt = llring_dequeue_burst(f->queue, (void **)batch->pkts(), slots);
+  if(cnt > 0) batch->set_cnt(prev+cnt);
```
barath: On its own line with brace.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  return cnt;
+}
+
+//Find the first priority level with Flows in ascending order
+int MLFQueue::FindTopLevel() {
+  for(int i = max_lvl_; i >= 0; i--) {
+    if(!levels_[i].empty()) return i;
```
barath: On its own line with brace.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  }
+  return 0;
+}
+
+// gets the 5 element identifier for a flow given a packet
+MLFQueue::Q_Id MLFQueue::GetId(bess::Packet* pkt) {
+      struct ether_hdr *eth = pkt->head_data<struct ether_hdr *>();
```
barath: Fix indent.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+
+      mem_free(old_queue);
+    }
+  }
+  ret = llring_sp_enqueue(f->queue, newpkt);
+  time(&f->timer);
+  return ret;
+}
+
+/*updates the metrics that determine the priority level of the flow
```
barath: Move these sorts of method comments to the .h file.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+//top_lvl is the location of the highest priority flows and update_pkts
+//indicates whether these flows still need their throughput updated in UpdateFlow
+void MLFQueue::UpdateAllFlows(bool update_pkts=true, int top_lvl=0) {
+  levels_.clear();
+  levels_.resize(max_lvl_+1);
+  time_t now;
+  time(&now);
+  // iterate through all flows and recalculate their priority
+  //and assign them to the corresponding priority level
+  for (std::map<Q_Id, Flow>::iterator it=flows_.begin(); it!=flows_.end(); ) {
```
barath: auto for iterator.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+      else UpdateFlow(f);
+      InsertFlow(f);
+      ++it;
+  }
+  //having a newly calibrated ready flows, updates the load average
+  load_avg_ = (max_lvl_ -1)/((float) max_lvl_) + (1.0/max_lvl_)*ready_flows_;
+}
+
+void MLFQueue::InsertFlow(Flow* f) {
+  int lvl = (int) f->priority;
```
barath: Given that use use the priorities as ints, why not just have them be ints?

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+  levels_[lvl].push_back(f);
+}
+
+void MLFQueue::Resize(uint8_t num_lvls) {
+  if(num_lvls - 1 == max_lvl_) { return; }
```
barath: On its own line.

core/modules/mlfqueue.cc
```diff
@@ -0,0 +1,346 @@
...
+                    bess::PacketBatch::kMaxBurst);
+  }
+
+  batch_size_ = size;
+  return pb_errno(0);
+}
+
+pb_error_t MLFQueue::SetMaxFlowQueueSize(uint32_t queue_size) {
+  if(queue_size == 0) {
+    return pb_error(EINVAL, "max queue size must be atleast 1");
```
barath: at least

justinemarie:
Note that Travis failed. Josh, if you click through the continuous integration link it will show you the compilation errors --- in particular we compile with some strict checks so it could be that things passed in your environment but not with some other configuration. (I once had Travis find a bug that only turned up with g++6!)

bessctl/conf/testing/module_tests/mlfqueue.py
```diff
@@ -0,0 +1,36 @@
+import sugar
+from module import *
+from port import *
+
+def basic_throughput_test():
```
justinemarie: It's really an output test -- that the thing outputs the packets you expect it to. A throughput test would be benchmarking packets per second. Rename?

bessctl/conf/testing/module_tests/mlfqueue.py
```diff
@@ -0,0 +1,36 @@
...
+    batch_basic = MLFQueue(num_lvls=5, batch_size=5, init_load=10.0)
+    packet_list = []
+    packet_list = gen_packet_list(scapy.TCP, '22.22.22.22', '22.22.22.22', 2)
+    packet_list += gen_packet_list(scapy.TCP, '31.14.15.69', '31.14.15.69', 3)
+    out.append([batch_basic,  # test this module
+                               1, 1,  # it has one input port and one output port
+                               packet_list])
+    return out
+
+def crash_test():
```
justinemarie: I like that you scoped each test within a function -- it makes the script read cleaner than mine!

bessctl/conf/testing/module_tests/mlfqueue.py
```diff
@@ -0,0 +1,36 @@
+import sugar
+from module import *
+from port import *
+
+def basic_throughput_test():
+    def gen_packet_list(protocol, input_ip, output_ip, num_pkts):
```
justinemarie: Can you put a comment here that just says, this generates the input for an output test where you expect the same packet to come out as goes in?

bessctl/conf/testing/module_tests/mlfqueue.py
```diff
@@ -0,0 +1,36 @@
...
+    packet_list = gen_packet_list(scapy.TCP, '22.22.22.22', '22.22.22.22', 2)
+    packet_list += gen_packet_list(scapy.TCP, '31.14.15.69', '31.14.15.69', 3)
+    out.append([batch_basic,  # test this module
+                               1, 1,  # it has one input port and one output port
+                               packet_list])
+    return out
+
+def crash_test():
+    return [MLFQueue(num_lvls=5, batch_size=32, init_load=10.0), 1, 1]
+
```
justinemarie: Would be good to have a test that actually looks at the core functionality of your Queue -- these tests just make sure the packets aren't modified and that the thing doesn't crash. But what about a test that shows that the Queue is balancing load as expected? (You can do this with a custom test.)

core/modules/mlfqueue.cc
TheJStone: I have errors that I would through if the passed-in arguments aren't within a scope and also use those functions for the command messages to test for scope as well. Would you rather that I just take what they give regardless?

PR #285, opened 2017-02-08
add module name to sockname to avoid socket-reuse between tests

justinemarie:
When we create the UnixSockets for testing we need to create a new one for each test. Tony's code revealed a bug where socket names were being reused between tests for different modules, so I added the module name to make sure the socket doesn't get reused. 

Would be great if this got merged quickly before our 1pm meeting with the undergrads.


sangjinhan:
Module names (in fact, all names) should be `[_a-zA-Z][_a-zA-Z0-9]*`. I guess the code verifying this condition was gone during the c++ porting. Could you replace `-` with `_`?

sangjinhan:
tumblr_m1jqugpiwu1qzex9io1_1280

PR #286, opened 2017-02-08
Added -y flag to prevent playbook execution from hanging on confirmation

c2tonyc2:
Fixed the minor error in the ansible playbook hanging on confirmation for the installation of linux headers


justinemarie:
Thanks for the fix, Tony!
image

PR #287, opened 2017-02-09
Stop crash if you free an empty packet batch

justinemarie:
So, there is an argument to be had whether or not we should have this -- I'm adding a check literally every time we free a packetbatch here. One could argue that it's the module's job to check that it's not handing in an empty packet batch -- in which case, I'll replace this pull request with one that just adds a comment saying "DO NOT CALL WITH CNT == 0".

That said, chasing down why things were crashing here was pretty obnoxious and I think it would be good to save future folks from shooting themselves in the foot. So, please discuss and let me know whether we prefer the comment or check route. This PR adds a check.

### DIFF
--- a/core/packet.h
+++ b/core/packet.h
@@ -385,6 +385,7 @@ int Packet::Alloc(Packet **pkts, size_t cnt, uint16_t len) {
 }
 
 void Packet::Free(Packet **pkts, int cnt) {
+  if(cnt == 0) return;
   struct rte_mempool *pool = pkts[0]->pool_;
 
   int i;


sangjinhan:
I think adding the check is reasonable.

Please reformat the code..?

There are places where the check is done individually, e.g. modules/buffer.cc, 
```
void Buffer::DeInit() {
  bess::PacketBatch *buf = &buf_;

  if (buf->cnt()) {   # <- now unnecessary
    bess::Packet::Free(buf);
  }
}
```

Could you find those spots and update them?

sangjinhan:
Another thing of the same theme is `RunNextModule()/RunChooseModule()` should not do anything if called with an empty batch. It would be great if someone can deal with this in a separate PR...

justinemarie:
@sangjinhan That sounds like a great plan. Let's keep this PR open and I'll make those changes, and then we can merge altogether.

sangjinhan:
Great, thanks!

core/packet.h
```diff
@@ -385,6 +385,7 @@ int Packet::Alloc(Packet **pkts, size_t cnt, uint16_t len) {
 }
 
 void Packet::Free(Packet **pkts, int cnt) {
+  if(cnt == 0) return;
```
barath: Just to be nitpicky:

if (cnt == 0) {
  return;
}

sangjinhan:
Let me take over this issue.

PR #289, opened 2017-02-09
bessctl: return mac addrs with port cmd responses

melvinw:
Scripts like `perftest/pktgen` might care about this sort of information.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -1032,7 +1032,7 @@ def _group(number):
 
 
 def _show_port(cli, port):
-    cli.fout.write('  %s/%s\n' % (port.name, port.driver))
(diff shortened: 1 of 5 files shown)


sangjinhan:
PortName/Driver is the canonical name of ports (ModuleName::Class is for modules). `show port` should show MAC addresses, but I think __str__ shouldn't include the MAC address by default...

Also, could you check if non-PMD drivers have the mac addr field appropriately (w/ generated one or zero-initialized one)?

melvinw:
Fixed `__str__`.  Yup, non-PMD drivers will get a randomized MAC if one isn't specified in CreatePort (https://github.com/NetSys/bess/blob/master/core/bessctl.cc#L245)

sangjinhan:
tumblr_m3osvuuigy1rn5cm9o1_1280

PR #291, opened 2017-02-09
Isolate background threads within a core

sangjinhan:
Background threads (spawned by either DPDK or gRPC) may interfere with worker threads, potentially introducing high jitter and transient packet loss. This commit provides a temporary fix to the issue, by isolating all non-worker threads in a separate core, which can be configurable with `taskset`.

### DIFF
--- a/core/dpdk.cc
+++ b/core/dpdk.cc
@@ -66,7 +66,7 @@ static ssize_t dpdk_log_writer(void *, const char *data, size_t len) {
 }
 
 static void init_eal(const char *prog_name, int mb_per_socket,
-                     int multi_instance, bool no_huge) {
+                     int multi_instance, bool no_huge, int default_core) {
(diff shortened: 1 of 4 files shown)


core/dpdk.cc
```diff
@@ -152,7 +152,38 @@ static void init_eal(const char *prog_name, int mb_per_socket,
   rte_openlog_stream(fopencookie(nullptr, "w", dpdk_log_funcs));
 }
 
+static int determine_default_core() {
```
barath: Could you add a comment saying that this returns the last core that is available, which is the "default" for use by non-workers?

sangjinhan: Yup, done.

sangjinhan:
🐈

justinemarie:
image

PR #292, opened 2017-02-10
Uses curl to fetch DPDK to avoid download flakiness due to urllib.

barath:


sangjinhan:
tumblr_m7l7zise4g1qdwizao1_1280

PR #295, opened 2017-02-11
make timestamp/measure offset configurable

melvinw:


sangjinhan:
Looks good. Some suggestions:
* How about having a 4-8 byte magic number to identify timestamped packets? Currently 0x01 is used, which is not reliable at all.
* If the packet length is not long enough for timestamp, the packet length needs to be adjusted (or just dropped). For example, the current code with 60B packets will have truncated timestamp, which silently disrupt the measurement results.
* Default offset could be based on UDP, not TCP?
* In the long-term: sequence number might be useful if we want to detect reordering or packet drop

melvinw:
Addressed all but the last bullet. Ended up using a 2 byte magic number instead of 4 or 8 because the latter resulted in a 26% performance hit.

sangjinhan:
tumblr_lwem4d8gsr1qzv52ko1_1280

PR #298, opened 2017-02-15
Random drop module

vivi:


core/modules/random_drop.h
```diff
@@ -0,0 +1,19 @@
+#ifndef BESS_MODULES_RANDOM_DROP_H_
+#define BESS_MODULES_RANDOM_DROP_H_
+
+#include "../module.h"
+#include "../module_msg.pb.h"
+#include "../utils/random.h"
+
+class RandomDrop final : public Module {
```
barath: Could you add a comment for the class?

core/modules/random_drop.h
```diff
@@ -0,0 +1,19 @@
...
+ public:
+  pb_error_t Init(const bess::pb::RandomDropArg &arg);
+  void ProcessBatch(bess::PacketBatch *batch) override;
+
+ private:
+  Random rng_;
```
barath: Could you add comments for each member variable?

core/modules/random_drop.h
```diff
@@ -0,0 +1,19 @@
...
+  uint threshold_;
+  const uint range_ = 1000000;  // for granularity
```
barath: Could you define this in the location / style for constants per the Google C++ style guide?

protobuf/module_msg.proto
```diff
@@ -386,3 +386,7 @@ message WildcardMatchArg {
   }
   repeated Field fields = 1;
 }
+
+message RandomDropArg {
```
barath: Could you add protobuf comment(s) here (see Justine's comments for guidance)?

sangjinhan:
Looks great. One high-level comment: perhaps this module can be more general (with almost the same code), so that it can be utilized in many different scenarios? Suppose this module is `RandomSplit`, then:
```
# emulating RandomDrop(drop_date=0.2)
a -> b::RandomSplit([{"prob": 0.8, "gate": 0}]) -> c
```
or
```
# weighted random distribution (50% for gate 0, 40% for gate 2, and 10% drop)
a -> b::RandomSplit([{"prob": 0.5, "gate": 0}, {"prob": 0.4, "gate": 2}])
b:0 -> c
b:2 -> d
```
As compared to your original module, this new syntax seems a bit uglier. One solution would be you implement the generic RandomSplit, and RandomDrop inherits it as a special case to maximize code reuse (within the same source file or in a separate .cc file)

justinemarie:
Hi Vivian,
The crash in Travis seems to be due to a bug that has already been patched in the Master branch. Can you do a git pull --rebase and re-commit so we can see that it should indeed pass Travis?

vivi:
Hey @sangjinhan, per what @barath and I discussed, we'll try check this in, and implement your suggestion in a future PR :)

core/modules/random_drop.h
```diff
@@ -0,0 +1,21 @@
+#ifndef BESS_MODULES_RANDOM_DROP_H_
+#define BESS_MODULES_RANDOM_DROP_H_
+
+#include "../module.h"
+#include "../module_msg.pb.h"
+#include "../utils/random.h"
+
+const uint kRange_ = 1000000;  // for granularity
```
barath: Sorry, just one more small change -- it would be best for this to be:

static const uint32_t kRange = 1000000;

And be defined in the public section of the RandomDrop class.

bessctl/conf/testing/run_module_tests.bess
```diff
@@ -145,7 +145,7 @@ for filename in glob.glob(path + "/conf/testing/module_tests/*.py"):
     for test in CUSTOM_TEST_FUNCTIONS:
         try:
             test()
-        except Error:
+        except:
```
justinemarie: Thanks for this fix. (I don't really know Python, lol)

bessctl/conf/testing/module_tests/random_drop.py
```diff
@@ -0,0 +1,47 @@
...
+rd1 = RandomDrop(drop_rate=0)
+test_packet = gen_packet(scapy.TCP, '22.22.22.22', '22.22.22.22')
+OUTPUT_TEST_INPUTS.append([rd1, 1, 1,
+                           [{'input_port': 0,
+                               'input_packet': test_packet,
+                                'output_port': 0,
+                                'output_packet': test_packet}]])
+
+## CUSTOM TESTS ##
+def half_drop_test():
```
justinemarie: One more tiny edit -- can you rewrite this test to test more than drop rate = 0.5 (like test at 0.75, 0.9, 0.3 as well?). You should be able to use function currying / partial evaluation to achieve this pretty quickly with your existing code (https://www.pydanny.com/python-partials-are-fun.html).

justinemarie:
Also @barath: how does codecov work? In this case, Vivian did write a test for her module, I thought codecov should see an improvement?

barath:
Codecov will only do coverage with explicit test routines as we've set it
up (all_test I think).  We could change it though.

barath:
img_1588

PR #299, opened 2017-02-15
Remove unused C-style error handling code

sangjinhan:
They used to be used to support Linux kernel style error handling, but not used any longer.


barath:
img_2049

PR #301, opened 2017-02-15
Fair queuing module

TheJStone:


core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+struct task_result DRR::RunTask(void *) {
+  bess::PacketBatch batch;
+  struct task_result ret;
+
+  batch.clear();
+  int total_bytes = 0;
+  if (flow_ring_ != NULL) {
+     total_bytes = GetNextBatch(&batch);
+  }
+  assert(total_bytes >= 0);  // TODO(joshua) do proper error checking
```
barath: Use DCHECK_LE(0, total_bytes) instead.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+  total_bytes++;
+
+  uint64_t cnt = batch.cnt();
+  if (cnt > 0) {
+    RunNextModule(&batch);
+  }
+
+  ret = (struct task_result){
+      .packets = cnt, .bits = (total_bytes + cnt * kPACKET_OVERHEAD) * 8,
```
barath: Move this calculation into its own line with a descriptive variable name.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+  int err;
+  double now = get_epoch_time();
+  Flow* f;
+  uint32_t total_bytes = 0;
+  int count = llring_count(flow_ring_);
+  int batch_size = batch->cnt();
+
+  // iterate through flows in round robin fashion until batch is full
+  while (!batch->full()) {
+    // checks to see if there has been no update after a full round
```
barath: Would you be able to revise this comment to explain why this is an interesting case (i.e., describe why rather than what or how)?

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+        batch_size = batch->cnt();
+      }
+    }
+    count--;
+
+    // gets the next flow in the round
+    if (!current_flow_) {
+      err =  llring_sc_dequeue(flow_ring_, reinterpret_cast<void**>(&f));
+      if (err < 0)
+        return err;
```
barath: Add braces.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+
+      if (llring_empty(f->queue) && !f->next_packet) {
+        // if the flow expired, remove it
+        if (now - f->timer > kTTL) {
+          RemoveFlow(f);
+        } else {
+          err = llring_sp_enqueue(flow_ring_, f);
+          if (err < 0)
+            return err;
```
barath: Add braces.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+      err =  llring_sc_dequeue(flow_ring_, reinterpret_cast<void**>(&f));
+      if (err < 0)
+        return err;
+
+      if (llring_empty(f->queue) && !f->next_packet) {
+        // if the flow expired, remove it
+        if (now - f->timer > kTTL) {
+          RemoveFlow(f);
+        } else {
+          err = llring_sp_enqueue(flow_ring_, f);
```
barath: I'm not sure I understand the purpose of this line -- could you explain it?

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+        break;
+      } else {
+        count = llring_count(flow_ring_);
+        batch_size = batch->cnt();
+      }
+    }
+    count--;
+
+    // gets the next flow in the round
+    if (!current_flow_) {
```
barath: Could you modularize this -- make this if statement's body into its own function.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+      int err = llring_sc_dequeue(f->queue, reinterpret_cast<void**>(&pkt));
+      if (err < 0)
+        return err;
+    } else {
+      pkt = f->next_packet;
+      f->next_packet = nullptr;
+    }
+
+    if (pkt->total_len() > f->deficit) {
+      f->next_packet = pkt;
```
barath: Do you do this because there is no peek() function?

core/modules/drr.h
```diff
@@ -0,0 +1,234 @@
...
+    #define kFLOW_QUEUE_SIZE 2048  // initial queue size for a flow
+    #define kQUEUE_GROWTH_FACTOR 2  // the scale at which a flow's queue grows
+    #define kFLOW_QUEUE_MAX 200000  // the max flow queue size if non-specified
+    #define kTTL 300  // time to live for flow entries
+    #define kPACKET_OVERHEAD 24  // additional bytes associated with packets
+    #define kDEFAULT_QUANTUM 1500  // default value to initialize qauntum_ to
+
+
+    // 5 tuple id to identify a flow from a packet header information.
+    struct Flow_Id {
```
barath: Rename to FlowId (no underscore).

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+  return id;
+}
+
+int DRR::AddNewFlow(bess::Packet *pkt, Flow_Id id) {
+  int err;
+
+  // creates flow
+  Flow* f = static_cast<Flow *>(mem_alloc_ex(sizeof(Flow), alignof(Flow), 0));
+  Flow init = {0, 0, id, 0, nullptr};
+  memcpy(f, &init, sizeof(init));
```
barath: Is there a reason you use mem_alloc here rather than new?  You could define a constructor on Flow then this would just be Flow *f = new Flow().

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+
+  // TODO(joshua) do proper error checking
+  err = AddQueue(&f->queue, static_cast<int>(kFLOW_QUEUE_SIZE));
+
+  if (err != 0)
+    return err;
```
barath: Braces.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+
+  flows_.Insert(id, f);
+
+  err = Enqueue(f, pkt);
+  if (err != 0)
+    return err;
```
barath: Braces.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+
+  // puts flow in round robin
+  err = llring_sp_enqueue(flow_ring_, f);
+  if (err != 0)
+    return err;
+
+  return 0;
+}
+
+void DRR::RemoveFlow(Flow* f) {
```
barath: Much of this could be placed in Flow's destructor, if you were to define one.  The only parts that would need to be outside would be parts that change flows_

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+  if (f->next_packet) {
+    bess::Packet::Free(f->next_packet);
+  }
+
+  mem_free(f->queue);
+  flows_.Remove(f->id);
+  mem_free(f);
+}
+
+int DRR::AddQueue(llring** queue, int slots) {
```
barath: It would be better to return an llring* and have the second argument to the method be an int *error that gets set upon error.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+    if (ret != 0) {
+      return ret;
+    }
+  }
+  ret = llring_sp_enqueue(f->queue, reinterpret_cast<void*>(newpkt));
+  f->timer = get_epoch_time();
+  return ret;
+}
+
+int DRR::ResizeQueue(llring **queue_loc, int new_size) {
```
barath: Here too it would be better to return an llring * and have an int *error second argument.  Then the semantics of the method would change slightly, in that the method would be destructive to its first argument (which would be an llring *old_queue).

core/modules/drr.h
```diff
@@ -0,0 +1,234 @@
...
+#include "../kmod/llring.h"
+#include "../utils/cuckoo_map.h"
+
+
+using bess::utils::IPAddress;
+using bess::utils::CIDRNetwork;
+using bess::utils::CuckooMap;
+
+/*
+  * This module implements Discrete Round Robin, a fair queueing algorithm, for flows.
```
barath: Deficit

core/modules/drr.h
```diff
@@ -0,0 +1,234 @@
...
+  *    * Max Number of flows: max number of flows the module will handle
+  *    * Max Flow Queue Size: the maximum size that any Flows queue can get
+  *          before the module will start dropping the flows packets
+  * COMMANDS
+  *    update quantum: cannot not be done live
+  *    update Max Flow Queue Size: can be done live
+*/
+class DRR final : public Module {
+ public:
+    #define kDEFAULT_NUM_FLOWS 4096  // the default max number of flows allowed
```
barath: This is a mix of macro and constant style.  If you want to define these here, then use static const int kDefaultNumFlows, etc. instead of macro style.  If you want macro style, define them outside of the class block as DEFAULT_NUM_FLOWS, etc.

core/modules/drr.h
```diff
@@ -0,0 +1,234 @@
...
+
+
+ private:
+    /*
+      sets the quantum: the number of bytes allocated to each flow on every round
+      Args:
+        size: the size to set the quantum to.
+      Returns:
+        0 on success and error value otherwise.
+    */
```
barath: You can squeeze the whitespace for these method comments as follows:

// Sets the quantum: the number of bytes allocated to each flow on every round.
// Takes the size to set the quantum to.  Returns 0 upon success or an error otherwise.

This makes it easy to read the method definitions in a single screen.

core/modules/drr.h
```diff
@@ -0,0 +1,234 @@
...
+  * PARAMETERS:
+  *    * quantum: the number of bytes allocated to each flow on each round
+  *    * Max Number of flows: max number of flows the module will handle
+  *    * Max Flow Queue Size: the maximum size that any Flows queue can get
+  *          before the module will start dropping the flows packets
+  * COMMANDS
+  *    update quantum: cannot not be done live
+  *    update Max Flow Queue Size: can be done live
+*/
+class DRR final : public Module {
```
barath: You should define a constructor and destructor.  The constructor should explicitly initialize all the members (preferably in the initializer list).

protobuf/module_msg.proto
```diff
@@ -87,6 +87,20 @@ message MeasureCommandGetSummaryResponse {
   uint64 latency_99_ns = 9;
 }
 
+message DRRArg {
+  uint32 num_flows = 1;//Number of flows to handle in module
```
barath: See the correct comment style in this pull request:

https://github.com/NetSys/bess/pull/304

core/modules/drr.cc
TheJStone: yes. its the solution we settled on a few meetings ago. same for current_flow_.

core/modules/drr.cc
TheJStone: Assuming its 187. there is a special case where you have to check if the non-empty queue becomes empty after dequeuing its packets so that you can reset its deficit counter to zero. However, if the queue has been empty, you don't reset its deficit counter. so, you check before trying to dequeue if it is an empty flow, and if so, you renqueue and continue in the while loop.

bessctl/conf/samples/drr.bess
```diff
@@ -0,0 +1,29 @@
...
+payload = 'helloworld'
+pkt_bytes1 = str(eth/ip1/udp/payload)
+pkt_bytes2 = str(eth/ip2/udp/payload)
+
+
+src1::Source() -> Rewrite(templates=[pkt_bytes1]) \
+  -> q::DRR(num_flows=4, quantum=1000) -> Sink()
+src2::Source() ->Rewrite(templates=[pkt_bytes2]) \
+  -> q
+
```
justinemarie: Oh okay goes to sink.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+    if (llring_empty(f->queue) && !f->next_packet) {
+      f->deficit = 0;
+    }
+
+    // if the flow doesn't have any more packets to give, reenqueue it
+    if (!f->next_packet || f->next_packet->total_len() > f->deficit) {
+      *err = llring_sp_enqueue(flow_ring_, f);
+      if (*err != 0) {
+        return total_bytes;
+	  }
```
melvinw: nit: Indentation is off here. I'd suggest running `clang-format`. It will format things to be consistent with the rest of the code.

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+      return total_bytes;
+    }
+
+    if (llring_empty(f->queue) && !f->next_packet) {
+      f->deficit = 0;
+    }
+
+    // if the flow doesn't have any more packets to give, reenqueue it
+    if (!f->next_packet || f->next_packet->total_len() > f->deficit) {
+      *err = llring_sp_enqueue(flow_ring_, f);
```
melvinw: nit: Is there any reason you're using llring_s{p,c}_{en,de}queue instead of llring_{en,de}queue throughout the code? llring_enqueue()/llring_dequeue() both check how the ring was initialized and pick the correct enqueue/dequeue function. If it doesn't impose too much overhead I'd suggest using the latter instead. The code is slightly easier to maintain this way (though I doubt we'd ever switch to mp or mc).

core/modules/drr.cc
```diff
@@ -0,0 +1,392 @@
...
+
+#include <string>
+#include <cmath>
+#include <iostream>
+#include <fstream>
+
+
+#include "../mem_alloc.h"
+
+int RoundToPowerTwo(uint32_t v) {
```
melvinw: This is defined in `core/utils/common.h` as `align_ceil_pow2`. If fields like `max_flow_queue_size_` can be changed to uint64_t, I recommend using it instead.

TheJStone: I use `RoundPowerTwo` for the llring which expects power of 2 in `uint32_t` for size. so, to keep everything in scope, I have been using `uint32_t` for `max_flow_queue_size` and `max_number_flows_`.

TheJStone: On that note, I notice that the return type is wrong.

melvinw: Ah, gotcha. Ignore me then.

justinemarie:
image

justinemarie:
Great job, Josh!

PR #304, opened 2017-02-16
Add comments to modules

justinemarie:
Yikes! That's a lotta words.


changlan:
Just a reminder that this PR should be squashed.

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
 message L2ForwardCommandLookupResponse {
-  repeated uint64 gates = 1;
-}
-
+  repeated uint64 gates = 1; /// The gate ID that the requested MAC address maps to
+}
+
+/**
+ * The L2Forward module has a command `populate(...)` which allows for fast creation
+ * of the forwarding table given a range of MAC addresses. The funciton takes in a
```
barath: funciton -> function

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
-  uint64 bits = 3;
-  uint64 total_latency_ns = 4;
-  uint64 latency_min_ns = 5;
-  uint64 latency_avg_ns = 6;
-  uint64 latency_max_ns = 7;
-  uint64 latency_50_ns = 8;
-  uint64 latency_99_ns = 9;
-}
-
+  double timestamp = 1; /// What time it is now?
```
barath: What format is this in?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
 
+/**
+ * The module WildcardMatch has a command `add(...)` which inserts a new rule into the WildcardMatch module. For an example of code using WilcardMatch see `bess/bessctl/conf/samples/wildcardmatch.bess`.
+ */
 message WildcardMatchCommandAddArg {
-  uint64 gate = 1;
-  int64 priority = 2;
-  repeated uint64 values = 3;
-  repeated uint64 masks = 4;
+  uint64 gate = 1; /// The gate to direct traffic matching this rule to.
```
barath: Maybe rephrase as "Traffic matching this new rule will be sent to this gate."

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
-    uint32 src_port = 3;  // Source port: 80. Wildcard if 0.
-    uint32 dst_port = 4;  // Destination port. Wildcard if 0.
-    bool established = 5; // Not implemented
-    bool drop = 6;        // Drop matched packets if true.
+    string src_ip = 1;    /// Source IP block in CIDR. Wildcard if "".
+    string dst_ip = 2;    /// Destination IP block in CIDR. Wildcard if "".
+    uint32 src_port = 3;  /// TCP/UDP source port. Wildcard if 0.
+    uint32 dst_port = 4;  /// TCP/UDP Destination port. Wildcard if 0.
+    bool established = 5; /// Not implemented
+    bool drop = 6;        /// Drop matched packets if true.
```
barath: Just to clarify -- if the default behavior of the ACL module is to drop, and the drop field is not set for a rule, does that mean that matching traffic is forwarded?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
   }
-  repeated Rule rules = 1;
+  repeated Rule rules = 1; ///A list of ACL rules.
 }
 
+/**
+ * The BPF module is an access control module that sends packets our on a particular gate based on whether they match a BPF filter.
```
barath: our -> out

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+ */
 message ExactMatchArg {
+  /**
+   * An ExactMatch Field specifies a field over which to check for ExactMatch rules. Field may be in EITHER the packet's data OR it's metadata attributes.
+   */
   message Field {
-    uint64 size = 1;
-    uint64 mask = 4;
+    uint64 size = 1; /// The length, in bytes, of the field to inspect.
+    uint64 mask = 4; /// A bitmask over the field to specify which bits to inspect (default 0xff).
```
barath: Is this the decimal representation of the mask?  If so, what byte ordering is it in (I assume host byte order)?  Also, does the fact that the mask is a uint64 mean that the size field above is implicitly limited to 64 bits?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
-  string duration = 6;
-  bool quick_rampup = 7;
-  uint32 ip_src_range = 8;
-  uint32 ip_dst_range = 9;
-  uint32 port_src_range = 10;
-  uint32 port_dst_range = 11;
-}
-
+  bytes template = 1; /// The packet "template". All data packets are derived from this template and contain the same payload.
+  double pps = 2; /// The total number of pps to generate.
```
barath: pps -> packets per second

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+  double flow_rate = 3; /// The number of new flows to create every second. flow_rate must be <= pps.
+  double flow_duration = 4; /// The lifetime of a flow in seconds.
+  string arrival = 5; /// The packet arrival distribution -- must be either "uniform" or "exponential"
+  string duration = 6; /// The flow duration distribution -- must be either "uniform" or "pareto"
+  bool quick_rampup = 7; /// Whether or not to populate the flowgenerator with initial flows (start generating full pps rate immediately) or to wait for new flows to be generated naturall (all flows have a SYN packet).
```
barath: naturall -> naturally

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+  uint32 ip_src_range = 8; /// When generating new flows, FlowGen modifies the template packet by changing the IP src, incrementing it by at most ip_src_range.
+  uint32 ip_dst_range = 9; /// When generating new flows, FlowGen modifies the template packet by changing the IP dst, incrementing it by at most ip_dst_range.
+  uint32 port_src_range = 10; /// When generating new flows, FlowGen modifies the template packet by changing the TCP port, incrementing it by at most port_src_range.
+  uint32 port_dst_range = 11; /// When generating new flows, FlowGen modifies the template packet by changing the TCP dst port, incrementing it by at most port_dst_range.
```
barath: For these ranges, what does it mean for something to be "incremented" at most by a range?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+ */
 message QueueArg {
-  uint64 size = 1;
-  int64 burst = 2;
-  bool prefetch = 3;
-}
-
+  uint64 size = 1; /// The maximum number of packets to store in the queue.
+  int64 burst = 2; /// The maximum number of packets to release from the queue in one batch.
+  bool prefetch = 3; /// Whether or not to prefetch packet data for the next batch after releasing. SANGJIN WHAT IS THE PURPOSE OF THIS?
```
barath: @sangjinhan -- your wisdom is required here.

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+ */
 message SetMetadataArg {
+  /**
+   * SetMetadata Attribute describes a metadata attribute and value to attach to every packet.
+   */
   message Attribute {
-    string name = 1;
-    uint64 size = 2;
+    string name = 1; /// The metadata attribute name.
+    uint64 size = 2; /// The size of values stored in this attribute.
```
barath: In bytes?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
     oneof value {
-      uint64 value_int = 3;
-      bytes value_bin = 4;
+      uint64 value_int = 3; /// An integer value to store in the packet.
```
barath: What byte order is the field vs. how is it stored in the packet?  I assume both are host byte order, which is fine, except that this will have curious interactions with modules that apply metadata to packets.

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
 
+/**
+ * The Source module generates packets with no payload contents.
+ */
 message SourceArg {
-  uint64 pkt_size = 1;
-  uint64 burst = 2;
-}
-
+  uint64 pkt_size = 1; /// The size of packet data to produce.
```
barath: in bytes

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+ */
 message SplitArg {
-  uint64 size = 1;
-  string name = 2;
-  int64 offset = 3;
-}
-
+  uint64 size = 1; /// The size of the value to read in bytes
+  string name = 2; /// The name of the metadata field to read.
+  int64 offset = 3; /// The offset of the data field to read.
```
barath: in bytes?

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
+  /**
+   * Update Field describes where in a packet's data to rewrite, and with what value.
+   */
   message Field {
-    int64 offset = 1;
-    uint64 size = 2;
-    uint64 value = 3;
+    int64 offset = 1; /// The offset in the packet in bytes to rewrite at.
+    uint64 size = 2; /// The number of bytes to rewrite.
+    uint64 value = 3; /// The value to write into the packet.
```
barath: Why is this a uint64 rather than a bytes field?

protobuf/module_msg.proto
sangjinhan: When `prefetch` is enabled, the module will perform CPU prefetch on the first 64B of each packet onto CPU L1 cache. This is just an optimization knob; whether it is enabled does not affect functionality. This option can be useful if it is already known that the packet data is cache-cold or is in another core's private cache.

Default value is `false`.

All Queue, PortInc, and QueueInc share the same semantics for this `prefetch` parameter.

protobuf/module_msg.proto
sangjinhan: How about alphabetically sort messages? BPFArg should be the first in that sense.

protobuf/module_msg.proto
```diff
@@ -6,385 +6,869 @@ package bess.pb;
...
 }
 
+/**
+ * The NAT module implements address translation, rewriting packet source addresses
+ * for a specified internal prefix with IPs according to a specified
+ * external prefix. To see an example
+ * of NAT in use, see [`bess/bessctl/conf/samples/nat.bess`](https://github.com/NetSys/bess/blob/master/bessctl/conf/samples/nat.bess)
+ *
+ * __Input Gates__: 1
+ * __Output Gates__: 1
```
changlan: There are 2 input gates and 2 output gates.

protobuf/module_msg.proto
justinemarie: @melvinw ?

protobuf/module_msg.proto
justinemarie: @sangjinhan ?

protobuf/module_msg.proto
justinemarie: @sangjinhan @changlan ?

protobuf/module_msg.proto
melvinw: Seconds, I think

melvinw: Rather, it's `tsc * tsc frequency`. So, more like seconds since boot

protobuf/module_msg.proto
changlan: The maximal size of the value is hard-coded to be 8 bytes, I think.

justinemarie: Okay.

protobuf/module_msg.proto
sangjinhan: Yes, size must between 1 and 8. At least for now.

Endianness is about how integers are stored in memory, and it has nothing to do with actual "value". Since protobuf is an abstract message layer, just abstract "value" should be used, not physical "representation".

This can be quite confusing, but the rule of thumb is like this: "**when integral types are used, one should not prematurely consider endianness**". For example if you want to match packets with privileged port numbers (< 1024), then the field definition should be `size == 2, mask == 0xfc00` (not `0x00fc`!), and the corresponding entry should be `0x0000'.

```
(suppose port number in the packet is 500 == 0x1f4)
                    Packet     Mask      entry A
representation      01 f4
value               0x01f4  &  0xfc00 == 0x0000     (matched with entry A!)
```

justinemarie:
I think this can be merged?

justinemarie:
oh wait no conflicts

justinemarie:
All better.

melvinw:
Sweet. Will merge once travis finishes

changlan:
This PR mistakenly adds `ansible-playbook.yml`. It should be removed?

melvinw:
Thanks! 
happycat

PR #305, opened 2017-02-16
Consolidate build scripts

changlan:
Now uses the same Ansible script for vagrant and Docker image


sangjinhan:
Would it be possible to reboot the VM (for Vagrant), before building BESS? While the latest version of Linux image/header is installed, building BESS fails since it is still running an older version until the system reboots.

sangjinhan:
There seems some redundancy between init.yml and bess.yml: bind mount and hugepage sysctl

sangjinhan:
One clarification question: bess.yml is for Docker & bare-metal cases, and init.yml is only used for Vagrant, right? If so, maybe we can rename init.yml to vm.yml or vagrant.yml, unless init.yml is a specially predefined filename?

sangjinhan:
https://github.com/NetSys/bess/wiki/Build-and-Install-BESS will also needs to be updated, once it is merged.

changlan:
- AFAIK there is no clean way to reboot the VM when provisioning. Is it possible to refresh the version without rebooting?

- Copying those files ensures those configs to be active after system reboots. If it is not required, I can simply remove those files. (Dev VM seems to be working fine without them)

sangjinhan:
One workaround for the kernel version issue is to install both linux-headers-generic (for the latest one) and linux-headers-`uname -r` (for the running kernel)? The old ansible-playbook.yml file did this.

It seems that the Ansible modules for sysctl and mount make permanent changes, so should survive reboots.

sangjinhan:
A few comments/questions regarding the latest version (mostly about Vagrant and bare-metal usage scenarios):
* BESS is not built automatically. Is it intended? (I am fine with this, just checking)
* Perhaps at the end of the Vagrant script we can reboot the system with `config.vm.provision :reload`?
  * This plugin is already installed at the beginning of Vagrantfile: https://github.com/aidanns/vagrant-reload
* A user wouldn't know where BESS (mounted at the /opt/bess) is. Maybe we can do one of the followings...
  * Create a symbolic link or do bind-mount in /home/vagrant to /opt/bess
  * Change home directory to /opt/bess
  * Make BESS available at /home/vagrant/bess, not /opt/bess
  * Show some message at the end of vagrant provisioning.
    * e.g., `Run 'vagrant ssh' to log into the VM. BESS can be found at /opt/bess."

sangjinhan:
Oh, one last thing: it would be safe to use 2 vCPUs by default. Some people said launching the VM on  a dual-core system made the system almost unusable.

changlan:
It is intentional, because it could take so long (without any output messages) that the user might think the VM hangs. Done with other stuff.

sangjinhan:
Works perfect. Thanks!

PR #306, opened 2017-02-17
Add queries to link status

sangjinhan:
This is a C++ port of #109, although it has been million years late...

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -899,28 +899,21 @@ class BESSControlImpl final : public BESSControl::Service {
 
   Status GetPortStats(ServerContext*, const GetPortStatsRequest* request,
                       GetPortStatsResponse* response) override {
-    const char* port_name;
-    port_stats_t stats;
-
-    if (!request->name().length())
-      return return_with_error(response, EINVAL,
-                               "Argument must be a name in str");
-    port_name = request->name().c_str();
-
-    const auto& it = PortBuilder::all_ports().find(port_name);
+    const auto& it = PortBuilder::all_ports().find(request->name());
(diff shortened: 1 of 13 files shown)


core/bessctl.cc
```diff
@@ -908,6 +926,33 @@ class BESSControlImpl final : public BESSControl::Service {
 
     return Status::OK;
   }
+
+  Status GetLinkStatus(ServerContext*, const GetLinkStatusRequest* request,
+                       GetLinkStatusResponse* response) override {
+    const char* port_name;
+    ::Port::LinkStatus status;
+
+    if (!request->name().length())
```
barath: Braces.

protobuf/bess_msg.proto
```diff
@@ -151,6 +151,18 @@ message ListWorkersResponse {
   repeated WorkerStatus workers_status = 2;
 }
 
+message GetLinkStatusRequest {
```
barath: Could you add protobuf comments that follow the style of Justine's?

sangjinhan: I believe the message name is self describing... Justine's comments are about module-specific messages (module_msg.proto), not system-wide messages.

sangjinhan:
Applied Barath's comments. Did a little bit of code polishing as well.

protobuf/bess_msg.proto
barath: Ok.  I could go either way on this.  In a narrow sense it is self descriptive but a) it's nice to have everything documented for consistency so no guesswork is required and the meaning of Link and Status aren't actually totally obvious in the BESS context.

melvinw:
Thanks! cat

PR #307, opened 2017-02-17
Add ability to change MAC for vport

sangjinhan:
VPort driver now allows changing MAC address via SIOC* ioctls.
(e.g., `ip link set <ifname> address xx:xx:xx:xx:xx:xx`, once the port is down state)


barath:
img_2066

PR #309, opened 2017-02-17
Fixes the behavior of TC tree size counting so as to avoid triggering…

barath:
… a race.

### DIFF
--- a/core/traffic_class.h
+++ b/core/traffic_class.h
@@ -1,5 +1,5 @@
-#ifndef BESS_TRAFFICCLASS_H_
-#define BESS_TRAFFICCLASS_H_
+#ifndef BESS_TRAFFIC_CLASS_H_
+#define BESS_TRAFFIC_CLASS_H_
 
 #include <deque>
 #include <list>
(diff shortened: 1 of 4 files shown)


melvinw:
Looks good. Just one question. Why store sizes in `TrafficClassBuilder::tc_count` instead of as a member of `TrafficClass`?

core/traffic_class.h
```diff
@@ -124,13 +124,9 @@ class TrafficClass {
...
 
-  size_t Size() const {
-    size_t sz = 0;
-    Traverse([](const TrafficClass *,
-                void *arg) { *reinterpret_cast<size_t *>(arg) += 1; },
-             static_cast<void *>(&sz));
-    return sz;
-  }
+  // Returns the number of TCs in the TC subtree rooted at this, including this
+  // TC.  Returns 0 upon error.
```
sangjinhan: Is it ever possible (in normal situations) to return 0? If not, shouldn't it be CHECK() or DCHECK()?

barath:
I didn't store it in TrafficClass to minimize unnecessary things being held
in TrafficClass objects which are on the fast path.

core/traffic_class.h
barath: Agreed.  Would be happy to change it to CHECK.

melvinw:
Gotcha. Fine by me.

core/traffic_class.h
sangjinhan: Yes please 👍

melvinw:
Thanks! 
5ih45x7

PR #310, opened 2017-02-18
Do not use exit() in a child process

sangjinhan:
gRPC crashes when a child process from frok() terminates with exit(). Some of its global (static) objects incorrectly free up resources (more specifically, threads) that are still being used by the parent process.

When VPort is created for a network namespace, the driver internally uses fork() so that its child process can attach to the namespace and set up IP addresses as specified. This patch replace exit() done in the child process with _exit(), so that the offending object destructors can be skipped.


barath:
img_0164

PR #311, opened 2017-02-18
Fix errors with module command arguments in sample scripts

sangjinhan:
We do not support `module_foo.command_bar(xxx, yyy)` syntax any longer since we switched to protobuf. Only keyword arguments are supported, as in `module_foo.command_bar(a=xxx, b=yyy)`. This PR updates outdated sample scripts in that regard.


barath:
img_0034

PR #315, opened 2017-02-23
Dynamically load modules

changlan:
- Compile modules into standalone dynamic libraries
- Load those modules in the `modules` directory at runtime.

### DIFF
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -1,5 +1,7 @@
 #include "bessd.h"
 
+#include <dirent.h>
+#include <dlfcn.h>
 #include <sys/file.h>
 #include <sys/resource.h>
 #include <sys/stat.h>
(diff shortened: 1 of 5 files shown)


core/bessd.cc
```diff
@@ -360,5 +362,47 @@ bool SetResourceLimit() {
...
+      }
+    }
+  }
+  closedir(dir);
+  return true;
+}
+
+std::string GetCurrentDirectory() {
+  char dest[PATH_MAX];
+  if (readlink("/proc/self/exe", dest, PATH_MAX) == -1) {
```
barath: readlink() doesn't guarantee null termination, so dest could either fill up to PATH_MAX or not but still not have a '\0' at the end, and strrchr() won't be able to find the end.

core/bessd.cc
```diff
@@ -360,5 +362,47 @@ bool SetResourceLimit() {
   }
 }
 
+static inline bool has_suffix(const std::string &s, const std::string &suffix) {
```
barath: Function comment.

core/main.cc
```diff
@@ -21,6 +21,8 @@ int main(int argc, char *argv[]) {
   google::ParseCommandLineFlags(&argc, &argv, true);
   bess::bessd::ProcessCommandLineArgs();
 
+  bess::bessd::LoadModules(bess::bessd::GetCurrentDirectory() + "modules");
```
barath: If this returns false, should it be fatal?

core/bessd.cc
```diff
@@ -360,5 +362,47 @@ bool SetResourceLimit() {
...
+}
+
+bool LoadModules(const std::string &directory) {
+  DIR *dir = opendir(directory.c_str());
+  if (!dir) {
+    return false;
+  }
+  dirent *entry;
+  while ((entry = readdir(dir)) != nullptr) {
+    if (has_suffix(entry->d_name, ".so")) {
```
barath: May also want to check d_type.

changlan:
kat plz? :]

sangjinhan:
^ↀᴥↀ^ meow

PR #317, opened 2017-02-26
Faster unit test for TcpFlowReconstruct

sangjinhan:
Unit test `TcpFlowReconstructTest.ReorderedReconstruction` takes around 15 seconds. This PR reduces the delay down to 1ms.

* Optimize `TcpFlowReconstruct` by using a sorted list of TCP segments ([start_offset, end_offset]), rather than maintaining a bitmap.
* Do not consider pure ACK packets when generating N! permutations for reordering test.

### DIFF
--- a/core/utils/tcp_flow_reconstruct.h
+++ b/core/utils/tcp_flow_reconstruct.h
@@ -19,39 +19,30 @@ namespace utils {
 class TcpFlowReconstruct {
  public:
   // Constructs a TCP flow reconstruction object that can hold initial_buflen
-  // bytes to start with.  If initial_buflen is zero, it is automatically set to
-  // 1.
-  TcpFlowReconstruct(uint32_t initial_buflen)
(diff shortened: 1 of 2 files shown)


core/utils/tcp_flow_reconstruct.h
```diff
@@ -97,25 +87,48 @@ class TcpFlowReconstruct {
...
+    // new segment                           |-------------|
+    // existing segments with a hole   |---A---|   |--B--|-C-|
+    //                                             ^
+    //                                             lower_bound(start)
+    {auto it = received_map_.lower_bound(start);
+    if (it != received_map_.begin()) {
+      auto it_prev = it;
+      it_prev--;
+      if (it_prev->second >= start) {
+        // case A. include it.
```
barath: This comment is a bit cryptic.

barath:
p2252470

PR #318, opened 2017-02-26
Fix deinit crash

sangjinhan:
When module Init() fails, the module's DeInit() function will be called immediately. This behavior is different from the legacy C version, where DeInit() was skipped for half-initialized modules. Some modules' DeInit() functions crash in this case.

How to reproduce:
```bessctl add module PortInc```
(module `Init()` fails since `port` is not specified, then `DeIniit()` crashes with a null pointer access)


barath:
img_1159

PR #319, opened 2017-02-26
Remove RoundRobin module test

sangjinhan:
round_robin_test.cc is the only per-module unit test at the moment, although it doesn't perform any useful testing. This PR removes the file to avoid a false sense of confidence. Justine's module test framework would be a more effective for module test, at least for macroscopic ones.


changlan:
image

PR #320, opened 2017-02-27
Suppress scapy IPv6 warning

sangjinhan:
Scapy can generate this warning when it is imported:

```
WARNING: No route found for IPv6 destination :: (no default route?)
```

This warning message is suprious in that it only matters using Scapy send() function for IPv6 packets, whereas bessctl sample scripts use Scapy for packet crafting, not packet injection.


melvinw:
Thanks! cat

PR #321, opened 2017-02-27
Do not swap endianness for value_bin

sangjinhan:
SetMetadata takes two different types of values: `value_int` and `value_bin`. When value_bin is given to the module, the data should be used verbatim. At the moment, `value_bin` is only used by the `samples/vxlan` example, which does not work correctly due to incorrect metadata values. This PR fixes this issue by not converting endianness for `value_bin` data.

This PR also fixes an IPEncap bug found during the process. The module occasionally generates incorrect IP checksum values, depending on the garbage values in the IP checksum field. DPDK checksum routines require the field to be set to zero before calculating checksum. Switched to utils/checksum.h which does not impose this limitation.


barath:
img_1687

PR #325, opened 2017-02-28
Import module library from CLI

changlan:


bessctl/commands.py
```diff
@@ -269,6 +269,11 @@ def get_var_attrs(cli, var_token, partial_word):
             var_desc = 'configuration filename'
             var_candidates = complete_filename(partial_word)
 
+        elif var_token == 'LIB_FILE':
+            var_type = 'filename'
+            var_desc = 'library filename'
+            var_candidates = complete_filename(partial_word)
```
sangjinhan: How about something like:
```
var_candidates = complete_filename(partial_word, suffix='.so', skip_suffix=True)
```
so that only *.so files can be considered for autocompletion? (`skip_suffix` is not implemented yet btw)

bessctl/commands.py
```diff
@@ -269,6 +269,11 @@ def get_var_attrs(cli, var_token, partial_word):
             var_desc = 'configuration filename'
             var_candidates = complete_filename(partial_word)
 
+        elif var_token == 'LIB_FILE':
+            var_type = 'filename'
+            var_desc = 'library filename'
```
sangjinhan: Perhaps "plugin" sounds better than "library"...?

sangjinhan: Adding `(*.so)` to the description might be helpful.

bessctl/commands.py
```diff
@@ -1190,6 +1195,10 @@ def show_mclass_list(cli, cls_names):
         _show_mclass(cli, cls_name, True)
 
 
+@cmd('import mclass LIB_FILE', 'Import the specified module library')
```
sangjinhan: We can envision that in the future we may be able to import multiple modules, drivers, hooks, scheduler policies, etc. from a single .so file package. In that sense, `import plugin PLUGIN_FILE` might be more general. Just `import PLUGIN_FILE` would also work, but there could be different types of "imports" to be added to bessctl (e.g., configuration)

sangjinhan:
Merged, thanks!

PR #326, opened 2017-03-01
Add support for function callstack in *.so files

sangjinhan:
While modules now can be loaded at runtime with shared objects, the current function backtrace (crash log) does not correctly locate source code from instruction pointers. This is because .so files can be loaded at an arbitrary memory location; the base address of .so file must be considered when corresponding memory addresses to position in the original ELF file.

Before:
```
...
(0): /home/sangjin/bess/core/modules/source.so(_ZN6Source7RunTaskEPv+0x55) [0x7f09e912c7b5]
    ?? ??:0
(1): /home/sangjin/bess/core/bessd(_ZN4Task9ScheduledEv+0x14) [0x6031a4]
    Task::Scheduled() at /home/sangjin/bess/core/task.cc:36
      -> 36:   struct task_result ret = m_->RunTask(arg_);
...
```

After:
```
...
(0): /home/sangjin/bess/core/modules/source.so(_ZN6Source7RunTaskEPv+0x185) [0x7f7da301f8a5]
    Source::RunTask(void*) at /home/sangjin/bess/core/modules/source.cc:81
         78:
         79:   int cnt = bess::Packet::Alloc(batch.pkts(), burst, pkt_size);
         80:
      -> 81:   *(int *)0 = 0;  // DIE
         82:
         83:   if (cnt > 0) {
         84:     batch.set_cnt(cnt);
(1): /home/sangjin/bess/core/bessd(_ZN4Task9ScheduledEv+0x14) [0x60d3a4]
    Task::Scheduled() at /home/sangjin/bess/core/task.cc:36
      -> 36:   struct task_result ret = m_->RunTask(arg_);
...
```

### DIFF
--- a/core/debug.cc
+++ b/core/debug.cc
@@ -1,7 +1,9 @@
 #include "debug.h"
 
+#include <dlfcn.h>
 #include <execinfo.h>
 #include <gnu/libc-version.h>
+#include <link.h>
 #include <sys/syscall.h>
 #include <ucontext.h>
 #include <unistd.h>
@@ -14,17 +16,15 @@
 #include <csignal>
 #include <cstdint>
 #include <cstdio>
-#include <cstdlib>
-#include <cstring>
 #include <fstream>
 #include <sstream>
-#include <stack>
 #include <string>
 
 #include "module.h"
 #include "packet.h"
 #include "scheduler.h"
 #include "traffic_class.h"
+#include "utils/format.h"
 
 namespace bess {
 namespace debug {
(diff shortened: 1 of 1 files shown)


core/debug.cc
```diff
@@ -139,125 +139,113 @@ static const char *si_code_to_str(int sig_num, int si_code) {
...
+      size_t bytes = fread(buf, 1, sizeof(buf), proc);
+      ret.write(buf, bytes);
+    }
+    pclose(proc);
   }
 
-  for (;;) {
-    char line[1024];
-    char *p;
+  return ret;
```
changlan: `std::stringstream` is not copyable, and its move constructor is not implemented until GCC 5 (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54316). Perhaps return a `string` instead?

sangjinhan: Oh, good point. Will fix.

changlan:
image

PR #334, opened 2017-03-16
drivers/pmd: Fix crash when driver_name is NULL.

ddiproietto:
It currently happens with DPDK ena pmd.  It should probably be fixed in
DPDK as well.


barath:
img_1686

justinemarie:
Is this Daniele's first cat? HAPPY CAT-DAT DANIELE!!!

PR #336, opened 2017-03-16
Minor tweak to Makefile

changlan:
1. Add root directory as default directory for headers
2. Add protobuf as modules' dependency


justinemarie:
Thanks!
image

Issue #335, opened 2017-03-16
Command 'interactive' does not work


nemethf:
It seems some part of #24 hasn't been pulled correctly. 
Reapplying the following hunk of @MurphyMc, the problem disappears.
```
diff --git a/bessctl/cli.py b/bessctl/cli.py
index 2ea19a2..6d58b98 100644
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -34,7 +34,16 @@ class CLI(object):
         self.last_cmd = ''

         self.interactive = interactive
-        if not interactive:
+        self.rl = None
+
+        self.maybe_go_interactive()
+
+    def maybe_go_interactive(self):
+        if self.interactive: return
+
+        self.interactive = self.fin.isatty() and self.fout.isatty()
+
+        if not self.interactive:
             self.rl = None
             return
```

without the patch:
```
 ./bessctl interactive
Traceback (most recent call last):
  File "./bessctl", line 202, in <module>
    run_cmds(cStringIO.StringIO('\n'.join(cmds)))
  File "./bessctl", line 178, in run_cmds
    cli.loop()
  File "./bessctl", line 112, in loop
    super(BESSCLI, self).loop()
  File "/export/netsys/src/bess/bessctl/cli.py", line 452, in loop
    self.process_one_line()
  File "/export/netsys/src/bess/bessctl/cli.py", line 394, in process_one_line
    self.call_func(func, args)
  File "./bessctl", line 64, in call_func
    super(BESSCLI, self).call_func(func, args)
  File "/export/netsys/src/bess/bessctl/cli.py", line 369, in call_func
    func(*args)
  File "/export/netsys/src/bess/bessctl/commands.py", line 1532, in interactive
    cli.maybe_go_interactive()
AttributeError: 'BESSCLI' object has no attribute 'maybe_go_interactive'
```

with the patch:
```
$ ./bessctl interactive
Type "help" for more information.
localhost:10514 $ 
```
or in more useful example:
```
$ ./bessctl show pipeline -- interactive
+-------------------+                      +----------+                      +-------------------+
|     port_inc0     |                      | macswap0 |                      |     port_out0     |
|      PortInc      |  :0 52231913504 0:   | MACSwap  |  :0 52231718560 0:   |      PortOut      |
| pmd_port0/PMDPort | -------------------> |          | -------------------> | pmd_port0/PMDPort |
+-------------------+                      +----------+                      +-------------------+
Type "help" for more information.
localhost:10514 $
```

[...]

PR #337, opened 2017-03-17
Restore interactive command line option.

justinemarie:
See https://github.com/NetSys/bess/issues/335#issuecomment-287471292

### DIFF
--- a/bessctl/bessctl
+++ b/bessctl/bessctl
@@ -132,7 +132,8 @@ def run_cli():
                          'Try "daemon start".'
 
     cli = BESSCLI(s, commands, ferr=sys.stderr, interactive=True)
-    cli.loop()
+    if cli.interactive:
+        cli.loop()
 
 
 def run_cmds(instream):
--- a/bessctl/bessctl
+++ b/bessctl/bessctl
@@ -127,9 +127,9 @@ def run_cli():
         s = BESS()
         s.connect()
     except BESS.APIError as e:
-        print >> stderr, e.message
-        print >> stderr, 'Perhaps bessd daemon is not running locally? ' \
-                         'Try "daemon start".'
+        print >> sys.stderr, e.message
+        print >> sys.stderr, 'Perhaps bessd daemon is not running locally? ' \
+                             'Try "daemon start".'
 
     cli = BESSCLI(s, commands, ferr=sys.stderr, interactive=True)
     if cli.interactive:
--- a/bessctl/bessctl
+++ b/bessctl/bessctl
@@ -122,39 +122,28 @@ class BESSCLI(cli.CLI):
         return '<disconnected> $ '
 
 
-def run_cli():
+def run_cli(instream=sys.stdin):
+    s = BESS()
+    cli = BESSCLI(s, commands, fin=instream)
+
     try:
-        s = BESS()
         s.connect()
     except BESS.APIError as e:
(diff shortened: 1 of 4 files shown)


justinemarie:
Recall that I literally don't know python so keep a watchful eye ;-)

bessctl/cli.py
```diff
@@ -441,6 +430,52 @@ def restore_echoctl(self):
         except:
             pass
 
+    def go_interactive(self):
```
MurphyMc: It was long enough ago that my memory is foggy, but IIRC, my earlier version was "maybe_go_interactive" because there was at least one case where it didn't: when you already *were* in interactive mode.  The idea being that actually typing "interactive" when you're already in interactive mode should do nothing rather than actually try to go into interactive mode again.

bessctl/bessctl
```diff
@@ -161,8 +131,7 @@ def run_cli():
         print >> stderr, 'Perhaps bessd daemon is not running locally? ' \
```
ddiproietto: Maybe I'm doing something wrong, I get an error about stderr not being defined here

bessctl/cli.py
```diff
@@ -441,6 +430,52 @@ def restore_echoctl(self):
...
+        self.fin = sys.stdin
+        self.fout = sys.stdout
+        self.interactive = sys.stdin.isatty() and sys.stdout.isatty()
+
+        # Colorize output to standard error
+        if self.interactive and sys.stderr.isatty():
+            stderr = ColorizedOutput(sys.stderr, '\033[31m')  # red (not bright)
+        else:
+            stderr = sys.stderr
```
ddiproietto: Again, maybe I'm doing something wrong, but unless I add

self.ferr = stderr

I don't get colors

bessctl/bessctl
justinemarie: Fixed. Thanks!

bessctl/cli.py
justinemarie: I decided to keep "go_interactive" because "maybe_go_interactive" suggests that I can call this function when not in interactive mode and come out still not in interactive mode. I did add a check that if I am already in interactive mode, I should return and not, e.g., re-load the history.

justinemarie:
Thanks for the comments all! Any more?

bessctl/cli.py
MurphyMc: Yeah, I agree that the name could be a bit misleading.  The actual semantics -- that it doesn't try to go into interactive if it's already in interactive -- was certainly the more important aspect!

justinemarie: So actually I went back on this, because if there's no TTY, it doesn't go into interactive mode -- so it *is* maybe! I changed the name as you suggested, ha.

justinemarie:
@ddiproietto give me a cat!

sangjinhan:
Made a few changes:
* Colorize output to stderr even in non-interactive mode, as long as stderr is a TTY and not explicitly forced not to do so.
* When bessctl is piped (e.g., `bessctl < commands.txt`), it should not run in interactive mode.
* `bessctl some command -- interactive -- another command` will run `another command` after the interactive mode is done.

sangjinhan:
@ddiproietto, please take a quick look and merge (and give her cats)

ddiproietto:
gatto

Issue #332, opened 2017-03-13
kmod doesn't compile on kernel > 4.7


ddiproietto:
It seems that after 4.7 paravirt_enabled() was removed:

https://github.com/torvalds/linux/commit/867fe800b4c423bce46e66ccb2ce91bebbd5afc6

Not sure how to replace it.

Thanks!

PR #344, opened 2017-03-20
Drop support for Guest/IVSHMEM

sangjinhan:
Once upon a time, the kernel module supported virtual ports for VM guests with IVSHMEM. Now it's time for the guest mode to retire, since now DPDK supports virtio.


sangjinhan:
As a side effect, this PR addresses #332.

barath:
img_2074

PR #348, opened 2017-03-23
Utility Codel

TheJStone:
Codel(Controlled Delay Management) is an Queue controller based on this
article http://queue.acm.org/detail.cfm?id=2209336

It provides an active queue management to help prevent bufferbloat by dropping
queue entries at an increasing rate if the delay in the queue is above the
target queue delay. The equation used to calculate drop intervals is based on TCP
throughput response to drop probability.

### DIFF
--- a/core/utils/codel.h
+++ b/core/utils/codel.h
@@ -1,0 +1,241 @@
+#ifndef BESS_UTILS_CODEL_H_
+#define BESS_UTILS_CODEL_H_
+
+#include <cmath>
+#include <deque>
+
+#include <glog/logging.h>
+
+#include "time.h"
+#include "queue.h"
+
+
+namespace bess {
+namespace utils {
+// Codel(Controlled Delay Management) is an Queue controller based on this
+// article http://queue.acm.org/detail.cfm?id=2209336
+
+// It provides an active queue management to help prevent bufferbloat by dropping
+// queue entries at an increasing rate if the delay in the queue is above the
+// target queue delay. The equation used to calculate drop intervals is based on TCP
+// throughput response to drop probability.
+
+// template argument T is the type that is going to be enqueued/dequeued.
+template <typename T>
+class Codel final: public Queue<T> {
+ public:
+  // default delay target for codel
+  static const uint64_t kDefaultTarget = 5000000; 
+  // default window size for codel
+  static const uint64_t kDefaultWindow = 100000000; 
+  // default number of slots in the codel queue 
+  static const int kDefaultSlots = 4096; 
+
+  typedef std::pair<uint64_t, T> Wrapper;
+
+  // Takes a drop function which is a function that should take a dropped object 
+  // and handle it removing the object potentially including freeing the object.
+  // If there is no need to handle a dropped object, NULL can be passed instead.
+  // target is the target delay in nanoseconds and the window is the buffer time u
+  // in nanosecond before changing into drop state.
+  Codel(void (*drop_func)(T)= NULL, size_t max_entries=0, uint64_t target = kDefaultTarget, 
+      uint64_t window = kDefaultWindow)
+      : delay_target_(target),
+        window_(window),
+        time_above_target_(0),
+        next_drop_time_(NanoSecondTime() + window),
+        drop_count_(0),
+        dropping_(0),
+        max_size_(max_entries),
+        queue_(),
+        drop_func_(drop_func) { }
+
+  virtual ~Codel() {
+    Wrapper w;
+    while (!queue_.empty()) {
+      Drop(queue_.front());
+      queue_.pop_front();
+    }
+  }
+
+  int Push(T obj) override {
+    if (max_size_ != 0 && queue_.size() >= max_size_) {
+      return -1;
+    }
+    Wrapper w(NanoSecondTime(), obj);
+    queue_.push_back(w);
+    return 0;
+  }
+
+  int Push(T* ptr, size_t count) override {
+    size_t i = 0;
+    for (; i < count; i++) {
+      if (Push(ptr[i])) {
+        break;
+      } 
+    }
+    return i;
+  }
+
+  // Retrieves the next entry from the queue and in the process, potentially drops
+  // objects as well as changes between dropping state and not dropping state.
+  int Pop(T &obj) override {
+    bool drop = false;
+    Wrapper w;
+    int err = RingDequeue(w, drop);
+    if (err != 0) {
+      dropping_ = 0;
+      return -2;
+    }
+
+    uint64_t now = NanoSecondTime();
+    if (dropping_) {
+      // if in dropping state, drop object until next drop time is greater
+      // than the current time.
+      err = DropDequeue(w, drop);
+    } else if (drop && ((now - next_drop_time_ < window_) ||
+                           (now - time_above_target_ >= window_))) {
+      // if not in dropping state, determine whether to enter drop state and if
+      // so, drop current object, get a new object and reset the drop counter.
+      Drop(w);
+      err = RingDequeue(w, drop);
+
+      if (err == 0) {
+        dropping_ = 1;
+        if (now - next_drop_time_ < window_ && drop_count_ > 2) {
+          drop_count_ -= 2;
+        } else {
+          drop_count_ = 1;
+        }
+        next_drop_time_ = NextDrop(now);
+      }
+    }
+    
+    // if there was a wrapper to dequeue, set the parameter object to the 
+    // wrapper's object
+    if (err == 0) {
+      obj = w.second;
+    } 
+    return err;
+  }
+
+  // Retrieves the next count entries from the queue and in the process, potentially
+  // drops objects as well as changes between dropping state and not dropping state.
+  // Does not necessarily return count if there are count present but some are dropped.
+  int Pop(T* objs, size_t count) override {
+    size_t i = 0;
+    T next_obj;
+    for (; i < count; i++) {
+      int err = Pop(next_obj);
+      if (err != 0) {
+        break;
+      }
+      objs[i] = next_obj;
+    }
+    return i;
+  }
+
+  size_t Capacity() override { return queue_.max_size(); }
+
+  bool Empty() override { return queue_.empty(); }
+
+  bool Full() override { 
+    if (max_size_ != 0) {
+      return max_size_ == queue_.size();
+    }
+    return queue_.size() == queue_.max_size();
+  }
+
+  size_t Size() override { return queue_.size(); }
+
+ private:
+  // Calls the drop_func on the object if the drop function exists
+  void Drop(Wrapper w) {
+    if (drop_func_ != NULL) {
+        drop_func_(w.second);
+    }
+  }
+
+  // Takes the relative time to determine the next time to drop.
+  // returns the next time to drop a object.
+  uint64_t NextDrop(uint64_t cur_time) {
+    return cur_time + window_ * pow(drop_count_, -.5);
+  }
+
+  // Gets the next object from the queue and determines based on current state,
+  // whether set the passed drop boolean to true(to tell the calling function to
+  // drop it). Takes a Wrapper to set to the next entry in the queue and a boolean 
+  // to set if the entry should be dropped. Returns 0 on success.
+  int RingDequeue(Wrapper &w, bool &drop) {
+    if (!queue_.empty()) {
+      w = queue_.front();
+      queue_.pop_front();
+    } else {
+      return -1;
+    }
+
+    uint64_t now = NanoSecondTime();
+    uint64_t delay_time = now - w.first;
+
+    // determine whether object should be dropped or to change state
+    if (delay_time < delay_target_) {
+      time_above_target_ = 0;
+    } else {
+      if (time_above_target_ == 0) {
+        time_above_target_ = now + window_;
+      } else if (now >= time_above_target_) {
+        drop = true;
+      }
+    }
+    return 0;
+  }
+
+  // Called while Codel is in drop state to determine whether to drop the current
+  // entries and dequeue the next entry. Will continue to drop entries until the 
+  // next drop is greater than the current time. Takes a Wrapper which is the next
+  // entry in the queue which will potentially be replaced and a boolean determing 
+  // if the entry should be dropped. Returns 0 on success.
+  int DropDequeue(Wrapper &w, bool &drop) {
+    uint64_t now = NanoSecondTime();
+    if (!drop) {
+      dropping_ = 0;
+    } else if (now >= next_drop_time_) {
+      while (now >= next_drop_time_ && dropping_) {
+        Drop(w);
+        drop_count_++;
+        int err = RingDequeue(w, drop);
+
+        if (err != 0 || !drop) {
+          dropping_ = 0;
+          return err;
+        }
+        next_drop_time_ = NextDrop(next_drop_time_);
+      }
+    }
+    return 0;
+  }
+
+  // Returns the current time in microseconds.
+  uint64_t NanoSecondTime() {
+    return tsc_to_ns(rdtsc());
+  }
+
+  uint64_t delay_target_;  // the delay that codel will adjust for
+  uint64_t window_;        // minimum time before changing state
+
+  // the time at which codel will change state to above target(0 if below)
+  uint64_t time_above_target_;
+  uint64_t next_drop_time_;  // the next time codel will drop
+
+  // the number of objects dropped while delay has been above target
+  uint32_t drop_count_;
+  uint8_t dropping_;       // whether in dropping state(above target for window)
+  size_t max_size_;
+  std::deque<Wrapper> queue_;  // queue
+  void (*drop_func_)(T);  // the function to call to drop a value
+};
+
+}  // namespace utils
+}  // namespace bess
+
+#endif  // BESS_UTILS_CODEL_H_
--- a/core/utils/codel_test.cc
+++ b/core/utils/codel_test.cc
@@ -1,0 +1,240 @@
+#include "codel.h"
+
+#include <cmath>
+#include <ctime>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+using bess::utils::Codel;
+void integer_drop(int* ptr) {
+  delete ptr;
+}
+
+// Simple test to make sure one can get back out object
+TEST(CodelTest, SingleInputOutput) {
+  Codel<int*> c(&integer_drop, 8, 5000000, 100000000);
+  int* val = new int();
+  ASSERT_NE(val, nullptr);
+  ASSERT_FALSE(c.Push(val));
+  ASSERT_EQ(c.Size(), 1);
+  int* output;
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, val);
+  ASSERT_EQ(c.Size(), 0);
+  EXPECT_EQ(output, val);
+  delete val;
+}
+
+// Simple test to make sure one can get back out non-pointer object
+TEST(CodelTest, NonPointerType) {
+  Codel<int> c(NULL, 8, 5000000, 100000000);
+  int val = 3439148;
+  ASSERT_FALSE(c.Push(val));
+  ASSERT_EQ(c.Size(), 1);
+  int output;
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, val);
+  ASSERT_EQ(c.Size(), 0);
+  EXPECT_EQ(output, val);
+}
+
+// Tests to make sure that low delay entries won't be dropped
+TEST(CodelTest, NoDropTest) {
+  Codel<int*> c(&integer_drop, 8, 5000000, 100000000);
+
+  int n = 4;
+  int* vals[n];
+  for (int i = 0; i < n; i++) {
+    vals[i] = new int();
+  }
+  for (int i = 0; i < n; i++) {
+    ASSERT_FALSE(c.Push(vals[i]));
+  }
+  EXPECT_EQ(c.Size(), n);
+
+  usleep(3000);  // low delay should not cause any drops
+  int* output = nullptr; 
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[0]);
+  
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[1]);
+  
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[2]);
+  
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[3]);
+  EXPECT_EQ(c.Size(), 0);
+
+  for (int i = 0; i < n; i++) {
+    delete vals[i];
+  }
+}
+
+// Tests to make sure that codel enters drop state at the proper time and
+// properly drops at the correct time.
+TEST(CodelTest, DropTest) {
+  Codel<int*> c(&integer_drop, 16, 5000000, 100000000);
+
+  int n = 10;
+  int* vals[n];
+  for (int i = 0; i < n; i++) {
+    vals[i] = new int();
+  }
+  for (int i = 0; i < n; i++) {
+    ASSERT_FALSE(c.Push(vals[i]));
+  }
+  EXPECT_EQ(c.Size(), n);
+
+  usleep(10000);  // all objects will have greater delay than target of 5ms
+  int* output = nullptr;
+
+  // should mark above target after dequeuing this entry
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[0]);
+
+  // still shouldn't drop yet because hasn't been interval since
+  // being above target
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[1]);
+  usleep(105000);  // wait above a window of time to make next entry drop
+
+  // Should be above target for window.
+  // should drop vals[2], enter drop state, and return 4th instead.
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[3]);
+
+  // should now in be drop state and the next drop should be 100ms.
+  usleep(72000);
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[4]);  // shouldn't be a drop yet only has been 72ms.
+
+  usleep(30000);
+  ASSERT_FALSE(c.Pop(output));
+  // vals[5] should be dropped now has been over 100ms.
+  ASSERT_EQ(output, vals[6]); 
+  // the next drop should be in approximately 70ms(100/sqrt(2))
+
+  usleep(40000);
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[7]);  // it has only been 40ms should not be a drop yet
+
+  usleep(34000);
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[9]);  // vals[8] should have been dropped
+
+  // codel queue should be empty and return nullptr
+  ASSERT_TRUE(c.Pop(output));
+  EXPECT_EQ(c.Size(), 0);
+
+  delete vals[0];
+  delete vals[1];
+  delete vals[3];
+  delete vals[4];
+  delete vals[6];
+  delete vals[7];
+  delete vals[9];
+}
+
+// Makes sure that codel will exit drop state at the proper time both when the
+// front entry is low delay and when the object that returns after dropping the
+// front entry is low delay.
+TEST(CodelTest, ChangeStateTest) {
+  Codel<int*> c(&integer_drop, 16, 5000000, 100000000);
+  int n = 10;
+  int* vals[n];
+  for (int i = 0; i < n; i++) {
+    vals[i] = new int();
+  }
+
+  for (int i = 0; i < 4; i++) {
+    ASSERT_FALSE(c.Push(vals[i]));
+  }
+  EXPECT_EQ(c.Size(), 4);
+
+  usleep(10000);  // all packets will have greater delay than target of 5ms
+  int* output;
+
+  // should change into mark above target after dequeuing this entry
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[0]);
+
+  usleep(105000);  // wait above a window of time to make next entry drop
+
+  // Should be above target for window.
+  // should drop vals[1], enter drop state, and return 3rd instead.
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[2]);
+
+  // should now in be drop state and the next drop should be 100ms.
+  usleep(104000);
+  ASSERT_FALSE(c.Push(vals[4]));
+  ASSERT_FALSE(c.Pop(output));
+  // vals[3] should be dropped now has been over 100ms.
+  ASSERT_EQ(output, vals[4]);
+  // val4 had a delay below target so should have left drop state.
+
+  for (int i = 5; i < 8; i++) {
+    ASSERT_FALSE(c.Push(vals[i]));
+  }
+  usleep(72000);  // if we didn't exit this would result in next drop
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[5]);  // should now be marked as above target
+
+  usleep(205000);  // wait above two windows of time to make next entry drop
+  // Should be above target for window.
+  // should drop vals[6], enter drop state, and return 8th instead.
+  ASSERT_FALSE(c.Pop(output));
+  ASSERT_EQ(output, vals[7]);
+
+  ASSERT_FALSE(c.Push(vals[8]));
+  ASSERT_FALSE(c.Push(vals[9]));
+  ASSERT_FALSE(c.Pop(output));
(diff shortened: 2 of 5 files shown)


core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+  uint8_t drop;  // used during dequeue to determin if the packet can be dropped
+};
+
+/*
+  This algorithm gets a fast approximate inverse sqrt
+  it is taken from https://cs.uwaterloo.ca/~m32rober/rsqrt.pdf
+
+  Takes the value to get the inverse square root for and
+  returns the inverse square root of d.
+*/
```
barath: You can make these comments shorter.  For example:

`// Returns 1/sqrt(d).  Uses approximation from https://cs.uwaterloo.ca/~m32rober/rsqrt.pdf`

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+static inline double InvSqrt(double d) {
+  double d2 = d * 0.5;
+  uint64_t magic;
+  memcpy(&magic, &d, sizeof(magic));
```
barath: memcpy() shouldn't be needed -- just cast like the code in the the thesis.

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+  Wrapper* w;
+  int err = __RingDequeue(c, &w);
+  if (err != 0) {
+    return err;
+  }
+  if (w == nullptr) {
+    c->dropping = 0;
+  }
+
+  // if in dropping state, drop pointers until next drop time is greater than
```
barath: Put this comment immediately inside the if body.

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+  // the current time.
+  double now = get_epoch_time();
+  if (c->dropping) {
+    err = __DropDequeue(c, &w);
+    if (err != 0) {
+      return err;
+    }
+
+    // if not in dropping state, determine whether to enter drop state and if
```
barath: Move this comment inside the else if body.

core/utils/codel_test.cc
```diff
@@ -0,0 +1,302 @@
...
+
+#include <gtest/gtest.h>
+#include <math.h>
+#include <time.h>
+
+#include "../mem_alloc.h"
+#include "../packet.h"
+
+// helper function that will sleep for specified number of milliseconds.
+void millisleep(uint32_t milliseconds) {
```
barath: I would eliminate millisleep and just call usleep() in your code.

core/utils/codel_test.cc
```diff
@@ -0,0 +1,302 @@
...
+  ASSERT_EQ(CodelDequeue(c, reinterpret_cast<void **>(&output)), 0);
+  EXPECT_EQ(CodelCount(c), 0);
+  EXPECT_EQ(output, val);
+  mem_free(val);
+  CodelFree(c);
+}
+
+// Test the InvSqrTest to ensure produces the right value and it is accurate
+TEST(CodelTest, InvSqrtTest) {
+  EXPECT_TRUE(fabs(InvSqrt(1) - 1.0) <= .01);
```
barath: Use EXPECT_LE

core/utils/codel.h
TheJStone: It causes type punning warning that is treated like an error in compilation. so, this seemed like the next best solution.

barath: You might create a union with the double and uint64_t, and then use that to do the type conversion.

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+/*
+  Codel(Controlled Delay Management) is an Queue controller based on this
+  article http://queue.acm.org/detail.cfm?id=2209336
+
+  It provides an active queue management to help prevent bufferbloat by dropping
+  queue entries at an increasing rate if the delay in the queue is above the
+  target queue delay. The equation used to calculate drop intervals is based on TCP
+  throughput response to drop probability.
+*/
+struct Codel {
```
apanda: I am a bit confused -- why is this in C style. A lot of the functions accept `codel*` as input which seems like reasonable to include in a class function.

@barath said the reason was for similarity to `llring`, I find this form of doing this to be a bit strange, and also worry it is hard to maintain. I would rather one just made a very shallow wrapper around `llring` as a class, in which case the compiler will eliminate most if not all of the indirection costs.

Mostly my concern is that the current style looks a bit strange -- `llring` is in C style due to compatibility with the kernel module, I am not sure we should add code in this style when not necessary.

I would prefer rewriting this as a class in line with the standard C++ style used elsewhere in Bess.

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+*/
+static inline double InvSqrt(double d) {
+  double d2 = d * 0.5;
+  uint64_t magic;
+  memcpy(&magic, &d, sizeof(magic));
+  magic = 0x5fe6eb50c7b537a9 - (magic >> 1);
+  double d3;
+  memcpy(&d3, &magic, sizeof(magic));
+  d3 = d3 * (1.5 - d2 * d3 * d3);
+  return d3;
```
apanda: Please do not use this method -- it has been obsolete since at least the early 2000s. Just use pow(x, -2.0) or one of the gcc builtins. This method is not a good idea.

core/utils/codel.h
barath: Josh, Panda and I discussed your reason for structuring it as close to llring as possible, and given that he also has concerns about it, I think the best bet is to write a parent class that is a generic queue wrapper class and have two child classes, one that is for llring and one that is for codel.  Melvin already wrote the llring wrapper, which you can find in the latest few commits here:
https://github.com/melvinw/bess/commits/work
So you can go take that, change it to inherit from a common parent class, and change codel to match the interface.

sangjinhan:
A few minor comments:
* Please consider put them in `bess:utils` namespace
* Do not use `mem_alloc()` functions. New code should use C++-style allocators.
* It seems `codel.h` does not need to include `packet.h`, since `FreePacketWrap()` shouldn't be part of `codel.h`
  * In general, all files in `core/utils/` should be useful even outside the context of BESS, thus should not include any header files in `core/`. While `core/utils` has some files violating this rule at the moment, but they will be fixed at some point.

core/utils/codel.h
```diff
@@ -0,0 +1,301 @@
...
+  double window;        // minimum time before changing state
+  // the time at which codel will change state to above target(0 if below)
+  double time_above_target;
+  double time_till_drop;  // the next time codel will drop
+  // the number of packets dropped while delay has been above target
+  uint32_t drop_count;
+  uint8_t dropping;  // whether in dropping state(above target for window)
+  llring* ring;      // queue
+  void (*drop_func)(void*);  // the function to call to drop a value
+};
```
nrdmao: The algorithm cited in the paper does not use any floating point variables. Generally speaking forwarding code does not use floating point for performance reasons. If the "need" for floating point is to compute the square root of the count then this should be computable with an integer version of the square root function.

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+      window_(window),
+      time_above_target_(0),
+      time_till_drop_(MicroSecondTime() + window),
+      drop_count_(0),
+      dropping_(0),
+      queue_(init_queue),
+      drop_func_(drop_func) {}
+
+template <template <typename type> class Q, typename T>
+Codel<Q, T>::~Codel() {
```
barath: virtual

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+    if (w->ptr != nullptr) {
+      drop_func_(w->ptr);
+    }
+    delete w;
+  }
+}
+
+template <template <typename type> class Q, typename T>
+uint64_t Codel<Q, T>::NextDrop(uint64_t cur_time) {
+  return cur_time + window_ * pow(drop_count_, -.5);
```
barath: I wonder if this pow() is strictly necessary, or whether cheating a bit on the algorithm can help performance -- what does the kernel implementation do?

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+  uint64_t now = MicroSecondTime();
+  if (!w->drop) {
+    dropping_ = 0;
+  } else if (now >= time_till_drop_) {
+    while (now >= time_till_drop_ && dropping_) {
+      Drop(w);
+      drop_count_++;
+      w = RingDequeue(err);
+      if (*err != 0) {
+        return nullptr;
```
barath: This may be a dumb question, but at this point, who owns w?  That is, if there was an error, does RingDequeue sometimes still return an object, and if so, who owns that memory / who is responsible for freeing it.  If RingDequeue never returns non-null upon error, then perhaps you can get rid of the err argument as suggested in an earlier comment.  (This applies below as well.)

core/utils/lock_free_queue.h
```diff
@@ -0,0 +1,112 @@
...
+   * Construct a new queue. Takes the size of backing ring buffer (must power of
+   * two and entries available will be one less than specified. default is 256),
+   * boolean where if true, queue is in single producer mode or if false, queue
+   * is in multi producer mode, boolean where if true, queue is in single consumer
+   * mode, or if false, queue is in multi consumer mode. default for both booleans
+   * is true.
+   */
+  LockFreeQueue(size_t capacity = kDefaultRingSize, bool single_producer = true,
+                bool single_consumer = true)
+      : capacity_(capacity) {
```
barath: If capacity must be a power of two, you should check that here (with CHECK) and indicate that a non-power-of-two arg will die.

core/utils/lock_free_queue.h
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+
+  /**
+   * Enqueue one object. Takes a pointer to the object to be added and integer
+   * pointer to set on error. error codes: -1 is Quota exceeded. The objects
+   * have been enqueued, but the high water mark is exceeded. -2 is not enough
+   * room in the ring to enqueue; no object is enqueued.
+   */
+  void Push(T* obj, int* err) {
```
barath: Same comment from before about the err arguments for methods in this class.

core/utils/lock_free_queue.h
```diff
@@ -0,0 +1,112 @@
...
+    return objs;
+  }
+
+  // Returns the capacity of the queue which is one more than the total available
+  // slots in the ring buffer
+  size_t Capacity() { return capacity_; }
+
+  // Returns the number of objects in the queue
+  size_t Size() { return llring_count(ring_); }
+  // Returns true if empty and false otherwise
```
barath: Add blank line.

core/utils/lock_free_queue.h
```diff
@@ -0,0 +1,112 @@
...
+  bool Empty() { return llring_empty(ring_); }
+
+  // Returns true if full and false otherise
+  bool Full() { return llring_full(ring_); }
+
+ private:
+  struct llring* ring_;
```
barath: Add comment on who owns this pointer.

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+        uint64_t target = kDefaultTarget, uint64_t window = kDefaultWindow);
+  ~Codel();
+
+  /**
+   * Enqueue one object. Takes a pointer to the object to be added and integer
+   * pointer to set on error. error codes: -1 is Quota exceeded. The objects
+   * have been enqueued, but the high water mark is exceeded. -2 is not enough
+   * room in the ring to enqueue; no object is enqueued.
+  */
+  void Push(T* ptr, int* err);
```
TheJStone: In the DRR module pr, you mentioned that I should keep with c++ style and pass error pointer to set rather than returning an error code which is c style. I can change these to c style instead if you prefer.

barath: Ah, yes -- I know this is probably seemingly conflicting advice.  But here's the general rules I try to follow:
* If the method has no return value, and either fails or succeeds, then return a bool indicating this (e.g. true upon success).
* If the method is as above but already has a return value, then try to include the error as part of the return (e.g. return nullptr to signal failure if the return value is a pointer).
* If the method has a more complex set of error conditions then use a int * argument (as you did here).

It seemed to me that you don't have that many error conditions, and I was wondering if you could just return a failure/success signal as a bool or as part of an existing return value (here and in other methods in this PR).

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+  */
+  void Push(T** ptr, size_t count, int* err);
+
+  /**
+   * Retrieves the next entry from the queue and in the process, potentially drops 
+   * pointers as well as changes between dropping state and not dropping state. 
+   * Takes integer pointer to set on error. error codes: -3 is not enough entries in
+   * the ring to dequeue, no object. Returns the pointer to the dequeued object if
+   * there is an object to dequeue. Otherwise, nullptr. Returns the pointer to the
+   * dequeued object if there is an object to dequeue. Otherwise, nullptr.
```
TheJStone: the nullptr is sufficient for the singular case. for multi pop, the error indicates less than count entries in the resulting object if there are not enough objects in the queue. however, I could make the user responsible for this and remove the error.

core/utils/codel.h
TheJStone: at this point, Dropdequeue owns w and Ringdequeue will return a dequeued object which was initialized on enqueue. the function that replaces it/drops it is responsible for freeing it which can either be dropdequeue or Pop depending.

core/utils/codel.h
TheJStone: fixed

core/utils/codel.h
TheJStone: fixed

core/utils/codel_test.cc
TheJStone: fixed

core/utils/codel.h
TheJStone: fixed

core/utils/codel.h
TheJStone: went with pow for now but might change this later to something similar to kernel implementation if this becomes a bottleneck.

core/utils/codel.h
TheJStone: fixed

core/utils/codel_test.cc
```diff
@@ -0,0 +1,288 @@
...
+  EXPECT_EQ(output, val);
+  delete val;
+}
+
+// Tests to make sure that low delay entries won't be dropped
+TEST(CodelTest, NoDropTest) {
+  auto q = new LockFreeQueue<Codel<LockFreeQueue, int>::Wrapper>(8);
+  Codel<LockFreeQueue, int> c(&integer_drop, q, 5000, 100000);
+
+  int *val1 = new int();
```
TheJStone: fixed to an array of allocated integer pointers. but values need to be explicitly freed individually as some are freed by codel.

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+class Codel : public Queue<T> {
+ public:
+  static constexpr uint64_t kDefaultTarget = 5000;  // default delay target for codel
+  static constexpr uint64_t kDefaultWindow = 100000;  // default window size for codel
+  static const int kDefaultSlots = 4096;  // default number of slots in the codel queue
+
+  // A wrapper class around the pointer to track the delay time and whether
+  // the packet can be dropped after dequeueing.
+  struct Wrapper {
+    T* ptr;              // a pointer to be enqueued
```
TheJStone: We decided to not change this as llring requires pointers and that will be the primary queue used by codel.

core/utils/codel.h
TheJStone: out of scope

core/utils/codel_test.cc
TheJStone: out of scope

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+    uint64_t timestamp;  // a timestamp of the time the pointer was enqueued
+    uint8_t drop;  // used during dequeue to determin if the packet can be dropped
+    Wrapper(T* obj, uint64_t stamp) : ptr(obj), timestamp(stamp), drop(0){ }
+  };
+
+  Codel(void (*drop_func)(T*), Q<Wrapper>* init_queue,
```
TheJStone: fixed

core/utils/codel.h
TheJStone: fixed to follow above general rules.

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+  void Push(T* ptr, int* err);
+
+  /**
+   * Enqueue multiple objects. Takes a pointer an table of object pointers, the
+   * number of objects to be added, and  pointer to set on error. error codes:
+   * -1 is Quota exceeded. some/all objects have been enqueued, but the high water
+   * mark is exceeded. -2 is not enough room in the ring to enqueue all objects;
+   * some/all objects not enqueued.
+  */
+  void Push(T** ptr, size_t count, int* err);
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+  */
+  T* Pop(int* err);
+
+  /**
+   * Retrieves the next count entries from the queue and in the process, potentially
+   * drops pointers as well as changes between dropping state and not dropping state.
+   * Takes the number of objects to be dequeued, and an integer pointer to set on
+   * error. error codes: -3 is not enough entries in the ring to dequeue, no object.
+   * Returns a pointer to a table of void * pointers (objects) that has been filled.
+   * Otherwise, nullptr.
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+  */
+  T** Pop(size_t count, int* err);
+
+  // Returns the capacity of the queue which one more than the total available
+  // slots in the ring buffer
+  size_t Capacity();
+
+  // Returns true if empty and false otherwise
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+
+  // Returns true if full and false otherise
+  bool Full();
+
+  // Returns the number of objects in the queue
+  size_t Size();
+
+ private:
+  /**
+   * Calls the drop function on the ptr and deallocates the Wrapper
```
TheJStone: yes, I did. fixed

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+   * returns the next time to drop a packet.
+   */
+  uint64_t NextDrop(uint64_t cur_time);
+
+  /**
+   * Gets the next object from the queue and determines based on current state,
+   * whether to mark in the wrapper that the object should be dropped.
+   * Takes an integer pointer to set on error.
+   */
+  Wrapper* RingDequeue(int* err);
```
TheJStone: null/non-null sufficient

core/utils/codel.h
```diff
@@ -0,0 +1,331 @@
...
+
+  /**
+   * Called while Codel is in drop state to determine whether to drop the current
+   * entries and dequeue the next entry. Will continue to drop entries until the next
+   * time till drop is greater than the current time. Takes a pointer to the wrapper
+   * pointer where the current entry is located and an integer pointer to set on 
+   * error.
+   */
+  Wrapper* DropDequeue(Wrapper* w, int* err);
```
TheJStone: it is. fixed

core/utils/codel.h
TheJStone: fixed

core/utils/codel.h
TheJStone: we decided to leave pow as there are not any readily available built-ins that work for this purpose. however, this might be changed to more closely resemble the kernel implementation if it becomes a bottleneck.

core/utils/lock_free_queue.h
TheJStone: fixed

core/utils/codel.h
TheJStone: out of scope

core/utils/lock_free_queue.h
TheJStone: fixed

core/utils/lock_free_queue.h
TheJStone: fixed

core/utils/lock_free_queue.h
TheJStone: ring_ is initialized and kept within the lock_free_queue.

sangjinhan:
I have not enough cycles to review this now. Please let me come back to this PR in a few days.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+    gettimeofday(&tv, nullptr);
+    return tv.tv_sec * 1e6 + tv.tv_usec;
+  }
+
+  uint64_t delay_target_;  // the delay that codel will adjust for
+  uint64_t window_;        // minimum time before changing state
+
+  // the time at which codel will change state to above target(0 if below)
+  uint64_t time_above_target_;
+  uint64_t time_till_drop_;  // the next time codel will drop
```
barath: Rename this member variable -- if I understand it right, it's not the time *until* drop, but rather the time *for* the next drop (i.e. absolute rather than relative).

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+
+  // the number of packets dropped while delay has been above target
+  uint32_t drop_count_;
+  uint8_t dropping_;       // whether in dropping state(above target for window)
```
barath: Does this need to be a uint8_t or is a bool sufficient?

core/utils/codel_test.cc
```diff
@@ -0,0 +1,234 @@
...
+  ASSERT_EQ(output, vals[3]);
+
+  // should now in be drop state and the next drop should be 100ms.
+  usleep(72000);
+  output = c.Pop();
+  ASSERT_EQ(output, vals[4]);  // shouldn't be a drop yet only has been 72ms.
+
+  usleep(30000);
+  output = c.Pop();
+  ASSERT_EQ(output,
```
barath: For weird line breaks like this, move the comment before the line and merge the code line.

core/utils/queue.h
```diff
@@ -0,0 +1,24 @@
...
+#define BESS_UTILS_QUEUE_H
+
+namespace bess {
+namespace utils {
+
+// Takes Template argument T that is the type to be enqueued and dequeued.
+template <typename T>
+class Queue {
+ public:
+  virtual ~Queue(){};
```
barath: The comments for methods should be here in this .h file and not in child classes that inherit and override these.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+    Wrapper* w;
+    while ((w = queue_->Pop())) {
+      Drop(w);
+    }
+    delete queue_;
+  }
+
+  // Enqueue one object. Takes a pointer to the object to be added. Returns 0 on
+  // success.
+  int Push(T* ptr) {
```
barath: Each method that overrides a virtual method in the parent class should have the override specifier.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+  // Otherwise, returns nullptr.
+  T** Pop(size_t count) {
+    if (Size() < count) {
+      return nullptr;
+    }
+
+    T** objs = new T*[count];
+    for (size_t i = 0; i < count; i++) {
+      T* next_obj = Pop();
+      if (!next_obj) {
```
TheJStone: ah, this was a hold over code before I added the check at the beginning of the function to make it atomic. it will just return a nullptr if there are not enough items to dequeue and will never touch this code. I deleted it.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+      delete w;
+    }
+    return err;
+  }
+
+  // Enqueue multiple objects. Takes a pointer an table of object pointers, the
+  // number of objects to be added. Returns 0 on success.
+  int Push(T** ptr, size_t count) {
+    if (Capacity() - Size() <= count) {
+      return -2;
```
TheJStone: llring has this interesting error scheme where -1 is when the enqueue is successful but the queue is now full and -2 is when the enqueue is unsuccessful because there is no room. Should I extend this support into Codel? if not, should I change this in the llring wrapper as well?

core/utils/codel.h
TheJStone: bool is sufficient

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+      Drop(w);
+    } else {
+      ptr = nullptr;
+    }
+    return ptr;
+  }
+
+  // Retrieves the next count entries from the queue and in the process, potentially
+  // drops pointers as well as changes between dropping state and not dropping state.
+  // Takes the number of objects to be dequeued. Returns a pointer to a table of
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+        if (now - time_till_drop_ < window_ && drop_count_ > 2) {
+          drop_count_ -= 2;
+        } else {
+          drop_count_ = 1;
+        }
+        time_till_drop_ = NextDrop(now);
+      }
+    }
+
+    T* ptr;
```
TheJStone: I changed this to make it clearer and more streamlined. fixed.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+      return nullptr;
+    }
+
+    uint64_t now = MicroSecondTime();
+    if (dropping_) {
+      // if in dropping state, drop pointers until next drop time is greater
+      // than
+      // the current time.
+      w = DropDequeue(w);
+
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+    Wrapper* w = RingDequeue();
+    if (!w) {
+      dropping_ = 0;
+      return nullptr;
+    }
+
+    uint64_t now = MicroSecondTime();
+    if (dropping_) {
+      // if in dropping state, drop pointers until next drop time is greater
+      // than
```
TheJStone: fixed

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+  int Push(T* ptr) {
+    Wrapper* w = new Wrapper(ptr, MicroSecondTime());
+    int err = queue_->Push(w);
+    if (err) {
+      delete w;
+    }
+    return err;
+  }
+
+  // Enqueue multiple objects. Takes a pointer an table of object pointers, the
```
TheJStone: yes. fixed

core/utils/codel.h
TheJStone: fixed

core/utils/codel_test.cc
TheJStone: fixed.

core/utils/codel.h
TheJStone: fixed. now `next_drop_time_`

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+  // before changing into drop state.
+  Codel(void (*drop_func)(T*), Queue<Wrapper>* init_queue,
+        uint64_t target = kDefaultTarget, uint64_t window = kDefaultWindow)
+      : delay_target_(target),
+        window_(window),
+        time_above_target_(0),
+        time_till_drop_(MicroSecondTime() + window),
+        drop_count_(0),
+        dropping_(0),
+        queue_(init_queue),
```
TheJStone: added.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+// queue entries at an increasing rate if the delay in the queue is above the
+// target queue delay. The equation used to calculate drop intervals is based on TCP
+// throughput response to drop probability.
+
+// Takes Template argument Q which is a templated child class of Queue and T the type
+// that is going to be enqueued/dequeued.
+template <typename T>
+class Codel : public Queue<T> {
+ public:
+  static constexpr uint64_t kDefaultTarget =
```
TheJStone: they were holdover from something that I was trying to do but failed. fixed.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
...
+      100000;  // default window size for codel
+  static const int kDefaultSlots =
+      4096;  // default number of slots in the codel queue
+
+  // A wrapper class around the pointer to track the delay time and whether
+  // the packet can be dropped after dequeueing.
+  struct Wrapper {
+    T* ptr;              // a pointer to be enqueued
+    uint64_t timestamp;  // a timestamp of the time the pointer was enqueued
+    uint8_t
```
TheJStone: bool is sufficient

core/utils/codel.h
TheJStone: Should I just have a buffer to store in in this case so that I can return nullptr and keep it atomic?

barath: Might be better to just return the number of elements returned as a second argument, and change the method's comment / guarantee.

TheJStone: then do the same thing in push? and then automatically resolves the question above for the llring wrapper.

barath: Let's invert the arguments actually -- return the number of elements actually popped, and take two arguments -- count (the requested number of elements to pop) and a pre-allocated T array of size count that is allocated and owned by the caller.  (This would then eliminate the new.)

core/utils/queue.h
```diff
@@ -0,0 +1,24 @@
...
+  virtual int Push(T*) = 0;
```
sangjinhan: I am a little bit concerned that the convention used here is quite different from general template usages.

For example, if someone wants to store "pointers" to objects, the template type T would be different between STL and bess::utils::Queue, which is quite confusing.
```
class Foo { ...} ;
Foo obj;

std::deque<Foo *> a;
a.push(&obj);

bess::utils::Queue<Foo> b;
b.push(&obj);
```

I understand that `llring` only takes pointer elements, but that is limitation of LockFreeQueue, not Queue. You can enforce only pointers to be used in LockFreeQueue, something like this (not tested):
```
#include <type_traits>

template<typename T>
class LockFreeQueue : Queue {
  static_assert(std::is_pointer<T>::value, "LockFreeQueue only support pointer types");
  ...
};
```

TheJStone: In converting the Codel to be pointer independent, I ran into an issue for the Pop method where I can no longer return null when there are no objects(this breaks my current implementation). I have a few alternative implementation options: 1. return an object and take an error parameter to set, 2. return error parameter and take an object by reference to set, 3. throw an error when there are no objects left. none seem very clean although the last is more conventional for queues. I was really unsure on what to go with. thoughts?

core/utils/codel_test.cc
```diff
@@ -0,0 +1,234 @@
+#include "./codel.h"
+
+#include <gtest/gtest.h>
+#include <math.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "../mem_alloc.h"
+#include "../packet.h"
```
TheJStone: not anymore. I removed them.

core/utils/codel_test.cc
```diff
@@ -0,0 +1,234 @@
+#include "./codel.h"
```
TheJStone: sure. fixed for the others as well.

core/utils/lock_free_queue.h
```diff
@@ -0,0 +1,101 @@
...
+#include "./common.h"
+#include "./queue.h"
+
+namespace bess {
+namespace utils {
+
+// A wrapper class for llring that extends the abstract class Queue. Takes a
+// template argument T which is the type to be enqueued and dequeued.
+template <typename T>
+class LockFreeQueue : public Queue<T> {
```
TheJStone: ah. fixed both and made codel final as well.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
+#ifndef BESS_UTILS_CODEL_H_
+#define BESS_UTILS_CODEL_H_
+
+#include <math.h>
+#include <sys/time.h>
+
+#include <iostream>
```
TheJStone: yes, sorry. I removed it.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
+#ifndef BESS_UTILS_CODEL_H_
+#define BESS_UTILS_CODEL_H_
+
+#include <math.h>
+#include <sys/time.h>
```
TheJStone: easy enough. converted everything in codel to be in nanoseconds.

core/utils/codel.h
```diff
@@ -0,0 +1,267 @@
+#ifndef BESS_UTILS_CODEL_H_
+#define BESS_UTILS_CODEL_H_
+
+#include <math.h>
```
TheJStone: fixed.

core/utils/codel_test.cc
```diff
@@ -0,0 +1,234 @@
...
+
+#include <gtest/gtest.h>
+#include <math.h>
+#include <time.h>
```
TheJStone: ah, sorry. fixed.

core/utils/queue.h
sangjinhan: 1. Seems infeasible, since you cannot create a dummy object to return if T is not [default constructible](http://en.cppreference.com/w/cpp/concept/DefaultConstructible).
2. sounds good. The return type could be either bool or int (0 or -errno).
3. Like it or not, the Google C++ Style Guide does not allow using exceptions.

TheJStone:
Updated with all suggested changes.

sangjinhan:
Merged, thanks!
tumblr_m8gz0alsmt1rcduwzo1_1280

PR #354, opened 2017-03-24
Execute `cd ..` even if `all_test` fails

### LINKED ISSUE #347: Error in the travis script
https://travis-ci.org/NetSys/bess/jobs/213513861#L483

When the command `$ cd core && ./all_test && cd ..` fails, the last `cd ..` is not executed due to the `&&` condition. As a result, all subsequent commands will automatically fail since the working directory is incorrect.

changlan:
This should fix #347.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_script:
 script:
   - ./container_build.py bess
   - ./container_build.py kmod_buildtest
-  - cd core && ./all_test && cd ..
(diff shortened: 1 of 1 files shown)


sangjinhan:
Perhaps using a subshell is cleaner...?

```
...
(cd core && ./all_test)
...
```

than

```
...
cd core && ./all_test
cd ..
...
```

changlan:
Nice. Fixed. #TIL

PR #356, opened 2017-03-28
Call rte_eth_tx_queue_setup first

sangjinhan:
For some mysterious bug in the DPDK virtio code, the PMD driver running in a VM crashes in `rte_eth_rx_burst()`, if a virtio device was earlier initialized with `rte_eth_rx_queue_setup()` called first. This PR reverses the order, invoking `rte_eth_tx_queue_setup()` first then `rte_eth_rx_queue_setup()`.

!@*&^($*&^

### DIFF
--- a/core/drivers/pmd.cc
+++ b/core/drivers/pmd.cc
@@ -235,6 +235,8 @@ pb_error_t PMDPort::Init(const bess::pb::PMDPortArg &arg) {
   }
   rte_eth_promiscuous_enable(ret_port_id);
 
+  // NOTE: As of DPDK 17.02, TX queues should be initialized first.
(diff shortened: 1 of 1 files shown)


core/drivers/pmd.cc
```diff
@@ -235,6 +235,16 @@ pb_error_t PMDPort::Init(const bess::pb::PMDPortArg &arg) {
   }
   rte_eth_promiscuous_enable(ret_port_id);
 
+  for (i = 0; i < num_txq; i++) {
```
barath: Maybe add a comment the the ordering of the call is to avoid a DPDK 17.02 bug?

sangjinhan: Sure thing!

sangjinhan:
🐈

PR #357, opened 2017-03-29
Merge LeafTrafficClass and Task

ddiproietto:
With this pull request LeafTrafficClass can only execute a single Task (more details in the last commit message).

To execute multiple Tasks we now have to use a RoundRobinClass with multiple leaves.  This can be slower, because the scheduler is called after every Task is executed.  Is it important to optimize for these use cases?


sangjinhan:
Let me review the code by today.

ddiproietto:
There is a bug here:  removing a module causes a crash.  It happens because the module needs to remove the leaves.

I'm closing this and opening a new pull request

Thanks!

PR #360, opened 2017-03-30
Add support for protobuf Map types

sangjinhan:
The existing Python API crashes upon use of the protobuf Map type, e.g., `MetadataTest` module has:
```
message MetadataTestArg {
  map<string, int64> read = 1;
  map<string, int64> write = 2;
  map<string, int64> update = 3;
}
```

This PR addresses this crash issue, along with fixes for outdated unit tests.


libbess-python/bess.py
```diff
@@ -10,7 +10,7 @@
 import time
 
 import service_pb2
-import proto_conv
+from protobuf_to_dict import *
```
changlan: It is generally recommended not to use wildcard import. How about `from protobuf_to_dict import protobuf_to_dict, dict_to_protobuf`?

libbess-python/protobuf_to_dict_test.py
```diff
@@ -1,8 +1,7 @@
 import unittest
 import bess_msg_pb2 as bess_msg
 import error_pb2 as error_msg
-import proto_conv
-
+from protobuf_to_dict import *
```
changlan: Same here.

libbess-python/bess.py
sangjinhan: yes my lord. coming soon...

changlan:
Merged, thanks!
image

PR #361, opened 2017-03-31
Specify localhost for ansible-playbook

changlan:


sangjinhan:
A few comments:
* You can completely ignore this: `-K` instead of `--ask-become-pass` for brevity?
* The pip task failed on a fresh Ubuntu 16.04 and I suspect it is related to [this](https://github.com/ansible/ansible/issues/13624). I changed the task like this and it worked (note that I removed protobuf, which is a dependency of grpcio anyway).
```
- name: Install required Python packages
  pip: name=grpcio
  become: true
```
* Why do we install Ansible from ppa, not from Ubuntu default repo?

sangjinhan:
Maybe `-t package` option should be there, since we don't want the "vagrant" section to be run?

changlan:
Yep, fixes are coming. 
Re: ppa, some versions of Ubuntu may have Ansible 1.x, the behavior of which is quite different.

PR #362, opened 2017-03-31
Embed leaf task v2

ddiproietto:
With this pull request LeafTrafficClass can only execute a single Task (more details in the last commit message).

To execute multiple Tasks we now have to use a RoundRobinClass with multiple leaves. This can be slower, because the scheduler is called after every Task is executed. Is it important to optimize for these use cases?


sangjinhan:
Code looks great, I don't have anything to comment on. 👍 

Perhaps we can revise the user interface and the default behavior of traffic classes to make it more intuitive. For example, with the `samples/queue` script:

```
...
src::Source() \
        -> Rewrite(templates=[test_packet]) \
        -> queue::Queue() \
        -> VLANPush(tci=2) \
        -> Sink()

bess.add_tc('fast', policy='rate_limit', resource='packet', limit={'packet': 9000000})
bess.add_tc('fast_leaf', policy='leaf', parent='fast', leaf_module_name=src.name)

bess.add_tc('slow', policy='rate_limit', resource='packet', limit={'packet': 1000000})
bess.add_tc('slow_leaf', policy='leaf', parent='slow', leaf_module_name=queue.name)
...

$ show tc
  worker 0 (6 classes)
    root_0            parent none        share 0   unlimited
    defaultrr_0       parent root_0      share 0   unlimited
    slow              parent defaultrr_0  share 0   limits: 1.000 Mpps
    slow_leaf         parent slow        share 0   unlimited
    fast              parent defaultrr_0  share 0   limits: 9.000 Mpps
    fast_leaf         parent fast        share 0   unlimited
```

A few thoughts:
* While the "important" TCs are only `slow` and `fast`, there are too many other TCs accompanied.
  * The leaf TCs must be made explicitly, almost always. This is pretty tedious...
* Maybe we don't need two "default" TCs: `root_0` and `defaultrr_0`.
  * As far as the sample scripts are concerned, none of them utilize `root_X`. 
  * Having a PriorityScheduler as the default root is a bit awkward.
  * RoundRobinScheduler for root would make more sense, since:
     * It has the lowest overheads,
     * It allows an arbitrary number of child TCs, and
     * It doesn't require any additional parameters to add children.
* If the user specifies a complete tree of TCs, neither root_0 nor defaultrr_0 would be useful.
* Usability issues:
   * Chasing the "parent" chains for `show tc` is not intuitive, when the user checks the TC hierarchy.
   * `share`, `unlimited`, or `priority` should be shown only when they are relevant.
   * It's hard to know what modules are associated with what TCs
* People seem to be confused with the concept of task and task ID...

Therefore, I'd like to suggest (@ddiproietto and @barath, please check if they are reasonable:
* Leaf TCs:
  * A default leaf TC is automatically created upon module creation
  * Its (internal) TC name is `!<module name>:<task name>`.
    * The `:<task name>` is optional, only for modules that create multiple tasks
      * only `PortInc` at the moment I think.
* `add_tc` is used to create non-leaf TCs. The parent of a TC may or may not be specified.
  * When no parent is specified, a `wid` may or may not be specified.
  * `update_tc` can be used to specify a parent (or change a parent)
* By default, BESS does not create any default TCs for workers. Instead, on `resume_all`, BESS performs the following actions as necessary:
  * If there are orphan TCs that are not bound to any worker, each is allocated to a worker in an arbitrary manner (e.g., round robin). After this, each worker has a forest of trees (each orphan is a root).
    * If there is only one orphan TC and it is non-leaf: the TC becomes the root of the worker.
    * Otherwise: a default RoundRobinTC is created as the root of the worker, and all leaf/non-leaf workers become its children.
  * At last, if there is an auto-created default root TC that has no child any longer, the TC is automatically removed.

To summarize, it would be like:
```
# API functions:
# add_tc(tc_name, policy, parent=None, wid=None, ...)    # NOTE: parent and wid cannot be specified both
# update_tc(tc_name, parent=None, wid=None, ...)   # you can do everything as add_tc, except for changing its kind (policy)
# attach_module(tc_name, mod_name, task_name=None, parent=None, wid=None)
# delete_tc(tc_name)

# new samples/queue.bess
...
bess.add_tc('fast', policy='rate_limit', resource='packet', limit=9000000)
bess.attach_module('fast', src.name)    # TODO: src.attach_to('fast') should also work

bess.add_tc('slow', policy='rate_limit', resource='packet', limit=1000000)
bess.attach_module('slow', queue.name)
...

# (with some cosmetic changes...)
$ show tc
  <worker 0>            round_robin
   +-- fast             rate_limit
   |     +-- (src)                      limits: 9.000 Mpps
   +-- slow             rate_limit
         +-- (queue)                    limits: 1.000 Mpps
```

sangjinhan:
It seems that the `Traverse` issue is still there. You can reproduce the bug by running the `samples/queue` script repeatedly.

```
<disconnected> $ daemon start
Done.
localhost:10514 $ run samples/queue
Done.
localhost:10514 $ run samples/queue
WARNING: The current pipeline will be reset. Are you sure? (type "yes") yes
Done.
localhost:10514 $ run samples/queue
*** Error: RPC failed to localhost:10514 - <_Rendezvous of RPC that terminated with (StatusCode.UNAVAILABLE, Endpoint read failed)>
From /tmp/bessd_crash.log (Mon Apr  3 22:51:55 2017):
A critical error has occured. Aborting...
Signal: 11 (Segmentation fault), si_code: 1 (SEGV_MAPERR: address not mapped to object)
pid: 13075, tid: 13077, address: 0x10, IP: 0x69d430
Backtrace (recent calls first) ---
(0): /home/sangjin/bess/core/bessd(_ZNK4bess22RoundRobinTrafficClass8TraverseEPFvPKNS_12TrafficClassEPvES4_+0x50) [0x69d430]
    bess::RoundRobinTrafficClass::Traverse(void (*)(bess::TrafficClass const*, void*), void*) const at /home/sangjin/bess/core/traffic_class.cc:385
         382:     child->Traverse(f, arg);
         383:   }
         384:   for (const auto &child : blocked_children_) {
      -> 385:     child->Traverse(f, arg);
         386:   }
         387: }
         388:
(1): /home/sangjin/bess/core/bessd(_ZNK4bess20PriorityTrafficClass8TraverseEPFvPKNS_12TrafficClassEPvES4_+0x43) [0x69d323]
    bess::PriorityTrafficClass::Traverse(void (*)(bess::TrafficClass const*, void*), void*) const at /home/sangjin/bess/core/traffic_class.cc:125
      -> 125:     child.c_->Traverse(f, arg);
(2): /home/sangjin/bess/core/bessd(_ZNK4bess12TrafficClass4SizeEv+0x29) [0x69dfe9]
    bess::TrafficClass::Size() const at /home/sangjin/bess/core/traffic_class.cc:22
      -> 22:       static_cast<void*>(&ret));
(3): /home/sangjin/bess/core/bessd(_ZN15BESSControlImpl11ListWorkersEPN4grpc13ServerContextEPKN4bess2pb12EmptyRequestEPNS4_19ListWorkersResponseE+0xcd) [0x629b2d]
    bess::Scheduler<Task>::NumTcs() const at /home/sangjin/bess/core/scheduler.h:198
      -> 198:     return root_->Size() - 1;
     (inlined by) BESSControlImpl::ListWorkers(grpc::ServerContext*, bess::pb::EmptyRequest const*, bess::pb::ListWorkersResponse*) at /home/sangjin/bess/core/bessctl.cc:401
      -> 401:       status->set_num_tcs(workers[wid]->scheduler()->NumTcs());
```

ddiproietto:
Thanks a lot for your comments!  The user interface you're suggesting seems a lot simpler, thanks.  I'll implement the changes and post a new version.

The Traverse issue can be reproduced also on master by running `samples/queue` and doing `show tc` repeatedly.  As discussed offline, maybe I can fix it by keeping static copies of the `*children_` structures that are not changed by the workers.

I'll post a new version as soon as I'm done, thanks for the suggestions!

barath:
Sangjin, I think your design / suggestions are good.

I had only one comment: "If there are orphan TCs that are not bound to any worker, each is allocated to a worker in an arbitrary manner (e.g., round robin). After this, each worker has a forest of trees (each orphan is a root)." -- I think the ultimate intent of this is simpler than this phrasing suggests.  If I understand right, all that happens is: if there is only one TC in the worker, it becomes the root; if there's more than one, they get attached to a round robin root.

ddiproietto:
I've implemented the interface as suggested above except:

* I've split the UpdateTc message into two: UpdateTcParam and UpdateTcParent. UpdateTcParam is used to change the tc's own parameters (like the limit), UpdateTcParent is used to move TCs.

* Since the scheduler keeps pointers to TCs in throttled_cache, UpdateTcParent is only implemented for leaf or orphan non-leaf.  For the same reason, I haven't implemented DeleteTc.

I've opened another pull request

Thanks

PR #363, opened 2017-04-02
Fix sigsegv in cuckoo_map.h

apanda:
Calling `*this = std::move(<something else>)` is unsafe outside of
the move constructor in C++11 (in particular *this is const). In this
case this resulted in a segfault. I changed this to a series of moves
of the inner elements. Interestingly in searching through [github](https://github.com/search?q="*this+%3D+std%3A%3Amove"+extension%3Acpp+extension%3Acxx&type=Code)
I have not found anyone use this outside of a move constructor, so I
wonder why we chose this construct.

### DIFF
--- a/core/utils/cuckoo_map.h
+++ b/core/utils/cuckoo_map.h
@@ -479,7 +479,11 @@ class CuckooMap {
       }
     }
 
-    *this = std::move(bigger);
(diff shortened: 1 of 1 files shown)


barath:
img_0786

PR #364, opened 2017-04-03
Implement Replicate module

justinemarie:
One of the projects I'm working on requires that I handle broadcast packets, which means that I need the ability to replicate these packets to multiple recipients when I see such a packet. This module creates n copies of a single packet and sends them out on n output gates.

### DIFF
--- a/core/modules/replicate.cc
+++ b/core/modules/replicate.cc
@@ -1,8 +1,8 @@
 #include "replicate.h"
 
 const Commands Replicate::cmds = {
-  {"set_gates", "ReplicateCommandSetGatesArg",
-    MODULE_CMD_FUNC(&Replicate::CommandSetGates), 0},
+    {"set_gates", "ReplicateCommandSetGatesArg",
+     MODULE_CMD_FUNC(&Replicate::CommandSetGates), 0},
 };
 
 pb_error_t Replicate::Init(const bess::pb::ReplicateArg &arg) {
@@ -42,24 +42,25 @@ pb_cmd_response_t Replicate::CommandSetGates(
 
 void Replicate::ProcessBatch(bess::PacketBatch *batch) {
   bess::PacketBatch out_gates[ngates_];
-  for(int i = 0; i < ngates_; i++){
+  for (int i = 0; i < ngates_; i++) {
     out_gates[i].clear();
   }
 
   for (int i = 0; i < batch->cnt(); i++) {
-    bess::Packet* tocopy = batch->pkts()[i];
+    bess::Packet *tocopy = batch->pkts()[i];
     out_gates[0].add(tocopy);
-    for(int j = 1; j < ngates_; j++){
-      bess::Packet* newpkt = bess::Packet::copy(tocopy);
-      if(newpkt){
+    for (int j = 1; j < ngates_; j++) {
+      bess::Packet *newpkt = bess::Packet::copy(tocopy);
+      if (newpkt) {
         out_gates[j].add(newpkt);
       }
     }
   }
 
-  for(int j = 0; j < ngates_; j++){
+  for (int j = 0; j < ngates_; j++) {
     RunChooseModule(gates_[j], &(out_gates[j]));
   }
 }
 
-ADD_MODULE(Replicate, "repl", "makes a copy of a packet and sends it out over n gates")
+ADD_MODULE(Replicate, "repl",
+           "makes a copy of a packet and sends it out over n gates")
--- a/core/modules/replicate.cc
+++ b/core/modules/replicate.cc
@@ -8,8 +8,8 @@ const Commands Replicate::cmds = {
 pb_error_t Replicate::Init(const bess::pb::ReplicateArg &arg) {
   pb_error_t err;
 
-  if (arg.gates_size() > MAX_REPL_GATES) {
-    return pb_error(EINVAL, "no more than %d gates", MAX_REPL_GATES);
+  if (arg.gates_size() > kMaxGates) {
+    return pb_error(EINVAL, "no more than %d gates", kMaxGates);
   }
 
   for (int i = 0; i < arg.gates_size(); i++) {
@@ -25,15 +25,14 @@ pb_cmd_response_t Replicate::CommandSetGates(
     const bess::pb::ReplicateCommandSetGatesArg &arg) {
   pb_cmd_response_t response;
 
-  if (arg.gates_size() > MAX_REPL_GATES) {
+  if (arg.gates_size() > kMaxGates) {
     set_cmd_response_error(
-        &response, pb_error(EINVAL, "no more than %d gates", MAX_REPL_GATES));
+        &response, pb_error(EINVAL, "no more than %d gates", kMaxGates));
     return response;
   }
 
   for (int i = 0; i < arg.gates_size(); i++) {
-    int elem = arg.gates(i);
-    gates_[i] = elem;
+    gates_[i] = arg.gates(i);
   }
 
   ngates_ = arg.gates_size();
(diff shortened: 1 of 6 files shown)


protobuf/module_msg.proto
```diff
@@ -720,13 +717,35 @@ message RoundRobinCommandSetModeArg {
...
+message ReplicateArg {
+  repeated int64 gates = 1; /// A list of gate numbers to send packet copies to.
+}
+
+/**
+ * The Replicate module has a function `setGates(...)` which changes
+ * the total number of output gates in the module.
+ */
+message ReplicateCommandSetGatesArg {
+  repeated int64 gates = 1; /// A list of gate numbers to round-robin the traffic over.
```
sangjinhan: It's not round robin :)

core/packet.h
```diff
@@ -181,7 +181,7 @@ class alignas(64) Packet {
     DCHECK_EQ(ret, 0);
   }
 
-  Packet *copy(Packet *src) {
+  static Packet *copy(Packet *src) {
```
sangjinhan: The function should check if the result of `__packet_alloc_pool()` is `nullptr`, and if so, should just return `nullptr` without doing anything.

sangjinhan: Also consider make the argument `const Packet *src`

core/modules/replicate.h
```diff
@@ -0,0 +1,36 @@
+#ifndef BESS_MODULES_REPLICATE_H_
+#define BESS_MODULES_REPLICATE_H_
+
+#include "../module.h"
+#include "../module_msg.pb.h"
+
+#define MAX_REPL_GATES 20
```
sangjinhan: Instead of using a macro, how about `static const gate_idx_t kMaxGates` as a public member of `class Replicate`?

core/modules/replicate.cc
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  ngates_ = arg.gates_size();
+  set_cmd_response_error(&response, pb_errno(0));
+  return response;
+}
+
+void Replicate::ProcessBatch(bess::PacketBatch *batch) {
+  bess::PacketBatch out_gates[ngates_];
+  memset(out_gates, 0, sizeof(bess::PacketBatch) * ngates_);
```
sangjinhan: You don't need to zero-initialize the whole arrays. `for(...) { out_gates[i].clear(); }` would be much faster, since it only clears the `cnt_` member.

core/modules/replicate.cc
```diff
@@ -0,0 +1,60 @@
...
+
+  for (int i = 0; i < batch->cnt(); i++) {
+    out_gates[0].add(batch->pkts()[i]);
+    for(int j = 1; j < ngates_; j++){
+      out_gates[j].add(bess::Packet::copy(batch->pkts()[i]));
```
sangjinhan: If `bess::Packet::copy()` returns `nullptr` the packet should be skipped.

core/modules/replicate.cc
```diff
@@ -0,0 +1,60 @@
...
+  set_cmd_response_error(&response, pb_errno(0));
+  return response;
+}
+
+void Replicate::ProcessBatch(bess::PacketBatch *batch) {
+  bess::PacketBatch out_gates[ngates_];
+  memset(out_gates, 0, sizeof(bess::PacketBatch) * ngates_);
+
+  for (int i = 0; i < batch->cnt(); i++) {
+    out_gates[0].add(batch->pkts()[i]);
```
sangjinhan: How about
```
Packet *org = batch->pkts()[i];
out_gates[0].add(org);
```
and reuse `org` in the loop below? I am afraid `batch->pkts()[i]` might be read from memory over and over again rather than reusing a register, since the compiler may incorrectly assume that `PacketBatch::add` or `bess::Packet::copy()` might have overwritten `batch->pkts()[i]`.

core/modules/replicate.cc
```diff
@@ -0,0 +1,60 @@
...
+  pb_cmd_response_t response;
+
+  if (arg.gates_size() > MAX_REPL_GATES) {
+    set_cmd_response_error(
+        &response, pb_error(EINVAL, "no more than %d gates", MAX_REPL_GATES));
+    return response;
+  }
+
+  for (int i = 0; i < arg.gates_size(); i++) {
+    int elem = arg.gates(i);
```
sangjinhan: Why not directly assign but use a temporary variable here?

sangjinhan:
The Travis error seems benign.

justinemarie:
Restarted the Travis job.

protobuf/module_msg.proto
justinemarie: Fixed.

core/packet.h
justinemarie: Check.

core/modules/replicate.h
justinemarie: Fixed

core/modules/replicate.cc
justinemarie: Fixed

core/modules/replicate.cc
justinemarie: Ok!

core/modules/replicate.cc
justinemarie: ack

core/packet.h
justinemarie: When I made it const something down the line started crashing...

core/modules/replicate.h
```diff
@@ -0,0 +1,36 @@
...
+#define BESS_MODULES_REPLICATE_H_
+
+#include "../module.h"
+#include "../module_msg.pb.h"
+
+
+class Replicate final : public Module {
+ public:
+  static const gate_idx_t kMaxGates = 20;
+  static const gate_idx_t kNumOGates = kMaxGates;
```
sangjinhan: Is it used anywhere..?

core/modules/replicate.cc
```diff
@@ -0,0 +1,65 @@
...
+void Replicate::ProcessBatch(bess::PacketBatch *batch) {
+  bess::PacketBatch out_gates[ngates_];
+  for(int i = 0; i < ngates_; i++){
+    out_gates[i].clear();
+  }
+
+  for (int i = 0; i < batch->cnt(); i++) {
+    bess::Packet* tocopy = batch->pkts()[i];
+    out_gates[0].add(tocopy);
+    for(int j = 1; j < ngates_; j++){
```
sangjinhan: Spaces around `(...)` :)

core/modules/replicate.cc
```diff
@@ -0,0 +1,65 @@
...
+      bess::Packet* newpkt = bess::Packet::copy(tocopy);
+      if(newpkt){
```
sangjinhan: ditto

core/packet.h
sangjinhan: Adding `const` to `head_data` will stop the compiler complaining, like this: 
```
T head_data(uint16_t offset = 0) const {
```

nrdmao:
Have you considered an option or version of this module that will do reference counting rather than packet copies? Particularly for packet data? If the use case is for broadcast, then packet copies seem like overkill.

sangjinhan:
I guess what @justinemarie presumed here is that cloning packets is an infrequent operation, used mostly in slow path. In this case the slower-but-simpler solution would be fine. Since the `Packet` class doesn't provide any abstractions like copy-on-write at the moment, reference counting or shallow packet copy without protection may cause incorrect behaviors.

One interesting thing to note is that the DPDK PMD drivers are optimized for "common" cases, running much slower with uncommon packets (e.g., refcnt >=2 ). This might nullify the performance benefit from zero copy.

justinemarie:
I actually just straight up didn't know DPDK supports refcounting -- if
there's a way to do this and we think it is better, I am happy to make the
change.

Nonetheless, since the primary use case here is broadcast traffic which is
less than common, I am less worried about it (as Sangjin says).

nrdmao:
OK, the rational for the current implementation of replicate sounds reasonable. Indeed broadcast should be rare relative to unicast traffic for normal subnet broadcasts. I guess I was just sort of probing about what people were thinking with respect to general multicast features. If one were to implement zero copy L3 multicast you would have to worry about things like modifications per multicast group member on the packet headers, for example for l2 encapsulation. This implies an ability to encapsulate using non-contiguous buffers and the drivers would have to be able to gather multiple buffers on egress. All modules may then have to be aware of these complications and deal with them.

sangjinhan:
tumblr_m4qd9w3qrt1r0wqrdo1_500

PR #366, opened 2017-04-04
Keep the ownership of FlowRecord within AvailablePorts

changlan:
AvailablePorts used to give up the ownership of the FlowRecord object after releasing a pair of IP/port. This design is not safe, as it depends on the caller to maintain the ownership correctly. When AvailablePorts got destructed, released FlowRecord objects could never be freed. Hence, a better design should keep the ownership within AvailablePorts during its lifetime.

### DIFF
--- a/core/modules/nat.h
+++ b/core/modules/nat.h
@@ -100,38 +100,44 @@ class AvailablePorts {
  public:
   // Tracks available ports within the given IP prefix.
   explicit AvailablePorts(const CIDRNetwork &prefix)
-      : prefix_(prefix), free_list_(), next_expiry_() {
-    uint32_t min = ntohl(prefix_.addr & prefix_.mask);
-    uint32_t max = ntohl(prefix_.addr | (~prefix_.mask));
+      : prefix_(prefix),
+        records_(),
+        free_list_(),
+        next_expiry_(),
+        min_(),
+        max_() {
+    min_ = ntohl(prefix_.addr & prefix_.mask);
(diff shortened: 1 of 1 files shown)


core/modules/nat.h
```diff
@@ -100,38 +100,45 @@ class AvailablePorts {
...
 
   // Returns a random free IP/port pair within the network and removes it from
   // the free list.
   std::tuple<IPAddress, uint16_t, FlowRecord *> RandomFreeIPAndPort() {
-    std::tuple<IPAddress, uint16_t, FlowRecord *> r = free_list_.back();
+    size_t i = free_list_.back();
     free_list_.pop_back();
-    return r;
+
+    uint32_t ip = i / (MAX_PORT - MIN_PORT + 1) + min_;
```
sangjinhan: 64bit integer division/modulo operations are one of the most expensive instructions in x86. For example, it takes more than 50% of CPU time of the `Measure` module. Would it be possible to avoid using division/modulo operations somehow?

changlan: I see. Would storing `std::pair<uint32_t, uint16_t>` in `free_list_` be faster?

sangjinhan: I think so, and it should use the same amount of memory. Are you going to reconstruct `i` with `ip` and `port`?

changlan: Yes.

sangjinhan: Sounds good

sangjinhan:
Merged, thanks!
4eflwgoudg20m07f0tafwbgpo1_500

PR #367, opened 2017-04-04
32 bit compile issues, Makefile envirnomental control variables

nrdmao:
- changes are required to get bess to compile with a 32 bit compiler.
- changes to the Makefile to allow compilation for an alternative architecture.
- changes to the Makefile to allow the option to link against shared libraries.

### DIFF
--- a/core/modules/exact_match.h
+++ b/core/modules/exact_match.h
@@ -64,7 +64,7 @@ class em_hash {
     }
     return init_val;
 #else
-    return rte_hash_crc(&key, key.key_len, init_val);
+    return rte_hash_crc(&key, len_, init_val);
 #endif
   }
 
--- a/core/utils/checksum.h
+++ b/core/utils/checksum.h
@@ -74,6 +74,7 @@ static inline uint32_t CalculateSum(const void *buf, size_t len) {
   }
 #endif
 
+#if __x86_64
   // Repeat 64-bit one's complement sum (at sum64) including carrys
   // 8 additions in a loop
   while (len >= sizeof(uint64_t) * 8) {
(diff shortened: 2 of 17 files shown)


core/utils/checksum.h
```diff
@@ -74,6 +74,7 @@ static inline uint32_t CalculateSum(const void *buf, size_t len) {
   }
 #endif
 
+#if __x86_64
```
sangjinhan: Since this code block doesn't have anything specific to x86, can we use something else (e.g., `__LP64__`)..? While BESS only supports i686 and x86_64, it would be good to keep the code architecture neutral as much as we can.

core/modules/exact_match.h
```diff
@@ -64,7 +64,7 @@ class em_hash {
     }
     return init_val;
 #else
-    return rte_hash_crc(&key, key.key_len, init_val);
+    return rte_hash_crc(&key, len_, init_val);
```
sangjinhan: Good point 👍

sangjinhan:
Thank you for the patches! I'll add a 32-bit build test to Travis to keep the code compatible with i686.

core/utils/checksum.h
```diff
@@ -109,9 +110,11 @@ static inline uint32_t CalculateSum(const void *buf, size_t len) {
   // Reduce 64-bit unsigned integer to 32-bit unsigned integer
   // Carry may happens, but no need to complete reduce here
   sum64 = (sum64 >> 32) + (sum64 & 0xFFFFFFFF);
+#endif
 
   // Repeat 16-bit one's complement sum (at sum64)
-  const uint16_t *buf16 = reinterpret_cast<const uint16_t *>(buf64);
+  typedef uint16_t __attribute__((__may_alias__)) u16_p;
```
sangjinhan: Why did you add this line? My test with g++ 4.8 / 5.4 does not complain about aliasing here...

nrdmao: Have you run the unit tests? It was not a compile time issue but a run time issue. It seems when there is optimization -O2 or greater the optimizer presumes strict aliasing. Note that this is the same technique used rte_raw_cksum(). I suspect when the remainder of the function is ifdef'd out (as is the case in 32 bit mode) the optimizer kicks in and the strict aliasing issue shows up (I don't think you see the optimizer effects on this section of the code in 64 bit mode).

core/utils/checksum.h
nrdmao: Will do.

nrdmao: Actually on second look, this code block is all x86 assembly code so it seems like x86_64 is the right thing to do, unless I misunderstand you.

sangjinhan: Oh you are right. What the heck I was thinking about...

core/utils/checksum.h
sangjinhan: I see. What I did was just build test (with `-m32` option to g++)i, as I don't have a 32-bit env to do runtime test. While it's entirely cosmetic, perhaps we can use `[[gnu:may_alias]]` instead?

sangjinhan:
Please add your name to CONTRIBUTORS.md :)

sangjinhan:
Thanks!

gxldem8kxnffpy9mq5bdgun4o1_1280

PR #373, opened 2017-04-08
Fix alloc-free mismatch in DRR module

sangjinhan:
In DRR module, there are two places where `llring` is allocated with a malloc-family function, but freed with C++ `delete`. This may cause potential memory corruption.

Also ran `clang-format` on the header file.


barath:
img_2667

PR #374, opened 2017-04-10
Add 32bit build/test to Travis

sangjinhan:
This PR contains patches for the Travis build environments. There are a few main changes to note:
* Add 32-bit i686 build and test environment to Travis.
  * 32-bit Docket images are added to Docker Hub (with `_32` suffices to the tag)
  * Since Travis only supports x86_64, for the i686 case BESS is cross-built and run.
* Consolidate all Vagrant/Travis/Docker related scripts in `env/`.
* Reduce the default amount hugepages for each NUMA node, from 2GB to 1GB
  * For VMs running on a less beefy machine (e.g., laptops), 4GB for a VM (2GB for hugepages and 2GB for normal memory) might be too much.
  * On 32-bit machines, it is challenging to reserve 2GB.
  * The 2GB default was set when BESS was being tested with 8192 containers. Since this configuration is somewhat extreme, 1GB should be enough.
  * One can still increase the amount of memory with `-m` option to `bessd`. E.g., `bessctl daemon start -m 2048`
  * Once merged, I will update wiki pages accordingly.
* Re-enable clang's Address/Leak/UndefinedBehavior sanitizers. While the travis script builds BESS in a container, `SANITIZE` env variable was not passed to the container.


nrdmao:
Looks good. Thanks Sangjin for doing this!

changlan:
Merged, thanks!
image

PR #376, opened 2017-04-12
bessctl: Fix 'command module'.

ddiproietto:
'ARG_TYPE' should be handled by get_var_attrs(), otherwise the cli will
expect a literal 'ARG_TYPE'.

Error conditions are handled without killing bessctl.

If run_module_command() is passed a wrong message name it fails instead
of defaulting to EmptyArg (to support EmptyArg we now include it in
module_msg).

'show mclass' now also shows the type of protobuf message required for
each command.

Finally, [CMD_ARGS...] is optional, so we shouldn't fail if it's not
provided.


changlan:
Thanks!
image

PR #377, opened 2017-04-13
Unload plugins

changlan:
- Use static object's constructor and destructor to initialize and destruct a module class.
- Check if workers are running before importing/unloading plugins.
- Misc. fixes (renaming, comments, etc.)

### DIFF
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -383,6 +383,7 @@ bool LoadPlugin(const std::string &path) {
 bool UnloadPlugin(const std::string &path) {
   auto it = plugin_handles.find(path);
   if (it == plugin_handles.end()) {
+    VLOG(1) << "Plugin " << path << " not found.";
     return false;
   }
   bool result = (dlclose(it->second) == 0);
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -368,11 +368,31 @@ static inline bool HasSuffix(const std::string &s, const std::string &suffix) {
          std::equal(suffix.rbegin(), suffix.rend(), s.rbegin());
 }
 
-bool LoadModule(const std::string &module_path) {
-  return (dlopen(module_path.c_str(), RTLD_NOW) != nullptr);
+static std::unordered_map<std::string, void *> plugin_handles;
+
+bool LoadPlugin(const std::string &path) {
+  void *handle = dlopen(path.c_str(), RTLD_NOW);
+  if (handle != nullptr) {
+    plugin_handles.emplace(path, handle);
+    return true;
+  } else {
+    return false;
+  }
 }
 
-bool LoadModules(const std::string &directory) {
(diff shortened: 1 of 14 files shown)


changlan:
Not ready yet. Libraries compiled with GCC do not execute destructors upon `dlclose()`.

core/bessd.cc
```diff
@@ -368,11 +368,32 @@ static inline bool HasSuffix(const std::string &s, const std::string &suffix) {
          std::equal(suffix.rbegin(), suffix.rend(), s.rbegin());
 }
 
-bool LoadModule(const std::string &module_path) {
-  return (dlopen(module_path.c_str(), RTLD_NOW) != nullptr);
+static std::unordered_map<std::string, void *> plugin_handles;
```
barath: Could you add a comment on key / values for this.

core/bessd.cc
```diff
@@ -368,11 +368,32 @@ static inline bool HasSuffix(const std::string &s, const std::string &suffix) {
...
+
+bool LoadPlugin(const std::string &path) {
+  void *handle = dlopen(path.c_str(), RTLD_NOW);
+  if (handle != nullptr) {
+    plugin_handles.emplace(path, handle);
+    return true;
+  } else {
```
barath: No need for else.

core/bessd.cc
```diff
@@ -368,11 +368,32 @@ static inline bool HasSuffix(const std::string &s, const std::string &suffix) {
...
+  }
+}
+
+bool UnloadPlugin(const std::string &path) {
+  auto it = plugin_handles.find(path);
+  if (it == plugin_handles.end()) {
+    VLOG(1) << "Plugin " << path << " not found.";
+    return false;
+  }
+  bool result = (dlclose(it->second) == 0);
```
barath: Minor -- maybe rename `result` to `ok` or `success`?

barath:
Also, not totally sure, but this might need to be rebased due to the TC merge.

sangjinhan:
Is this the issue the `-fno-gnu-unique` option trying to address? (adding here for future reference) https://www.redhat.com/archives/posix-c++-wg/2009-August/msg00002.html

What the article says "static class members" (it seems we don't have static variables in inline functions) are the offending variables which prevents from unloading? These are all I found such variables:
* `static std::map<std::string, Module *> all_modules_` (class Module)
* `static std::map<std::string, Port *> all_ports_` (class Port)
* `static std::unordered_map<std::string, TrafficClass *> all_tcs_` (class TrafficClass)

Ideally module plugins should never have a reference to those symbols... Perhaps we should separate `class Module` (what plugins implement) from `class ModuleWrapper` (internally used by the BESS framework) in the long term.

changlan:
Good point. Making an issue so that we can keep track.

PR #378, opened 2017-04-13
scheduler: Fix destructor of an empty Scheduler.

ddiproietto:
Since the previous commit, the scheduler can be empty (root_ == nullptr).
The destructor must handle this case correctly, otherwise bessd will
crash.

The bug can be simply reproduced with:

./bessctl daemon start -- add worker 0 0 -- delete worker 0 -- daemon stop

Reported by Sangjin (thanks!)


barath:
kona zen card

PR #380, opened 2017-04-14
Recalculate checksum with TTL update

shinae-woo:
- TTL always be decreased by 1, so remove TODO for customizable parameter.
- Change free_batch to drop_batch and forward dropped packets into
output gate 1, so that someone may take the dropped packet later.
(E.g., sending ICMP TTL exceeded in transit messages)


sangjinhan:
Merged, thanks!

PR #384, opened 2017-04-15
Replace environment variable names in .bess files

shinae-woo:
- Do not use `SN_*`, but use `BESS_*`


justinemarie:
This is great. Thank you for the clean-up work!
image

PR #385, opened 2017-04-15
Avoid undefined bit shift operations

sangjinhan:
This PR fixes subtle bugs caused by undefined bit shift operations. For example in the following code from `core/modules/exact_match.cc`
```
if (field.mask() == 0) {
    /* by default all bits are considered */
    f->mask = ((uint64_t)1 << (f->size * 8)) - 1;
```

The code behavior is unknown when `f->size == 8`, since `1ull << 64` is an undefined operation in C++. On pre-Haswell machines the code works as intended (`f->mask` becomes `0b111111...111`). However, on Haswell or later `f->mask` results in `0b000000...000`. This difference comes from the use of new bit instructions (BMI) on newer processors. As a result, the ExactMatch module does not work correctly with 8-byte fields on some machines.

### DIFF
--- a/core/modules/l2_forward.cc
+++ b/core/modules/l2_forward.cc
@@ -99,6 +99,7 @@ const union {
 } _mask = {.val = {0x8000ffffFFFFffffull, 0x8000ffffFFFFffffull,
                    0x8000ffffFFFFffffull, 0x8000ffffFFFFffffull}};
 
+// Do not call these functions directly. Use find_index() instead. See below.
 static inline int find_index_avx(uint64_t addr, uint64_t *table) {
   __m256d _addr = (__m256d)_mm256_set1_epi64x(addr | (1ull << 63));
   __m256d _table = _mm256_load_pd((double *)table);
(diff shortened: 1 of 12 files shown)


core/modules/l2_forward.cc
```diff
@@ -96,11 +96,11 @@ static uint32_t l2_alt_index(uint32_t hash, uint32_t size_power,
...
   uint64_t val[4];
   __m256d _mask;
-} _mask = {.val = {0x8000ffffFFFFfffflu, 0x8000ffffFFFFfffflu,
-                   0x8000ffffFFFFfffflu, 0x8000ffffFFFFfffflu}};
+} _mask = {.val = {0x8000ffffFFFFffffull, 0x8000ffffFFFFffffull,
+                   0x8000ffffFFFFffffull, 0x8000ffffFFFFffffull}};
 
 static inline int find_index_avx(uint64_t addr, uint64_t *table) {
-  __m256d _addr = (__m256d)_mm256_set1_epi64x(addr | ((uint64_t)1 << 63));
+  __m256d _addr = (__m256d)_mm256_set1_epi64x(addr | (1ull << 63));
```
barath: I know you're not making significant changes to this function (and the other static function below), but could you add a comment for these functions anyway?

barath:
Is there any way to add a unit test for this?

core/modules/l2_forward.cc
sangjinhan: Sure thing, will add.

sangjinhan:
There would be no general way to prevent future misuse of shift operations. Modules could be tested in integration tests, but let's do that in a separate PR.

barath:
photo on 2011-12-25 at 11 37

PR #387, opened 2017-04-16
Increase max # of worker threads: 4 -> 64

sangjinhan:
The hardcoded limit of 4 CPU cores may not be enough for some use cases...

### DIFF
--- a/core/worker.h
+++ b/core/worker.h
@@ -15,8 +15,6 @@
 
 #define MAX_WORKERS 4
 
-#define MAX_MODULES_PER_PATH 256
-
 // XXX
 typedef uint16_t gate_idx_t;
 #define MAX_GATES 8192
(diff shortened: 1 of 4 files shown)


core/worker.h
```diff
@@ -99,8 +97,8 @@ class Worker {
  private:
   volatile worker_status_t status_;
 
-  int wid_;  /* always [0, MAX_WORKERS - 1] */
-  int core_; /* TODO: should be cpuset_t */
+  int wid_;   // always [0, kMaxWorkers - 1]
+  int core_;  // TODO: should be cpuset_t
```
barath: While updating this, want to make the change to cpuset_t as the TODO indicates?  (If it's more involved, never mind.)

sangjinhan: It is about having a set of CPU cores (not only one) on which a worker thread can be scheduled by OS (if the set has just one core, the behavior is same as before). Since it would require many changes here and there, let's do it sometime later in a separate PR.

melvinw:
Thanks! cat

PR #392, opened 2017-04-17
Add pre-commit hooks to format C++/Python code

changlan:
Run `./install_git_hooks.sh` to enable.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -7,7 +7,7 @@ PY_FORMAT="autopep8"
 PY_EXTS=".py"
 
 command_exists() {
-	hash $1 2>/dev/null
+	command -v $1 >/dev/null 2>&1
 }
 
 match_extension() {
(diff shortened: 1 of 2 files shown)


sangjinhan:
Two nitpicky suggestions:
- How about `_` (underscore) rather than `-`, for consistency with other filenames?
- "hook" is a overloaded term since we have "gate hooks" in the framework. "install_git_hooks.sh" might be better..?

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+	local filename=$(basename $1)
+	local extension=".${filename##*.}"
+	local ext
+
+	for ext in $2; do [[ "$ext" == "$extension" ]] && return 0; done
+	return 1
+}
+
+if ! command_exists $CXX_FORMAT; then
+	printf "Error: $CXX_FORMAT executable not found.\n"
```
sangjinhan: to stderr

changlan:
Sounds good. Updated.

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+CXX_FORMAT="clang-format"
+CXX_EXTS=".c .h .cpp .hpp .cc .hh .cxx .hxx"
+
+PY_FORMAT="autopep8"
+PY_EXTS=".py"
+
+command_exists() {
+	hash $1 2>/dev/null
```
sangjinhan: might be `which` is better since it is free of side effects

barath:
I have a few tweaks to the .clang-format file; I'll make those a separate PR.

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+
+if ! command_exists $PY_FORMAT; then
+	printf "Error: $PY_FORMAT executable not found.\n"
+	exit 1
+fi
+
+for file in `git diff-index --cached --name-only HEAD` ; do
+	printf "$file"
+	if match_extension $file "$CXX_EXTS"; then
+		$CXX_FORMAT -i -style=file $file
```
sangjinhan: `-style=file`...?

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+	exit 1
+fi
+
+if ! command_exists $PY_FORMAT; then
+	printf "Error: $PY_FORMAT executable not found.\n"
+	exit 1
+fi
+
+for file in `git diff-index --cached --name-only HEAD` ; do
+	printf "$file"
```
sangjinhan: More descriptive message here would be nice

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+}
+
+if ! command_exists $CXX_FORMAT; then
+	printf "Error: $CXX_FORMAT executable not found.\n"
+	exit 1
+fi
+
+if ! command_exists $PY_FORMAT; then
+	printf "Error: $PY_FORMAT executable not found.\n"
+	exit 1
```
sangjinhan: I am not sure whether we should give one a prompt to ask if they want to continue nontheless...

.hooks/pre-commit
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+for file in `git diff-index --cached --name-only HEAD` ; do
+	printf "$file"
+	if match_extension $file "$CXX_EXTS"; then
+		$CXX_FORMAT -i -style=file $file
+	elif match_extension $file "$PY_EXTS"; then
+		$PY_FORMAT -i $file
+	fi
+	git add $file
```
sangjinhan: How about showing diff made by the formatter (if any) and asks if the user wishes to continue committing the files? Sorry it would need a non-trivial amount of additional work...

.hooks/pre-commit
changlan: It means `clang-format` should use `.clang-format` file in a parent directory.

.hooks/pre-commit
changlan: I was confused by this post: http://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script. Which one is the best?

sangjinhan: Oh good to know. `command -v` seems good since it is portable and side effect free..?

.hooks/pre-commit
sangjinhan: NVM, didn't realize I still can do `git commit --no-verify`

sangjinhan:
Merged, thanks!
tumblr_m4rw7zfaev1r6jd7fo1_1280

Issue #388, opened 2017-04-17
Codecov test coverage doesn't make much sense


sangjinhan:
I am somewhat puzzled by the Codecov results. Coverage graph fluctuates for some non-obvious reasons...

https://codecov.io/gh/NetSys/bess/

justinemarie:
It also doesn't include the python tests, right? Just the C++/Google tests?

sangjinhan:
We are not even sure about if that's the case. Need to verify :-)

justinemarie:
Yikes!

sirspinach:
Python tests (`./sanity_check.sh` and `bessctl -- run testing/run_module_tests`) should be included in code coverage count. At least that is how the .travis.yml is set up.

sirspinach:
In this job, code coverage according to codecov.io fell to 4%. https://travis-ci.org/NetSys/bess/jobs/221876348

It would be nice if at the very bottom, along with submitting to codecov.io, we could see the local machine output what it thinks code coverage is.

changlan:
Let me look into this.

changlan:
Should be fixed by #492.

sangjinhan:
Not sure if it was actually fixed. #499 decreases the coverage from 64% to 58%, although it is a relatively small change....

changlan:
The number seems pretty stable now.

PR #395, opened 2017-04-17
travis: output coverage percentage from genhtml

sirspinach:
Adding this output should make it clearer whether our fluctuating
code coverage is a problem with codecov.io or a problem with our
tests.

The coverage percentage generated by genhtml will not be the same as
the percentage generated by codecov.io (because aren't looking at
all of the same files and count differently for some line types),
but is roughly proportional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ before_install:
 install:
   - sudo apt-get install -y python
   - pip install grpcio scapy
+  - "[[ ${DEBUG:-0} == 0 ]] || sudo apt-get install -y lcov"
   - "[[ ${SANITIZE:-0} == 0 ]] || sudo apt-get install -y llvm-3.8"
   - "[[ $TAG_SUFFIX != _32 ]] || sudo apt-get install -y lib32gcc1"
   - ln -s /build/dpdk-17.02 deps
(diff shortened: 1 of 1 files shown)


sirspinach:
for #388

.travis.yml
```diff
@@ -47,7 +48,7 @@ script:
   - ./bessctl/bessctl -- daemon start -- run testing/run_module_tests
 
 after_success:
-  - "[[ ${DEBUG:-0} == 0 ]] || bash <(curl -s https://codecov.io/bash)"
+  - "[[ ${DEBUG:-0} == 0 ]] || (bash <(curl -s https://codecov.io/bash) && cd core && ./coverage run)"
```
sangjinhan: How about `(( ... || ...) && ( ... && ...))`, to make the command side-effect-free and clearer in terms of operator precedences?

sangjinhan: (side effect from `cd`)

sirspinach: hmmm... I think the `cd` has no side effect outside of its subshell. But yeah, I'll make it clearer.

sangjinhan:
tumblr_m7dyx6y7jq1qafbq7o1_1280

PR #397, opened 2017-04-18
Fix and enhance NAT checksum update

ddiproietto:


core/utils/checksum.h
```diff
@@ -298,17 +298,46 @@ static inline uint16_t CalculateIpv4TcpChecksum(const Ipv4Header &iph,
                                   ntohs(iph.length) - (iph.header_length << 2));
 }
 
+// Incremental checksum udpdate
```
barath: update

core/utils/icmp.h
```diff
@@ -5,8 +5,8 @@ namespace utils {
 
 // A basic ICMP header definition.
 struct[[gnu::packed]] IcmpHeader {
-  uint16_t type;      // ICMP packet type.
-  uint16_t code;      // ICMP packet type.
+  uint8_t type;       // ICMP packet type.
```
barath: Given that this was off and we didn't catch it before, I wonder if there's an easy way for you to add a test / tests to this PR for the various checksum calculations?

core/utils/checksum.h
ddiproietto: fixed, thanks

core/utils/icmp.h
ddiproietto: good point, will do that

barath:
Thanks, LGTM.  Will merge after CI finishes.

barath:
photo 110

PR #404, opened 2017-04-19
Allow python modules to be created for existing BESS modules.

justinemarie:
This allows you to instantiate a module in python for a BESS module that already exists; just invoke the module by name and declare _do_not_create in the constructor. An example is in conf/samples/reattachment.bess.


justinemarie:
Wait one more thing I need to add (check to make sure module exists).

justinemarie:
Okay put the right check in now :-)

bessctl/module.py
```diff
@@ -22,10 +22,21 @@ def __init__(self, **kwargs):
...
+            #create an object
+            ret = self.bess.create_module(self.__class__.__name__, name,
                                       self.choose_arg(None, kwargs))
-
-        self.name = ret.name
+            self.name = ret.name
+        else:
+            #bind to a pre-existing object, check if it's real
+            assert self.name != None
+            all_modules = self.bess.list_modules().modules
```
sangjinhan: `list_modules()` enumerates all modules, thus potentially expensive if there are a lot. Instead you can use (untested):
```
info = self.bess.get_module_info(name)
check if self.__class__.__name__ agrees with info.mclass
```

bessctl/module.py
```diff
@@ -22,10 +22,21 @@ def __init__(self, **kwargs):
...
+        if(not _do_not_create):
+            #create an object
+            ret = self.bess.create_module(self.__class__.__name__, name,
                                       self.choose_arg(None, kwargs))
-
-        self.name = ret.name
+            self.name = ret.name
+        else:
+            #bind to a pre-existing object, check if it's real
+            assert self.name != None
```
sangjinhan: `name`, not `self.name`

sanity_check.sh
```diff
@@ -13,7 +13,8 @@ TESTS='exactmatch.bess
 	l2_forward.bess
 	multicore.bess
 	queue.bess
-	reattachment.bess
+  reattachment.bess
+  reattachment.bess
```
sangjinhan: Why twice?

Also, the file has mixed use of tabs and spaces.

bessctl/module.py
justinemarie: fixed

bessctl/module.py
justinemarie: fixed

sanity_check.sh
justinemarie: Copy/paste error :-) Fixed.

justinemarie:
This is ready for another lookthrough and/or commit as soon as it finishes in Travis.

bessctl/module.py
```diff
@@ -22,10 +22,18 @@ def __init__(self, **kwargs):
         else:
             name = None
 
-        ret = self.bess.create_module(self.__class__.__name__, name,
+        if(not _do_not_create):
```
sangjinhan: no parenthesis here

bessctl/module.py
```diff
@@ -22,10 +22,18 @@ def __init__(self, **kwargs):
...
+            ret = self.bess.create_module(self.__class__.__name__, name,
                                       self.choose_arg(None, kwargs))
-
-        self.name = ret.name
+            self.name = ret.name
+        else:
+            #bind to a pre-existing object, check if it's real
+            assert name != None
+            info = self.bess.get_module_info(name)
+            assert info != None
```
sangjinhan: Checking `info` here doesn't do anything. It won't return `None` at all. Just either success or exception.

sanity_check.sh
```diff
@@ -29,7 +30,7 @@ function fail
 }
 
 echo "This script runs a collection of BESS sample scripts and makes sure \
-nothing blows up. Sit back and relax."
+  nothing blows up. Sit back and relax."
```
sangjinhan: You have three spaces between `sure` and `nothing`.

bessctl/conf/samples/reattachment.bess
```diff
@@ -0,0 +1,50 @@
...
+
+# locally emulate two input ports
+Source() -> SetMetadata(attrs=[{'name': 'in_port', 'size': 2, 'value_int': 1}]) -> Rewrite(templates=pkts) -> wm
+
+wm:0 -> Sink()
+wm:1 -> Sink()
+wm:2 -> Sink()
+wm.set_default_gate(gate=2)
+
+# 172.16.100.1/32, port 80, in_port 1
```
sangjinhan: As it is derived from the `wildcardmatch.bess` example, we have quite redundancy which is not quite relevant to the point of this test script. I would suggest keeping only `wm` and nothing else to make the purpose of the test clearer.

sanity_check.sh
```diff
@@ -5,21 +5,22 @@ SCRIPTS='./bessctl/conf/samples'
...
-	wildcardmatch.bess'
+flowgen.bess
+generic_encap.bess
+hash_lb.bess
+igate.bess
+iplookup.bess
+l2_forward.bess
+multicore.bess
+queue.bess
+reattachment.bess
```
sangjinhan: Perhaps a bit off the topic, but we should be careful to add files here. This sanity test takes about half of the total Travis build time, being the most time consuming step (and keeps growing).

Given that what this script does is mostly identical to `wildcardmatch.bess` below, how about not adding this file, until we derive a solution for the sanity check to shorten its test time?

bessctl/module.py
```diff
@@ -22,10 +22,18 @@ def __init__(self, **kwargs):
...
-        ret = self.bess.create_module(self.__class__.__name__, name,
+        if(not _do_not_create):
+            #create an object
+            ret = self.bess.create_module(self.__class__.__name__, name,
                                       self.choose_arg(None, kwargs))
-
-        self.name = ret.name
+            self.name = ret.name
+        else:
+            #bind to a pre-existing object, check if it's real
```
sangjinhan: Space

bessctl/module.py
```diff
@@ -22,10 +22,18 @@ def __init__(self, **kwargs):
         else:
             name = None
 
-        ret = self.bess.create_module(self.__class__.__name__, name,
+        if(not _do_not_create):
+            #create an object
```
sangjinhan: Space

sanity_check.sh
justinemarie: Good point. I will do this. Why don't I just put the reattachment bit inside wildcardmatch.bess?

sangjinhan: That should work too!

bessctl/module.py
justinemarie: Fixed (why?)

bessctl/module.py
justinemarie: What does this mean?

bessctl/module.py
justinemarie: Fixed

sanity_check.sh
justinemarie: fixed

justinemarie:
*bump*

melvinw:
Is `conf/samples/reattachment.bess` supposed to be part of this PR? I can't find it.

justinemarie:
No, it has been removed.

bessctl/module.py
```diff
@@ -22,10 +22,17 @@ def __init__(self, **kwargs):
...
-
-        self.name = ret.name
+        if not _do_not_create:
+            # create an object
+            ret = self.bess.create_module(self.__class__.__name__, name,
+                                          self.choose_arg(None, kwargs))
+            self.name = ret.name
+        else:
+            # bind to a pre-existing object, check if it's real
+            assert name != None
```
melvinw: Could you add an error message to this assertion (e.g. https://github.com/NetSys/bess/blob/master/bessctl/commands.py#L92)

bessctl/module.py
```diff
@@ -22,10 +22,17 @@ def __init__(self, **kwargs):
...
+            info = self.bess.get_module_info(name)
+            assert self.mclass == info.mclass
```
melvinw: Same here

bessctl/module.py
justinemarie: Fixed.

bessctl/module.py
justinemarie: Fixed.

melvinw:
Thanks!
cat

PR #409, opened 2017-04-21
module: Add burst parameter to FlowGen

changlan:
So that we can change “burst” parameter at runtime (akin to Source).

### DIFF
--- a/bessctl/conf/perftest/flowgen.bess
+++ b/bessctl/conf/perftest/flowgen.bess
@@ -32,7 +32,7 @@ flowgens = []
 for i in range(num_cores):
     bess.add_worker(wid=i, core=i)
     pkt_src = FlowGen(template = pkt_template, pps = init_pps / num_cores, flow_rate = num_flows / flow_time, flow_duration = flow_time, \
-        arrival = 'uniform', duration='uniform', quick_rampup = True)
+        arrival = 'uniform', duration='uniform', quick_rampup = True, burst = 32)
     flowgens.append(pkt_src)
     bess.attach_module(pkt_src.name, wid=i)
 
@@ -56,4 +56,3 @@ while True:
     bess.resume_all()
     time.sleep(sleeptime) # give it some time to warm up to the new rate
     prev_portstats = myport.get_port_stats().inc.packets
-
--- a/bessctl/conf/perftest/flowgen.bess
+++ b/bessctl/conf/perftest/flowgen.bess
@@ -32,7 +32,7 @@ flowgens = []
 for i in range(num_cores):
     bess.add_worker(wid=i, core=i)
     pkt_src = FlowGen(template = pkt_template, pps = init_pps / num_cores, flow_rate = num_flows / flow_time, flow_duration = flow_time, \
-        arrival = 'uniform', duration='uniform', quick_rampup = True, burst = 32)
+        arrival = 'uniform', duration='uniform', quick_rampup = True)
     flowgens.append(pkt_src)
     bess.attach_module(pkt_src.name, wid=i)
 
--- a/core/modules/flowgen.cc
+++ b/core/modules/flowgen.cc
@@ -201,6 +201,13 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
     quick_rampup_ = 1;
   }
 
+  if (arg.burst() <= bess::PacketBatch::kMaxBurst) {
+    burst_ = arg.burst();
+  } else {
+    return pb_error(EINVAL, "'burst' must be no greater than %ld",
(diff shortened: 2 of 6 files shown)


core/modules/flowgen.cc
```diff
@@ -200,6 +200,13 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
     quick_rampup_ = 1;
   }
 
+  if (arg.burst() <= bess::PacketBatch::kMaxBurst) {
+    burst_ = arg.burst();
+  } else {
+    return pb_error(EINVAL, "'burst' must be no greater than %ld",
```
sangjinhan: `%ld` has different sizes in i386 and x86_64. `PRId64` would be better.

core/modules/flowgen.cc
```diff
@@ -200,6 +200,13 @@ pb_error_t FlowGen::ProcessArguments(const bess::pb::FlowGenArg &arg) {
     quick_rampup_ = 1;
   }
 
+  if (arg.burst() <= bess::PacketBatch::kMaxBurst) {
```
sangjinhan: Could you separate `set_burst()` from this function, like other modules (such as `Source`, `PortInc`, etc.)? By having a separate function one can adjust burstiness of modules without doing `bess.pause_all()`

sangjinhan: Another delicate point: `Init()` should set burst to 32 (default) if 0 is given, since it may have been unspecified in protobuf. `set_burst()` should accept 0 (generating no more packets) since it is explicitly given.

bessctl/conf/perftest/flowgen.bess
```diff
@@ -32,7 +32,7 @@ flowgens = []
 for i in range(num_cores):
     bess.add_worker(wid=i, core=i)
     pkt_src = FlowGen(template = pkt_template, pps = init_pps / num_cores, flow_rate = num_flows / flow_time, flow_duration = flow_time, \
-        arrival = 'uniform', duration='uniform', quick_rampup = True)
+        arrival = 'uniform', duration='uniform', quick_rampup = True, burst = 32)
```
sangjinhan: Feel free to ignore this: we can skip this (in favor of brevity over completeness)

core/modules/flowgen.cc
changlan: `FlowGen` already has a `update()` command. How about using that command? (I am fine with adding a new command if you prefer)

sangjinhan: `update()` is not thread safe, so one must pause the entire pipeline with `bess.pause_all()`

protobuf/module_msg.proto
```diff
@@ -470,6 +478,7 @@ message FlowGenArg {
   uint32 ip_dst_range = 9; /// When generating new flows, FlowGen modifies the template packet by changing the IP dst, incrementing it by at most ip_dst_range.
   uint32 port_src_range = 10; /// When generating new flows, FlowGen modifies the template packet by changing the TCP port, incrementing it by at most port_src_range.
   uint32 port_dst_range = 11; /// When generating new flows, FlowGen modifies the template packet by changing the TCP dst port, incrementing it by at most port_dst_range.
+  uint64 burst = 12; /// The number of packets per batch to release.
```
sangjinhan: Sorry about this additional roundtrip --  how about removing `burst` from the `update()`? One can just call `update()` followed by `set_burst()`, so the redundancy doesn't seem to be strictly necessary.

core/modules/flowgen.cc
```diff
@@ -22,7 +22,9 @@ using bess::utils::TcpHeader;
 const double PARETO_TAIL_LIMIT = 0.99;
 
 const Commands FlowGen::cmds = {
-    {"update", "FlowGenArg", MODULE_CMD_FUNC(&FlowGen::CommandUpdate), 0}};
+    {"update", "FlowGenArg", MODULE_CMD_FUNC(&FlowGen::CommandUpdate), 0},
+    {"set_burst", "FlowGenCommandSetBurstArg",
+     MODULE_CMD_FUNC(&FlowGen::CommandSetBurst), 0}};
```
sangjinhan: `mt_safe` should be 1

sangjinhan:
Thanks, mind if I do squash merge?

sangjinhan:
Also don't forget to update the protobuf file too

changlan:
Oh, you mean remove `burst` in `FlowGenArg` as well?

sangjinhan:
Never mind, I forgot `Update` and `Init` share the same parameters :-)

sangjinhan:
gxldem8kxodkzmrvqccsom2zo1_500

PR #410, opened 2017-04-22
Module constraints part 2

apanda:
This improves on the previous patch in a few ways:

- We now account for all modules called by a worker when deciding on placement.

- We check to see that only thread safe modules are called from multiple workers.

The only bit left is to perhaps always run this when `resume_all` is called, however there are some semantic challenges there which I need to work through.


core/module.h
```diff
@@ -409,10 +428,20 @@ class Task {
   }
 
   int GetSocketConstraints() const {
-    if (t_) {
-      return t_->GetSocketConstraints();
+    if (module_) {
```
barath: Method comment.

core/module.h
```diff
@@ -409,10 +428,20 @@ class Task {
...
+      return module_->ComputePlacementConstraints();
+    } else {
+      return UNCONSTRAINED_SOCKET;
+    }
+  }
+
+  void AddActiveWorker(int wid) {
```
barath: Method comment.

core/module.cc
```diff
@@ -248,6 +245,58 @@ void Module::DeregisterAllAttributes() {
...
+  if (num_active_workers < min_allowed_workers_ ||
+      num_active_workers > max_allowed_workers_) {
+    LOG(ERROR) << "Mismatch in number of workers for module " << name_
+               << " min required " << min_allowed_workers_ << " max allowed "
+               << max_allowed_workers_ << " attached workers "
+               << num_active_workers << " returning fatal";
+    return CHECK_FATAL_ERROR;
+  }
+
+  for (auto widx : active_workers_) {
```
barath: widx -> wid (for consistency)?

core/module.h
```diff
@@ -155,6 +158,12 @@ class ModuleBuilder {
 
 class ModuleTask;
 
+enum CheckResults {
```
barath: Could you give this a more descriptive name (since I could imagine other things being checked) and add a comment?  For the name, another option is to move it someplace more general and leave it as is.

core/module.h
```diff
@@ -166,7 +175,11 @@ class Module {
         attr_offsets_(),
         tasks_(),
         igates_(),
-        ogates_() {}
+        ogates_(),
+        active_workers_(),
+        node_constraints_(UNCONSTRAINED_SOCKET),
+        min_allowed_workers_(1),
```
barath: In the common case, is the goal of the min to find modules that never got connected?

core/module.h
```diff
@@ -271,6 +279,18 @@ class Module {
 
   const std::vector<bess::OGate *> &ogates() const { return ogates_; }
 
+  int ComputePlacementConstraints() const;
```
barath: Method comment.  Also, would it be better to use a uint64_t just to be explicit?

core/module.h
```diff
@@ -271,6 +279,18 @@ class Module {
...
+
+  void ResetActiveWorkerSet() { active_workers_.clear(); }
+
+  int GetActiveWorkerCount() { return active_workers_.size(); }
```
barath: Is this method needed?

core/module.h
```diff
@@ -271,6 +279,18 @@ class Module {
...
+
+  const std::unordered_set<int> &GetActiveWorkers() { return active_workers_; }
```
barath: Per style guide, rename to active_workers().  Also, method can be made const.

core/module.h
apanda: Precisely.

apanda:
@barath Your comments should have been addressed.

core/module.cc
```diff
@@ -286,10 +286,10 @@ CheckResults Module::CheckModuleConstraints() const {
     return CHECK_FATAL_ERROR;
   }
 
-  for (auto widx : active_workers_) {
-    int socket = 1 << workers[widx]->socket();
+  for (auto wid : active_workers_) {
+    int socket = 1 << workers[wid]->socket();
```
barath: Should this be placement_constraint and 1ull as well?

core/modules/port_inc.cc
```diff
@@ -11,7 +11,7 @@ pb_error_t PortInc::Init(const bess::pb::PortIncArg &arg) {
   queue_t num_inc_q;
   int ret;
   pb_error_t err;
-  int placement_constraint;
+  int placement;
```
barath: Should this now be type placement_constraint?

apanda: Good catch, fixed.

bessctl/conf/samples/multicore-phy.bess
```diff
@@ -1,38 +1,14 @@
...
-# 3. "daemon reset" should remove workers as well
-
-# Syntax: bind worker:<worker ID> => core:<core ID>
-# Bind a worker thread to the specified CPU core.
-# It creates a new worker thread if not already exists.
-#bind worker:0 => core:0
-#bind worker:1 => core:1
-#
+# This sample exists merely as a way to check
+# that constraint detection is correctly working.
```
apanda: Done below.

barath:
photo on 2014-08-21 at 18 39 2

PR #411, opened 2017-04-24
protobuf: Convert camelCase to snake_case

changlan:
Modules commands in the comments are in camelCase. BESS code uses
snake_case.


justinemarie:
This is very helpful, thank you! At some point I need to think about how to re-do the auto-gen code -- I think my choice to autogen from protobuf was not the right one.

justinemarie:
image

PR #412, opened 2017-04-24
Add clear() command to Measure module

sangjinhan:
At the moment, `Measure` resets the two histograms (RTT and jitter) (with 1M elements each) only when `Histogram::Insert()` is called, i.e., when packets are received. It takes about 10ms, artificially inflating the latency of pending packets. The measurement results will show higher tail latency than it should be.

As a quick fix, this PR adds `clear()` command to reset histograms in an explicit manner, so that the Measure module can perform `Histogram::Insert()` without delayed by histogram reset.

```
bess.pause_all()
stats = measure.get_summary()
measure.clear()                 <- histograms are reset here
bess.resume_all()
...                             <- not here
```


changlan:
image

PR #413, opened 2017-04-25
Add and connect modules without stopping workers

ddiproietto:
WIth this changes, it's possible to add modules and connect them (as long as they're not in use).

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -784,7 +784,7 @@ def add_module(cli, mclass, module, pause_workers, args):
 
 @cmd('add connection MODULE MODULE [OGATE] [IGATE] [PAUSE_WORKERS]',
      'Add a connection between two modules')
-def add_connection(cli, m1, m2, ogate, igate, pause_workers):
+def add_connection(cli, m1, m2, ogate, igate, pause_workers='pause'):
     if ogate is None:
         ogate = 0
 
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -796,7 +796,6 @@ def add_connection(cli, m1, m2, ogate, igate, pause_workers):
     try:
         cli.bess.connect_modules(m1, m2, ogate, igate)
     finally:
-        cli.bess.resume_all()
         if pause_workers != 'no_pause':
             cli.bess.resume_all()
 
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -782,20 +782,23 @@ def add_module(cli, mclass, module, pause_workers, args):
         cli.fout.write('  The new module "%s" has been created\n' % ret.name)
 
 
-@cmd('add connection MODULE MODULE [OGATE] [IGATE]',
+@cmd('add connection MODULE MODULE [OGATE] [IGATE] [PAUSE_WORKERS]',
      'Add a connection between two modules')
-def add_connection(cli, m1, m2, ogate, igate):
+def add_connection(cli, m1, m2, ogate, igate, pause_workers):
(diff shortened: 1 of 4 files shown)


core/bessctl.cc
```diff
@@ -344,6 +344,54 @@ static Module* create_module(const std::string& name,
   return m;
 }
 
+static void collect_tc(const bess::TrafficClass* c, int wid,
```
barath: Nit: I believe these function names should be camel case to match the style guide (worth running it through Google cpplint).

core/bessctl.cc
```diff
@@ -550,17 +573,26 @@ class BESSControlImpl final : public BESSControl::Service {
       });
     }
 
+    if (request->wid() < 0) {
```
barath: Isn't request->wid() already saved in wid_filter?

Also, is this if block meant to catch the case that no specific wid was requested?

bessctl/commands.py
```diff
@@ -762,32 +767,38 @@ def add_port(cli, driver, port, args):
...
     if igate is None:
         igate = 0
 
-    cli.bess.pause_all()
+    if pause_workers != 'no_pause':
+        cli.bess.pause_all()
     try:
         cli.bess.connect_modules(m1, m2, ogate, igate)
     finally:
         cli.bess.resume_all()
```
sangjinhan: You don't need this line any longer..?

core/bessctl.cc
ddiproietto: Good point. There are several functions like this in worker (there's a C++-ify TODO comment. How about I send a separate pull request for this? Otherwise I can push an extra commit here just form the naming.

core/bessctl.cc
ddiproietto: This block is meant to catch the case where no specific wid was requested (request->wid()==-1).

request->wid() is saved into wid_filter, but then wid_filter is altered:

```
    if (wid_filter < 0) {
      i = 0;
      wid_filter = num_workers - 1;
    } else {
      i = wid_filter;
    }
```

bessctl/commands.py
ddiproietto: Oops, good catch, removed

core/bessctl.cc
barath: Sounds good -- no need to do it here; Panda's PR also runs into this.

core/bessctl.cc
barath: Ok, sounds good.

barath:
BESS code looks good, maybe someone can look at the bessctl code?

bessctl/commands.py
```diff
@@ -762,32 +767,37 @@ def add_port(cli, driver, port, args):
...
     if module is None:
         cli.fout.write('  The new module "%s" has been created\n' % ret.name)
 
 
-@cmd('add connection MODULE MODULE [OGATE] [IGATE]',
+@cmd('add connection MODULE MODULE [OGATE] [IGATE] [PAUSE_WORKERS]',
      'Add a connection between two modules')
-def add_connection(cli, m1, m2, ogate, igate):
+def add_connection(cli, m1, m2, ogate, igate, pause_workers):
     if ogate is None:
```
justinemarie: Set a default value to "no_pause" here to make clear what the default is?
(Presumably legacy calls to this function don't have this parameter either?)

ddiproietto: Added the default value (it should be "pause" to maintain the same behavior)

Thanks!

barath:
photo 74

PR #414, opened 2017-04-25
traffic_class: Fix descendant removal from RR TC.

ddiproietto:
If the last unblocked child is blocked or removed, `next_child_` must
wrap around, otherwise bessd may crash.

To reproduce:

```
daemon start
run samples/roundrobin
add module Queue queue0
delete module queue0
add module Queue queue0
delete module queue0
...
```


barath:
photo 100

PR #415, opened 2017-04-25
Add a modules option to load the modules from an arbitrary directory

nrdmao:
This option provides a degree of flexibility to BESS installations where the location of bessd is decouple from the location of the module .so's. For example, for consistency with other system daemons, one might install bessd in /usr/bin but the module .so's in /usr/lib/bess/modules. Under such circumstances bessd may be started as such: bessd -modules /usr/lib/bess/modules

### DIFF
--- a/core/main.cc
+++ b/core/main.cc
@@ -40,8 +40,8 @@ int main(int argc, char *argv[]) {
 
   // Load plugins
   if (!bess::bessd::LoadPlugins(FLAGS_modules)) {
-    PLOG(FATAL) << "LoadPlugins() failed to load from directory: " <<
-      FLAGS_modules;
+    PLOG(FATAL) << "LoadPlugins() failed to load from directory: "
+                << FLAGS_modules;
   }
 
   // TODO(barath): Make these DPDK calls generic, so as to not be so tied to
--- a/core/main.cc
+++ b/core/main.cc
@@ -39,9 +39,9 @@ int main(int argc, char *argv[]) {
   bess::bessd::WritePidfile(pidfile_fd, getpid());
 
   // Load plugins
-  if (!bess::bessd::LoadPlugins(bess::bessd::GetCurrentDirectory() +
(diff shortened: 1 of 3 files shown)


core/main.cc
```diff
@@ -39,9 +39,9 @@ int main(int argc, char *argv[]) {
   bess::bessd::WritePidfile(pidfile_fd, getpid());
 
   // Load plugins
-  if (!bess::bessd::LoadPlugins(bess::bessd::GetCurrentDirectory() +
-                                "modules")) {
-    PLOG(FATAL) << "LoadPlugins() failed";
+  if (!bess::bessd::LoadPlugins(FLAGS_modules)) {
+    PLOG(FATAL) << "LoadPlugins() failed to load from directory: " <<
```
changlan: indentation

core/opts.cc
```diff
@@ -19,7 +20,8 @@ DEFINE_bool(s, false, "Show TC statistics every second");
 DEFINE_bool(d, false, "Run BESS in debug mode (with debug log messages)");
 DEFINE_bool(a, false, "Allow multiple instances");
 DEFINE_bool(no_huge, false, "Disable hugepages");
-
+DEFINE_string(modules, bess::bessd::GetCurrentDirectory() + "modules",
+	      "Load modules from the specified directory");
```
changlan: add an empty line after this.

core/main.cc
```diff
@@ -39,9 +39,9 @@ int main(int argc, char *argv[]) {
...
+      FLAGS_modules;
```
changlan: indentation

core/opts.cc
nrdmao: OK, done.

core/main.cc
nrdmao: OK, done.

changlan:
image

PR #416, opened 2017-04-25
Constraint checks pt 3

apanda:
Run constraint checks before resuming by default. Most failures are
recorded on the log, fatal failures cause resume to fail.

### DIFF
--- a/core/module.cc
+++ b/core/module.cc
@@ -282,8 +282,11 @@ CheckConstraintResult Module::CheckModuleConstraints() const {
     LOG(ERROR) << "Mismatch in number of workers for module " << name_
                << " min required " << min_allowed_workers_ << " max allowed "
                << max_allowed_workers_ << " attached workers "
-               << num_active_workers << " returning fatal";
-    return CHECK_FATAL_ERROR;
+               << num_active_workers;
+    if (num_active_workers > max_allowed_workers_) {
+      LOG(ERROR) << "Violates thread safety, returning fatal error";
+      return CHECK_FATAL_ERROR;
+    }
   }
 
   for (auto wid : active_workers_) {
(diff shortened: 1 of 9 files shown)


core/worker.cc
```diff
@@ -105,7 +105,13 @@ static void resume_worker(int wid) {
...
 }
 
-void resume_all_workers() {
+
+/*!
+ * Attach orphan TCs to workers. Note this does not ensure optimal placement.
+ * This method can only be called when all workers are paused.
+ */
+void attach_orphans() {
+  assert(!is_any_worker_running());
```
melvinw: This should be `CHECK`. We compile with DNEDBUG, so this won't have the intended behavior unless you do a debug build

core/modules/sink.h
```diff
@@ -5,6 +5,8 @@
 
 class Sink final : public Module {
  public:
+  Sink() { max_allowed_workers_ = Worker::kMaxWorkers; }
```
barath: Are there other modules this is true for?  I.e. other modules that are inherently thread safe?

Also, what is the right annotation for the various Queue* modules?

apanda: I was not sure, I will look through the `Queue*` things to check, but at least `QueueInc` and `QueueOut` are not thread safe (they read from PMD queue).

apanda: Last commit should have fixed the concern with queue. Will walk through all the modules in general.

core/module.cc
```diff
@@ -265,11 +265,12 @@ placement_constraint Module::ComputePlacementConstraints() const {
 
 void Module::AddActiveWorker(int wid) {
   active_workers_.insert(wid);
-  LOG(WARNING) << "Added active worker " << wid << " to " << name_;
-  for (size_t i = 0; i < ogates_.size(); i++) {
-    if (ogates_[i]) {
-      auto next = static_cast<Module *>(ogates_[i]->arg());
-      next->AddActiveWorker(wid);
+  if (propagate_workers_) {
+    for (size_t i = 0; i < ogates_.size(); i++) {
```
barath: Maybe use more idiomatic loop:
for (auto ogate : ogates_) ...?

apanda:
@barath Resolved your comment.

apanda:
@barath @sangjinhan @ddiproietto added the `attach_orphans` change from the e-mail.

barath:
photo on 2012-05-12 at 13 54

PR #417, opened 2017-04-26
Do not use "name" as a field name

sangjinhan:
The Split module has a protobuf type called `SplitArg`, which has `name` field for attribute name. `name` shouldn't be used as a field name though, since it is reserved for module name, e.g., `Split(name='split0', name='attr_foo', size=4)` has two `name`s in its parameter list, causing an error.

This PR changes the field name to `attribute`, which is the field name used by other modules, including `ExactMatch`, `SetMetadata`, etc.

I think there should be some systematic way to check if `name` is misused in `module_msg.proto`, but I am not sure what the best way would be...


changlan:
image

PR #418, opened 2017-04-26
IPLookup: Parameterize max_rules and max_tbl8s

changlan:
`max_rules` and `number_tbl8s` of `rte_lpm` were hardcoded. This PR makes them parameters of the module.


sangjinhan:
Sorry I wasn't very clear on this. Could you also add `max_tbl8s` (or a different name) as well? It refers to the maximum number of IP prefixes with smaller than /24, and the default value might be not enough for wide-area FIBs.

sangjinhan:
tumblr_m7f6r7mpyd1rro1o5o1_1280

PR #419, opened 2017-04-26
module: Remove `burst` from Init arguments

changlan:
And set `burst_` to default `kMaxBurst` on module initialization


sangjinhan:
Let me add some backgrounds for this PR: 
- Given that all those modules support a separate command `set_burst()`, supporting the burst parameter also in `Init()` is somewhat redundant.
- `Init()` and `SetBurst()` behave differently with `burst == 0`, which is a little confusing.

PR #420, opened 2017-04-27
module: Optimize RunModuleSplit().

ddiproietto:
This commit adds a level of indirection to the splits_ thread local
variable: instead of keeping an entire batch for each possible output
gate, we keep a pointer to a batch.

This makes RunModuleSplit() faster, because we avoid copying the whole
batch (we only use a pointer now).

I tried running samples/acl on my laptop before and after the change
and looking at how many packets are generated by source0:

| source0 burst | before (Mpps) | after (Mpps) |
| -------------:| -------------:| ------------:|
|            32 |          83.4 |         92.5 |
|             1 |          16.7 |         17.6 |


sangjinhan:
h9hdqoysloq4wugqi75bkoddo1_400

PR #422, opened 2017-04-27
module: Fix minor memory leaks

changlan:


core/modules/url_filter.cc
```diff
@@ -135,7 +135,9 @@ inline static bess::Packet *GenerateResetPacket(
 pb_error_t UrlFilter::Init(const bess::pb::UrlFilterArg &arg) {
   for (const auto &url : arg.blacklist()) {
     if (blacklist_.find(url.host()) == blacklist_.end()) {
-      blacklist_.emplace(url.host(), Trie());
+      blacklist_.emplace(std::piecewise_construct,
```
sangjinhan: This change looks good to me.

Perhaps this is what happened in the previous code, which set off a "use-after-free" ASAN alarm?
1. `Trie` is created as a temporary object
1. `emplace()` creates a value object in place, with a copy constructor
  * (`Trie`'s implicitly declared copy constructor)
1. The temporary object is destructed.

If it is the case, besides efficiency, the root cause of the issue was that `Trie` is not safe for copy (it's not RAII) but it still has default copy constructors. How about either...
1. Making `Trie` RAII
  * It seems not strictly necessary for `Trie` to have a pointer of `Node`, instead of embedding it.
1. Adding `DISALLOW_COPY_AND_ASSIGN` to `Trie`

changlan: Nice catch. I think making `Trie` RAII is the right thing to do.

sangjinhan: Oh I just realized that it would be non-trivial, given that `Node` itself is not safe for copy at the moment. To make `Trie` RAII we will need recursive copy. I am fine with making it non-copyable if there is no urgent demand for copy constructors.

sangjinhan:
tumblr_m7mecfkqbu1qz5dg8o1_1280

PR #423, opened 2017-04-27
bessctl: allow pause/resume of individual workers

melvinw:
Some operations only operate on datastructure within a single worker
(e.g., poking the Measure module). Pausing all workers for such
operations can lead to unecessary packet drops by workers that have
nothing to do with the operation anyway. To the end of making this
behavior avoidable , this patch exposes `pause_worker()` and
`resume_worker()` and adds corresponding RPCs.

### DIFF
--- a/bessctl/conf/perftest/flowgen.bess
+++ b/bessctl/conf/perftest/flowgen.bess
@@ -28,17 +28,17 @@ pkt_template = str(eth/ip/tcp/payload)
 # This script is multi threaded but connects to a single port.
 myport = PMDPort(port_id=0, num_inc_q=num_cores, num_out_q=num_cores)
 
-flowgens = []
-for i in range(num_cores):
-    bess.add_worker(wid=i, core=i)
+flowgens = dict()
+for wid in range(num_cores):
+    bess.add_worker(wid=wid, core=wid)
     pkt_src = FlowGen(template = pkt_template, pps = init_pps / num_cores, flow_rate = num_flows / flow_time, flow_duration = flow_time, \
         arrival = 'uniform', duration='uniform', quick_rampup = True)
-    flowgens.append(pkt_src)
-    bess.attach_module(pkt_src.name, wid=i)
+    flowgens[wid] = pkt_src
+    bess.attach_module(pkt_src.name, wid=wid)
 
-    pkt_src -> IPChecksum() -> QueueOut(port=myport, qid = i)
+    pkt_src -> IPChecksum() -> QueueOut(port=myport, qid = wid)
 
-    QueueInc(port=myport, qid=i) -> Sink()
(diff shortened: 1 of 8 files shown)


melvinw:
Not quite ready. Needs a sample script

bessctl/conf/perftest/flowgen.bess
```diff
@@ -28,12 +28,12 @@ pkt_template = str(eth/ip/tcp/payload)
...
 myport = PMDPort(port_id=0, num_inc_q=num_cores, num_out_q=num_cores)
 
-flowgens = []
+flowgens = dict()
 for i in range(num_cores):
     bess.add_worker(wid=i, core=i)
     pkt_src = FlowGen(template = pkt_template, pps = init_pps / num_cores, flow_rate = num_flows / flow_time, flow_duration = flow_time, \
         arrival = 'uniform', duration='uniform', quick_rampup = True)
-    flowgens.append(pkt_src)
+    flowgens[i] = pkt_src
```
changlan: For the sake of clarity, perhaps `wid` is better than `i`.

melvinw: Good point. Done.

apanda:
Recently @sangjinhan mentioned it seemed unlikely we would pause individual workers -- in particular there were existing bugs with things like `attach_workers` if this happened. I added him to get his opinion.

sangjinhan:
Oh, what I meant was there is no "current" use case in `bessctl/conf` sample scripts, not saying it would be unnecessary in the future. There would be some scenarios that pausing a single worker is enough (e.g., manipulating a scheduler TC tree).

In general, the reason why I chose not to expose knobs for pausing/resuming individual workers is:
* It is hard to know (especially for a controller) whether pausing a single worker is enough to do some particular operations. For example, as far as the `Flowgen` modules in the sample script are concerned, it is clear that only one worker thread is dealing with the module. But in general cases it might not be true. I think we should not assume controllers doing right things...

By the way, did you check if the sample script actually works? The `update()` command seems still marked as not thread-safe. No matter whether a single worker is running or paused, it would be rejected by `bessctl.cc` if not all workers are paused (unless there was only one worker).

melvinw:
Ah, okay. I'm fine with killing this PR for now or, alternatively, marking this code as **_POTENTIALLY DANGEROUS_** and making it more clear in the documentation what the appropriate use cases may be. I'll do that and update the logic in `bessctl.cc` if we keep it open (sorry, I neglected to test the sample script before submitting this).

sangjinhan:
Perhaps you can add LOG(WARN) for `PauseWorker()` in `bessctl.cc`, saying "experimental" and "support may discontinue"?

sangjinhan:
tumblr_m7rv9znvyv1qz5dg8o1_1280

Issue #421, opened 2017-04-27
AddActiveWorker() fails with pipeline loops


changlan:
`AddActiveWorker()` loops forever for some pipeline. One example is `samples/nat.bess`:
```
Source() -> Rewrite(templates=packets) -> 0:nat:0 -> mac -> ip -> 1:nat:1 -> Sink()
```

`bessctl` command:
```
localhost:10514 $ run samples/nat
```

Output from `bessd` (with ASAN enabled):
```
...
W0427 19:48:35.617143  5857 module.h:466] Adding active worker for wid 0 to source0
ASAN:DEADLYSIGNAL
=================================================================
==5855==ERROR: AddressSanitizer: stack-overflow on address 0x7f77eee69fe8 (pc 0x000000889b1d bp 0x7f77eee6a040 sp 0x7f77eee69fd0 T2)
    #0 0x889b1c in std::_Hashtable<int, int, std::allocator<int>, std::__detail::_Identity, std::equal_to<int>, std::hash<int>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, true, true> >::_M_find_before_node(unsigned long, int const&, unsigned long) const /usr/bin/../lib/gcc/x86_64-linux-gnu/6.2.0/../../../../include/c++/6.2.0/bits/hashtable.h:1412
    #1 0x889303 in std::_Hashtable<int, int, std::allocator<int>, std::__detail::_Identity, std::equal_to<int>, std::hash<int>, std::__detail::_Mod_range_hashing, std::__detail::_Default_ranged_hash, std::__detail::_Prime_rehash_policy, std::__detail::_Hashtable_traits<false, true, true> >::_M_find_node(unsigned long, int const&, unsigned long) const /usr/bin/../lib/gcc/x86_64-linux-gnu/6.2.0/../../../../include/c++/6.2.0/bits/hashtable.h:634:28
[...]

PR #424, opened 2017-04-27
Account for chains with cycles

apanda:
My constraint checking code assumed there were no cycles, this is not
true for cases such as the NAT. Addresses issue #421

### DIFF
--- a/core/module.cc
+++ b/core/module.cc
@@ -246,12 +246,12 @@ void Module::DeregisterAllAttributes() {
 }
 
 placement_constraint Module::ComputePlacementConstraints(
-    std::unordered_set<const Module *> &visited) const {
+    std::unordered_set<const Module *> *visited) const {
   // Take the constraints we have.
   int constraint = node_constraints_;
   // Follow the rest of the tree unless we have already been here.
-  if (visited.find(this) == visited.end()) {
-    visited.insert(this);
+  if (visited->find(this) == visited->end()) {
+    visited->insert(this);
     for (size_t i = 0; i < ogates_.size(); i++) {
       if (ogates_[i]) {
         auto next = static_cast<Module *>(ogates_[i]->arg());
@@ -268,13 +268,15 @@ placement_constraint Module::ComputePlacementConstraints(
   return constraint;
 }
 
-void Module::AddActiveWorker(int wid, ModuleTask *t) {
-  uint64_t worker_mask = 1ull << wid;
-  if (!(worker_mask & active_workers_)) {  // Already accounted for this worker
-    active_workers_ |= worker_mask;
+void Module::AddActiveWorker(int wid, const ModuleTask *t) {
+  if (!(active_workers_[wid])) {  // Have not already accounted for
+                                  // worker.
+    active_workers_[wid] = true;
+    // Check if we should propagate downstream. We propagate if either
+    // `propagate_workers_` is true or if the current module created the task.
     bool propagate = propagate_workers_;
-    if (!propagate) {  // Do not propagate unless this module created worker.
-      for (const auto task : tasks_) {  // Check to see if this is the case.
+    if (!propagate) {
+      for (const auto task : tasks_) {
         if (t == task) {
           propagate = true;
           break;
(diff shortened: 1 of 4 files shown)


core/module.cc
```diff
@@ -245,57 +245,77 @@ void Module::DeregisterAllAttributes() {
...
-void Module::AddActiveWorker(int wid) {
-  active_workers_.insert(wid);
-  if (propagate_workers_) {
-    for (auto ogate : ogates_) {
-      if (ogate) {
-        auto next = static_cast<Module *>(ogate->arg());
-        next->AddActiveWorker(wid);
+void Module::AddActiveWorker(int wid, ModuleTask *t) {
+  uint64_t worker_mask = 1ull << wid;
+  if (!(worker_mask & active_workers_)) {  // Already accounted for this worker
```
barath: The comment reads as logically inverted, I think.  The if statement is when we have *not* accounted for this worker, right?

core/module.cc
```diff
@@ -245,57 +245,77 @@ void Module::DeregisterAllAttributes() {
...
+    active_workers_ |= worker_mask;
+    bool propagate = propagate_workers_;
+    if (!propagate) {  // Do not propagate unless this module created worker.
+      for (const auto task : tasks_) {  // Check to see if this is the case.
```
barath: These two comments are a bit confusingly worded.

core/module.cc
```diff
@@ -245,57 +245,77 @@ void Module::DeregisterAllAttributes() {
...
-  if (propagate_workers_) {
-    for (auto ogate : ogates_) {
-      if (ogate) {
-        auto next = static_cast<Module *>(ogate->arg());
-        next->AddActiveWorker(wid);
+void Module::AddActiveWorker(int wid, ModuleTask *t) {
+  uint64_t worker_mask = 1ull << wid;
+  if (!(worker_mask & active_workers_)) {  // Already accounted for this worker
+    active_workers_ |= worker_mask;
+    bool propagate = propagate_workers_;
```
barath: Is propagate_workers_ going to change below?  If not, I think bool propagate might not be needed.

core/module.cc
```diff
@@ -245,57 +245,77 @@ void Module::DeregisterAllAttributes() {
...
 
-void Module::AddActiveWorker(int wid) {
-  active_workers_.insert(wid);
-  if (propagate_workers_) {
-    for (auto ogate : ogates_) {
-      if (ogate) {
-        auto next = static_cast<Module *>(ogate->arg());
-        next->AddActiveWorker(wid);
+void Module::AddActiveWorker(int wid, ModuleTask *t) {
+  uint64_t worker_mask = 1ull << wid;
```
sangjinhan: Can we use `vector<bool>` instead of `uint64_t` for active workers and masks? I am a little bit afraid that it makes an implicit assumption that the maximum number of workers is always <= 64.

core/module.h
```diff
@@ -291,18 +291,28 @@ class Module {
    * Compute placement constraints based on the current module and all
    * downstream modules (i.e., modules connected to out ports.
    */
-  placement_constraint ComputePlacementConstraints() const;
+  placement_constraint ComputePlacementConstraints(
+      std::unordered_set<const Module *> &visited) const;
```
sangjinhan: Google style guide requires using a pointer for mutable arguments, and const reference for immutable arguments.

core/module.cc
apanda: No -- but we choose to propagate either when propagate_workers_ is true or when the task is created by this worker, so propagate is just the logical conjunction of those cases. I can restructure the code if you want but the values are different right now.

core/module.cc
apanda: I had it as unordered_set<int> before, @barath had suggested this move, am happy to go back if preferable.

barath: I think moving back to undordered_set is preferable to vector<bool>.

sangjinhan: I might be missing some context here, but why is unordered_set a better fit than vector<bool>? vector<bool> is more efficient then in terms of both time and space. If we are concerned with dynamic memory allocation, we can choose static-sized std::bitset instead.

barath: No big reason.  I guess in the back of my mind, if we were trying to be very general, then we would allow for an arbitrary number of workers, in which case we would benefit from a set for which we don't need to know the max value in advance.  But if we assume we know how many workers are possible (e.g. max workers) then vector<bool> is good.

apanda:
@barath, @sangjinhan your comments should be addressed at this point.

barath:
photo on 2009-09-24 at 21 26 3

PR #425, opened 2017-04-28
Remove print statement confirming no-violations.

justinemarie:
The new safety checking features print out "No violations" every time one runs a command in bessctl. This gets tedious -- it's likely easier to have the code only print if violations are found, and just silently succeed when there are no problems.


sangjinhan:
tumblr_m7fehvxjft1r6spllo1_500

PR #426, opened 2017-04-28
git: Fix a number of issues in hooks

changlan:
- Fix pre-commit hook
- Make sure pwd is the root directory of BESS repo
- Make sure clang-format and autopep8 exist before installing hooks

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -40,12 +40,12 @@ refresh_commit() {
 }
 
 if ! command_exists $CXX_FORMAT; then
-	@echo "Error: $CXX_FORMAT executable not found.\n" >&2
+	echo "Error: $CXX_FORMAT executable not found.\n" >&2
 	exit 1
 fi
 
 if ! command_exists $PY_FORMAT; then
-	@echo "Error: $PY_FORMAT executable not found.\n" >&2
+	echo "Error: $PY_FORMAT executable not found. Try 'pip install $PY_FORMAT'?\n" >&2
 	exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


.hooks/pre-commit
```diff
@@ -40,12 +40,12 @@ refresh_commit() {
...
 
 if ! command_exists $CXX_FORMAT; then
-	@echo "Error: $CXX_FORMAT executable not found.\n" >&2
+	echo "Error: $CXX_FORMAT executable not found.\n" >&2
 	exit 1
 fi
 
 if ! command_exists $PY_FORMAT; then
-	@echo "Error: $PY_FORMAT executable not found.\n" >&2
+	echo "Error: $PY_FORMAT executable not found. Try 'pip install $PY_FORMAT'?\n" >&2
```
melvinw: Could you add this to the ansible script?

changlan: Good point

melvinw: Though, it may make more sense as a separate PR. Up to you

env/packages.yml
```diff
@@ -46,8 +46,11 @@
       when: ansible_kernel | match(".*-generic")
 
     - name: Install list of Python packages
-      pip: name=grpcio
+      pip: name={{item}}
       become: true
+      with_items:
```
sangjinhan: I am concerned with `with_items` (https://github.com/NetSys/bess/pull/361). Instead, How about using `apt` (`python-pep8` and `python-autopep8`)?

Note that `python-autopep8` package alone is not enough; its dependency package `pep8` would only support Python 3, so `python-pep8` should be also installed.

changlan: Sounds good. Updated.

sangjinhan:
tumblr_m7nm24ecr21qgjltdo1_1280

PR #427, opened 2017-04-30
Minor updates to git pre-commit hook

sangjinhan:
* Fix errors caused by `@echo`
* Formatters are now customizable (e.g., `clang-format-3.7` instead of `clang-format`)
* Detect corner cases that are not correctly handled by the script:
  * Bypassed staging with `git commit -a` or `git commit -- <files>`
  * Partial staging with `git add --patch`
* Show reformatting differences in color

Will need to rebase this PR once #426 is merged, since there are some overlaps.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -44,6 +44,8 @@ get_patch() {
 }
 
 apply_and_commit() {
+	# The "git add" we use below do not work with non-default index file,
+	# which is used by "git commit -a" or "git commit <files>"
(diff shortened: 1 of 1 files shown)


sangjinhan:
It also fixes the bug that reformat-and-commit doesn't work when two or more files are reformatted.

sangjinhan:
Rebased to resolve merge conflicts with #426

changlan:
image

PR #428, opened 2017-04-30
Fix bessctl auto completion with special chars

sangjinhan:
`readline` treats some special characters (including backtick and `~!@#$%^&*()-=+[{]}\|;:'",<>?/` as delimiters. This behavior interferes with `bessctl` causing incorrect auto completions. Examples:

```
localhost:10514 $ track disable *<tab>
localhost:10514 $ track disable ** _
...
localhost:10514 $ run file ~/my-modu<tab>
localhost:10514 $ run file ~/my-my-modules/_
```

This PR fixes the issue by declaring such characters non-delimiters.

### DIFF
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -215,12 +215,6 @@ class CLI(object):
         candidates = []
         num_full_matches = 0
 
-        # ignore partial_word set by readline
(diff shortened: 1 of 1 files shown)


changlan:
image

PR #430, opened 2017-05-01
trie: Add tests for copy constructor and assignment operator

changlan:
- Add tests for copy constructor and assignment operator;
- Fix implicit assignment operator;
- Use `std::unique_ptr`.

### DIFF
--- a/core/utils/trie.h
+++ b/core/utils/trie.h
@@ -12,31 +12,29 @@ class Trie {
  public:
   // Node definition
   struct Node {
-    Node() : leaf(), children() {
-      for (int i = 0; i < 256; i++) {
-        children[i] = nullptr;
-      }
-    }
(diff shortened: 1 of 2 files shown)


sangjinhan:
Does it have any performance implications?

changlan:
`std::unique_ptr`'s memory footprint is same as regular pointers, and its constructor and destructor should have the same performance as manual `new` and `delete`.

sangjinhan:
I was concerned about runtime overheads (for pointer dereference). It seems that there is none: http://stackoverflow.com/questions/22295665/how-much-is-the-overhead-of-smart-pointers-compared-to-normal-pointers-in-c

Thanks!

sangjinhan:
jjkybd3nsaqkbdxdy4fytymi_500

PR #432, opened 2017-05-01
Revert "module: Optimize RunModuleSplit()."

ddiproietto:
This reverts commit 05af48881575d7b81bd45956c00fc4ce6d3b6604.

It causes crashes in travis tests, but not locally.  Until we debug
it let's revert the patch.

### DIFF
--- a/core/module.cc
+++ b/core/module.cc
@@ -486,33 +486,32 @@ void Module::RunSplit(const gate_idx_t *out_gates,
   gate_idx_t pending[bess::PacketBatch::kMaxBurst];
   bess::PacketBatch batches[bess::PacketBatch::kMaxBurst];
 
-  bess::PacketBatch **splits = ctx.splits();
+  bess::PacketBatch *splits = ctx.splits();
 
-  // phase 1: collect unique ogates into pending[] and add packets to local
-  // batches, using splits to remember the association between an ogate and a
-  // local batch
+  /* phase 1: collect unique ogates into pending[] */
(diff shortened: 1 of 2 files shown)


justinemarie:
Update: Daniele and I just talked about this. We're not sure what is causing these (non-deterministic) errors but it was something around when this commit happened, possibly in combination with this commit -- but it clearly isn't this commit alone.

I'm going to check out a bunch of old versions of the repo and run them through Travis repeatedly to see whether we can track it down.

justinemarie:
This is the commit one prior to when the PR was merged:
https://travis-ci.org/NetSys/bess/builds/227782156

justinemarie:
This is the commit one before when the PR was branched:
https://travis-ci.org/NetSys/bess/builds/227783582

justinemarie:
So the alternative two paths succeeded -- I guess we should go ahead and revert here and return to this optimization later?

Note that this revert commit originally failed as well, so reverting is not guaranteed to fix everything 100%...

barath:
I'm curious if we understand why the optimization failed.

justinemarie:
We don't, we've been trying to track it down -- but the problem is the bug is backing up all the other PRs because Travis is failing all of them. So Daniele proposed we revert and come back to it.

barath:
Thanks.  Agreed, just was curious.

justinemarie:
Daniele and I are doing some git surgery.

justinemarie:
image

PR #435, opened 2017-05-02
driver and kernel module support for 32 bit user and 64 bit kernel

nrdmao:
These changes were required to get bessd to function in 32 bit mode as a user process against a 64 bit mode kernel.

### DIFF
--- a/core/drivers/vport.cc
+++ b/core/drivers/vport.cc
@@ -33,10 +33,6 @@
 
 #define ROUND_TO_64(x) ((x + 32) & (~0x3f))
 
-/* We cannot directly use phys_addr_t on 32bit machines,
- * since it may be sizeof(phys_addr_t) != sizeof(void *) */
-typedef uintptr_t paddr_t;
-
 static inline int find_next_nonworker_cpu(int cpu) {
   do {
     cpu = (cpu + 1) % sysconf(_SC_NPROCESSORS_ONLN);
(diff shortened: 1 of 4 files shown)


core/drivers/vport.cc
```diff
@@ -573,10 +579,12 @@ int VPort::RecvPackets(queue_t qid, bess::Packet **pkts, int max_cnt) {
 int VPort::SendPackets(queue_t qid, bess::Packet **pkts, int cnt) {
   struct queue *rx_queue = &out_qs_[qid];
 
-  paddr_t paddr[bess::PacketBatch::kMaxBurst];
+  phys_addr_t paddr[bess::PacketBatch::kMaxBurst];
 
   int ret;
 
+  assert(cnt <= bess::PacketBatch::kMaxBurst);
```
sangjinhan: It causes a build error only when `DEBUG=1`: https://travis-ci.org/NetSys/bess/jobs/228102442

```
In file included from /usr/include/c++/5/cassert:43:0,
                 from drivers/../packet.h:9,
                 from drivers/../module.h:13,
                 from drivers/../port.h:14,
                 from drivers/vport.h:5,
                 from drivers/vport.cc:1:
drivers/vport.cc: In member function 'virtual int VPort::SendPackets(queue_t, bess::Packet**, int)':
drivers/vport.cc:586:14: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
   assert(cnt <= bess::PacketBatch::kMaxBurst);
```

Perhaps an easy fix could be `static_cast<unsigned>(cnt) <= ...`, which is somewhat handy as it can detect `cnt < 0` as well. Or alternatively, you can try `CHECK_LE()` (or `DCHECK_LE()` might be better), although I am not sure if that macro would also causes the same warning.

nrdmao: OK, will static_cast<size_t>(cnt) as kMaxBurst of that type.

sangjinhan:
tumblr_m07rxzwbhx1qzf9gso1_500

PR #436, opened 2017-05-03
rewrite: Honor round robin behavior.

ddiproietto:
The Rewrite module is not really round robin (e.g., by looking at the
`show pipeline` numbers for `samples/exactmatch`, we can notice a
slight difference between gates that are supposed to have the same
rate).

This commit fixes the problem.

### DIFF
--- a/core/modules/rewrite.cc
+++ b/core/modules/rewrite.cc
@@ -107,7 +107,7 @@ inline void Rewrite::DoRewrite(bess::PacketBatch *batch) {
 
   next_turn_ = start + cnt;
   if (next_turn_ >= bess::PacketBatch::kMaxBurst) {
-    next_turn_ -= bess::PacketBatch::kMaxBurst;
+    next_turn_ = next_turn_ % num_templates_;
   }
 }
 


core/modules/rewrite.cc
```diff
@@ -107,7 +107,7 @@ inline void Rewrite::DoRewrite(bess::PacketBatch *batch) {
 
   next_turn_ = start + cnt;
   if (next_turn_ >= bess::PacketBatch::kMaxBurst) {
-    next_turn_ -= bess::PacketBatch::kMaxBurst;
+    next_turn_ = next_turn_ % num_templates_;
```
sangjinhan: The hack of using `kNumSlots` was to avoid integer modulo instruction, which used to appear significant overhead when profiled. Perhaps we can achieve the fairness without using `%`? (e.g. lookup table)

ddiproietto: I seem, I had no idea, thanks for pointing this out.

Do you think it is expensive, even if it's done once per batch?

I can try a lookup table and benchmark the two solution

Alternatively, we can leave the code as it is

sangjinhan: If I remember correctly it added about 1-2 cycles per "packet", although I expect the performance impact will highly depend on what modules are used downstream (i.e., whether those modules use any integer division/modulo instructions).

Issue #434, opened 2017-05-02
Codecov will not die.


justinemarie:
I cannot seem to get rid of codecov despite trying many times. Who configured it in the first place? Is there some checkbox I am missing?

changlan:
Close this since we are fixing the codecov issue

PR #440, opened 2017-05-05
Stabilizes code coverage data

changlan:
- If `DEBUG=1`, run unit tests only
- Use gcov-5
- Remove `lcov` on CI since it seems to be useless

This PR should address #434 and #388.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,5 +65,6 @@ core/*.gcda
 core/*.gcno
 core/*/*.gcda
 core/*/*.gcno
+*.gcov
 coverage.info
 coverage_html
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ services:
 
 env:
   global:
-    - GCOV_PREFIX_STRIP=2
     - ASAN_SYMBOLIZER_PATH=/usr/lib/llvm-3.8/bin/llvm-symbolizer
     - ASAN_OPTIONS=log_path=/tmp/saniziter,log_exe_name=1
     - LSAN_OPTIONS=suppressions=$TRAVIS_BUILD_DIR/core/lsan.suppress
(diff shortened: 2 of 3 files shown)


sangjinhan:
Could you briefly describe here the changes and why they are needed, for future reference, particularly for the followings?
* Installing `g++-5`: is it for `gcov-5`? If so, could you add a comment in the travis script?
* Why left only the C++ unit tests? Because other tests do not accumulate the results correctly? Because others require different `GCOV_PREFIX_STRIP` values?

I believe the use of `lcov` was for tests (#395) and it is safe for removal.

changlan:
- Yes. I added the comment.
- Those runtime integration tests have some sort of nondeterminism that leads to fluctuations of code coverage data we saw. I remove them and only include unit tests.

justinemarie:
Chang you are the king.

sangjinhan:
Even if they are non-deterministic to some extent, they should have been cumulative; i.e., non-unittest should only increase the coverage from the baseline (28% from unit tests). But we have seen the coverage number going far below 28% many times. This may suggest that the real problem is "the result from later tests did not accumulate correctly", not "later tests had non-determinism"? Could you check on this...?

My concern is that `DEBUG=1` used to have served two roles: 1. running the code in debug mode and 2. collecting coverage data. The new change is for the role 2 but nullifying the role 1. One option might be separating `COVERAGE=1` from `DEBUG=1`, this is what I would do as a last resort...

Perhaps the ideal case would be, when `DEBUG=1`, 
1. All tests run, no matter whether they are unit or non-unit tests,
1. All tests generate coverage data in a cumulative manner, and
1. The final resulting coverage is reasonably stable.

changlan:
Sorry, I guess I wasn't clear. The baseline (unit tests) wasn't accumulated correctly since `GCOV_PREFIX_STRIP` was incorrect. Therefore the old coverage result was actually only reflecting the later tests, and those tests have non-determinism. That's probably why the data is fluctuating. So my initial intention is that the coverage should reflect the unit tests. 

You are right about the two roles. I am looking into on how to stabilize the coverage of other tests. If that turns out to be nontrivial, perhaps the option I will go is:

- Separate `DEBUG` and `COVERAGE`
- If `COVERAGE=1`, run unit tests only.

changlan:
Done. Let's see if the data is stable.

sangjinhan:
jjkybd3nsbyvm3jol2gbrofl_500

PR #441, opened 2017-05-08
Fix errors in DPDK custom configuration file

sangjinhan:
This PR fixes the issue that Mellanox PMD drivers are not built even when it should, which is caused by the following the problems:

* CONFIG_RTE_... should have a value `y` or `n`, not 'Y' or 'N'.
* Mellanox drivers should be enabled by default
  * It's supposed to be "disabled" when dependency is not available.
  * (not that it is enabled when dependency is available)


melvinw:
Merged, thanks! 
rp7zm8ri1tty

PR #444, opened 2017-05-09
Modularizes the scheduler.

barath:
The current scheduler ("default") is made the default.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -256,7 +256,7 @@ def get_var_attrs(cli, var_token, partial_word):
         elif var_token == '[SCHEDULER]':
             var_type = 'name'
             var_desc = 'specify the type of scheduler'
-            var_candidates = ['fast']
+            var_candidates = ['default']
 
         elif var_token == 'PORT':
             var_type = 'name'
@@ -769,11 +769,7 @@ def run_file(cli, conf_file, env_map):
 
 @cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
 def add_worker(cli, wid, core, scheduler=None):
-    s = scheduler
-    if s is None:
-        s = 'fast'
-    cli.bess.add_worker(wid, core, s)
-
+    cli.bess.add_worker(wid, core, scheduler or '')
 
 @cmd('add port DRIVER [NEW_PORT] [PORT_ARGS...]', 'Add a new port')
 def add_port(cli, driver, port, args):
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -253,6 +253,11 @@ def get_var_attrs(cli, var_token, partial_word):
             var_type = 'name'
             var_desc = 'specify a name of the new port'
 
+        elif var_token == '[SCHEDULER]':
+            var_type = 'name'
+            var_desc = 'specify the type of scheduler'
+            var_candidates = ['fast']
+
         elif var_token == 'PORT':
             var_type = 'name'
             var_desc = 'name of a port'
@@ -763,8 +768,11 @@ def run_file(cli, conf_file, env_map):
 
 
 @cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
-def add_worker(cli, wid, core, scheduler='fast'):
-    cli.bess.add_worker(wid, core, scheduler)
+def add_worker(cli, wid, core, scheduler=None):
+    s = scheduler
+    if s is None:
+        s = 'fast'
+    cli.bess.add_worker(wid, core, s)
 
 
 @cmd('add port DRIVER [NEW_PORT] [PORT_ARGS...]', 'Add a new port')
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -762,9 +762,9 @@ def run_file(cli, conf_file, env_map):
     _run_file(cli, os.path.expanduser(conf_file), env_map)
 
 
-@cmd('add worker WORKER_ID CORE', 'Create a worker')
-def add_worker(cli, wid, core):
-    cli.bess.add_worker(wid, core)
+@cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
+def add_worker(cli, wid, core, scheduler='fast'):
+    cli.bess.add_worker(wid, core, scheduler)
 
 
 @cmd('add port DRIVER [NEW_PORT] [PORT_ARGS...]', 'Add a new port')
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -768,7 +768,7 @@ def run_file(cli, conf_file, env_map):
 
 
 @cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
-def add_worker(cli, wid, core, scheduler=None):
+def add_worker(cli, wid, core, scheduler):
     cli.bess.add_worker(wid, core, scheduler or '')
 
 @cmd('add port DRIVER [NEW_PORT] [PORT_ARGS...]', 'Add a new port')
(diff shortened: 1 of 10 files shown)


bessctl/commands.py
```diff
@@ -762,9 +762,9 @@ def run_file(cli, conf_file, env_map):
     _run_file(cli, os.path.expanduser(conf_file), env_map)
 
 
-@cmd('add worker WORKER_ID CORE', 'Create a worker')
-def add_worker(cli, wid, core):
-    cli.bess.add_worker(wid, core)
+@cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
```
ddiproietto: I think the "[SCHEDULER]" keyword needs to be defined, otherwise the command interpreter expects a literal "[SCHEDULER].  I had to add this:

```
diff --git a/bessctl/commands.py b/bessctl/commands.py
index 382b7a7..59701d2 100644
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -253,6 +253,11 @@ def get_var_attrs(cli, var_token, partial_word):
             var_type = 'name'
             var_desc = 'specify a name of the new port'
 
+        elif var_token == '[SCHEDULER]':
+            var_type = 'name'
+            var_desc = 'specify the type of scheduler'
+            var_candidates = ['fast']
+
         elif var_token == 'PORT':
             var_type = 'name'
             var_desc = 'name of a port'
```

bessctl/commands.py
```diff
@@ -762,9 +762,9 @@ def run_file(cli, conf_file, env_map):
...
+def add_worker(cli, wid, core, scheduler='fast'):
```
ddiproietto: For a reason that I don't understand (I'm not great in python), specifying a default value like this doesn't work.

All the other functions here use a different pattern: default value=None and checking for None in the body of the function.  That seems to work

core/scheduler.h
```diff
@@ -4,6 +4,7 @@
 #include <iostream>
 #include <queue>
 #include <sstream>
+#include <stack>
```
ddiproietto: Not a big deal, is this used here?

bessctl/commands.py
barath: Thanks -- good catch.

bessctl/commands.py
barath: Fixed.

core/scheduler.h
barath: Removed.

bessctl/commands.py
melvinw: testing something.

core/scheduler.h
```diff
@@ -260,9 +181,107 @@ class Scheduler final {
...
+          }
+        }
+      }
+
+      ScheduleOnce();
+    }
+  }
+
+  // Runs the scheduler once.
+  void ScheduleOnce() __attribute__((always_inline)) {
```
sangjinhan: We don't need `__attribute__((always_inline))` here; the attribute is mostly irrelevant in C++ context. As long as a member function is defined in the class body it will be inlined according to the spec.

core/scheduler.h
```diff
@@ -260,9 +181,107 @@ class Scheduler final {
...
+      now = rdtsc();
+      this->stats_.cycles_idle += (now - this->checkpoint_);
+    }
+
+    this->checkpoint_ = now;
+  }
+
+  // Selects the next TrafficClass to run.
+  LeafTrafficClass<CallableTask> *Next(uint64_t tsc)
+      __attribute__((always_inline)) {
```
sangjinhan: same

core/worker.cc
```diff
@@ -281,8 +288,14 @@ void *run_worker(void *_arg) {
   return ctx.Run(_arg);
 }
 
-void launch_worker(int wid, int core) {
-  struct thread_arg arg = {.wid = wid, .core = core};
+void launch_worker(int wid, int core, [[maybe_unused]] const std::string &scheduler) {
+  SchedulerType scheduler_type = SchedulerType::FAST;
```
sangjinhan: Can we check `scheduler` here (even in a hardcoded way), rather than deferring error check until `Worker::Run()`?
```
if (scheduler == "fast") {
  arg.scheduler = new FastScheduler<Task>();
} else {
  CHECK(0);
}
```

My concerns are: 1. a new type `SchedulerType` is introduced without much necessity, 2. the argument "scheduler" here is not being used, 3. `Worker::Run()` is too late to handle error gracefully (notifying the client that invalid scheduler is specified).

core/worker.cc
```diff
@@ -298,7 +311,7 @@ void launch_worker(int wid, int core) {
 Worker *get_next_active_worker() {
   static int prev_wid = 0;
   if (num_workers == 0) {
-    launch_worker(0, FLAGS_c);
+    launch_worker(0, FLAGS_c, "fast");
```
sangjinhan: Perhaps we can make "fast" as a default value for the third function parameter, and not specify it here..?

core/bessctl.cc
```diff
@@ -1515,7 +1520,7 @@ class BESSControlImpl final : public BESSControl::Service {
       if ((wid != -1 && !is_worker_active(wid)) ||
           (wid == -1 && num_workers == 0)) {
         if (num_workers == 0 && (wid == 0 || wid == -1)) {
-          launch_worker(0, FLAGS_c);
+          launch_worker(0, FLAGS_c, "fast");
```
sangjinhan: Perhaps we can make "fast" as a default value for the third function parameter, and not specify it here..? At the moment all callers of `launch_worker()` must be updated whenever a default scheduler is switched.

bessctl/commands.py
```diff
@@ -762,9 +767,12 @@ def run_file(cli, conf_file, env_map):
     _run_file(cli, os.path.expanduser(conf_file), env_map)
 
 
-@cmd('add worker WORKER_ID CORE', 'Create a worker')
-def add_worker(cli, wid, core):
-    cli.bess.add_worker(wid, core)
+@cmd('add worker WORKER_ID CORE [SCHEDULER]', 'Create a worker')
+def add_worker(cli, wid, core, scheduler=None):
```
sangjinhan: Or even better, the Python code should not bother with what the default scheduler is (bessd can choose one if a scheduler is not specified). Currently too many files are coupled with what schedulers are there and what the default is.

protobuf/bess_msg.proto
```diff
@@ -62,8 +62,9 @@ message ListWorkersResponse {
 }
 
 message AddWorkerRequest {
-  int64 wid = 1;   /// Worker ID to be added
-  int64 core = 2;  /// CPU core ID on which the worker would run
+  int64 wid = 1;         /// Worker ID to be added
+  int64 core = 2;        /// CPU core ID on which the worker would run
+  string scheduler = 3;  /// Default is "fast".
```
sangjinhan: Mistake here. Otherwise LGTM.

barath: Thanks, fixed.

core/scheduler.h
```diff
@@ -18,6 +18,10 @@ struct sched_stats {
   uint64_t cycles_idle;
 };
 
+enum class SchedulerType {
```
sangjinhan: One final nit: now we can safely remove this new type since it is nowhere used.

core/worker.cc
```diff
@@ -22,6 +22,8 @@
 #include "utils/time.h"
 
 using bess::Scheduler;
+using bess::DefaultScheduler;
+using bess::SchedulerType;
```
sangjinhan: not required any longer

core/scheduler.h
barath: Agreed, updated.

core/worker.cc
barath: Updated.

sangjinhan:
tumblr_m8dxt0yaue1qmbg8bo1_500

PR #446, opened 2017-05-10
Fix `WorkerSplit` and `Bypass` model

apanda:
Both modules are thread safe and `worker_split.bess` uses them as such.

### DIFF
--- a/core/modules/bypass.h
+++ b/core/modules/bypass.h
@@ -11,6 +11,7 @@ class Bypass final : public Module {
   static const gate_idx_t kNumIGates = MAX_GATES;
   static const gate_idx_t kNumOGates = MAX_GATES;
 
+  Bypass() { max_allowed_workers_ = Worker::kMaxWorkers; }
   void ProcessBatch(bess::PacketBatch *batch) override;
 };
 
--- a/core/modules/worker_split.h
+++ b/core/modules/worker_split.h
@@ -7,6 +7,8 @@ class WorkerSplit final : public Module {
  public:
   static const gate_idx_t kNumOGates = Worker::kMaxWorkers;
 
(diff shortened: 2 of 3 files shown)


core/modules/bypass.h
```diff
@@ -11,6 +11,7 @@ class Bypass final : public Module {
   static const gate_idx_t kNumIGates = MAX_GATES;
   static const gate_idx_t kNumOGates = MAX_GATES;
 
+  Bypass() { max_allowed_workers_ = Worker::kMaxWorkers; }
```
barath: Move to initializer list?

core/modules/worker_split.h
```diff
@@ -7,6 +7,8 @@ class WorkerSplit final : public Module {
  public:
   static const gate_idx_t kNumOGates = Worker::kMaxWorkers;
 
+  WorkerSplit() { max_allowed_workers_ = kNumOGates; }
```
barath: Move to initializer list?

core/modules/bypass.h
apanda: It is a variable declared in the base class, and cannot be initialized in the initializer list of a derived class, see for [example](http://stackoverflow.com/questions/7405740/how-can-i-initialize-base-class-member-variables-in-derived-class-constructor).

core/modules/worker_split.h
apanda: It is a variable declared in the base class, and cannot be initialized in the initializer list of a derived class, see for [example](http://stackoverflow.com/questions/7405740/how-can-i-initialize-base-class-member-variables-in-derived-class-constructor).

barath:
Ah, good point -- I didn't notice that.  LGTM.

barath:
img_0932

PR #447, opened 2017-05-10
port: Virtualize Port and delete dead codes

changlan:
`Init()`, `DeInit()`, `RecvPackets()`, `SendPackets()` should always be defined
when creating a new driver. To make the process less error-prone, we can enforce
this by making those functions purely virtual.


barath:
LGTM.  Do all the other drivers properly override methods from Port?

changlan:
Yes

barath:
kona zen card

PR #448, opened 2017-05-14
Fix false style error detection

sangjinhan:
When there are multiple files to commit, LF is used as a separator between diff results. It causes the pre-commit hook to (incorrectly) recognize the LF characters that there are something requiring reformatting.


changlan:
image

PR #449, opened 2017-05-14
Revamp endianness and protocol code

sangjinhan:
This PR updates the code to use `BigEndian` types (`be16_t`, `be32_t`, and `be64_t`) whenever applicable. Since these are strongly typed, mistakes regarding endianness can be statically detected. All protocol header declaration now use `BigEndian` types for multi-byte fields. For example, the following code will cause build errors.
```
  Tcp *tcp;  // seq_num and ack_num are declared of be32_t type
  tcp->ack_num = tcp_received->seq_num + 1;  // "+" between be32_t and uint32_t is not defined
```

Instead, the code should be explicit about typing, like either of these.
```
tcp->ack_num = tcp_received->seq_num + be32_t(1);  // one way
tcp->ack_num = be32_t(tcp_received->seq_num.value() + 1);  // another way
```

The idea is from the Linux kernel source code, where developers are required to explicitly convert between cpu-order and network-order variables (because of C limitations, gcc cannot detect errors directly; Linux requires external verifier for this).

* Update existing code using DPDK or libc functions to use `utils/*.h`for strict type checking.
* VXLAN header was added
* Some structs were renamed for clarity and brevity:
  * `EthHeader` -> `Ethernet`
  * `Ipv4Header` -> `Ipv4`
  * `TcpHeader` -> `Tcp`
  * `UdpHeader` -> `Udp`
  * `IcmpHeader` -> `Icmp` 
  * `CIDRNetwork` -> `Ipv4Prefix`

### DIFF
--- a/core/modules/update.cc
+++ b/core/modules/update.cc
@@ -1,6 +1,6 @@
 #include "update.h"
+
 #include "../utils/endian.h"
-#include <rte_byteorder.h>
 
 const Commands Update::cmds = {
     {"add", "UpdateArg", MODULE_CMD_FUNC(&Update::CommandAdd), 0},
@@ -14,18 +14,18 @@ CommandResponse Update::Init(const bess::pb::UpdateArg &arg) {
 void Update::ProcessBatch(bess::PacketBatch *batch) {
   int cnt = batch->cnt();
 
-  for (int i = 0; i < num_fields_; i++) {
+  for (size_t i = 0; i < num_fields_; i++) {
     const auto field = &fields_[i];
 
-    uint64_t mask = field->mask;
-    uint64_t value = field->value;
+    be64_t mask = field->mask;
+    be64_t value = field->value;
     int16_t offset = field->offset;  // could be < 0
 
     for (int j = 0; j < cnt; j++) {
       bess::Packet *snb = batch->pkts()[j];
       char *head = snb->head_data<char *>();
 
-      uint64_t *p = reinterpret_cast<uint64_t *>(head + offset);
+      be64_t *p = reinterpret_cast<be64_t *>(head + offset);
       *p = (*p & mask) | value;
     }
   }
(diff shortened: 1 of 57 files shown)


core/modules/update.cc
```diff
@@ -34,48 +34,37 @@ void Update::ProcessBatch(bess::PacketBatch *batch) {
...
-
-    if (offset < 0) {
-      return CommandFailure(EINVAL, "too small 'offset'");
+    if (field.offset() + size > SNBUF_DATA) {
+      return CommandFailure(EINVAL, "too large 'offset'");
     }
 
-    offset -= (8 - size);
-    mask = (1ull << ((8 - size) * 8)) - 1;
+    be64_t value(field.value() << ((8 - size) * 8));
```
barath: Is there a unit test that covers this code path?  I ask because this change is slightly more than just big endian tweaks.

core/modules/url_filter.h
```diff
@@ -17,44 +17,44 @@
...
+    return memcmp(this, &other, sizeof(*this)) == 0;
   }
 };
 
+static_assert(sizeof(Flow) == 16, "Flow must be 16 bytes.");
+
 // Hash function for std::unordered_map
 struct FlowHash {
   std::size_t operator()(const Flow &f) const {
-    static_assert(sizeof(Flow) == 2 * sizeof(uint64_t),
```
barath: Don't we want to keep the static assert to ensure that the flow struct is exactly 16 bytes?

core/utils/endian_test.cc
```diff
@@ -0,0 +1,156 @@
+#include "endian.h"
+
+#include <gtest/gtest.h>
+
+using namespace bess::utils;
+
+namespace {
+
+TEST(EndianTest, Creation) {
```
barath: Is there any way to test the opposite endianness than the current system?

core/modules/url_filter.h
sangjinhan: The static assert is still there, but outside the struct.

core/modules/update.cc
sangjinhan: `samples/update.bess` covers both `Update` and `RandomUpdate` (which has been updated similarly).

core/modules/url_filter.h
barath: Ah, great -- I missed that.

barath:
photo on 2011-08-12 at 21 26 2

PR #450, opened 2017-05-14
Skip ZeroCopyVPort unittests when run by a non-root user

sangjinhan:
The unittests for ZeroCopyVPort silently fails if not run by root during DPDK initialization. This commit will just skip (with a log message) the test in that case.


barath:
img_0970

PR #452, opened 2017-05-15
Enhance versioning

sangjinhan:
`version.h` is updated whenever `Makefile` is executed. This confuses `make`, since it thinks something useful has been done:
```sh
$ make
...
$ make
$        # "make: Nothing to be done for 'all'." is not shown since version.h has been updated.
```

This PR fixes this behavior of `Makefile`, not updating `version.h` when unnecessary (e.g., modifying modules or unit tests). Also other minor updates include:
* Allow custom versioning w/o `git describe`
  * This enables building BESS when git is not available.
* Print out version info in log messages and `bessd -t`
* Use 7 characters for commit hashes


melvinw:
iqc2qop Thanks!

PR #457, opened 2017-05-16
Fix incorrect "gates" arguments in examples

sangjinhan:
`gates` takes a list of gate numbers, not the number of gates.


changlan:
image

PR #458, opened 2017-05-16
Fix name collision of pktgen on multiple workers

sangjinhan:
`perftest/pktgen.bess` fails when used with multiple workers. This is because every worker thread tries to create a tc named`bit_limit`.

```
$ BESS_PORTS=2 BESS_CORES=2 BESS_RATELIMIT_MBPS=10000 bessctl run perftest/pktgen -- monitor port
…
*** Error: Unhandled exception in the configuration script (most recent call last)
  File "/home/sangjin/bess/bessctl/conf/perftest/pktgen.bess", line 61, in <module>
    bess.add_tc('bit_limit', policy='rate_limit', resource='bit', limit={'bit': rate_limit * 1000000})
  File "/home/sangjin/bess/bessctl/../libbess-python/bess.py", line 464, in add_tc
    return self._request('AddTc', request)
  File "/home/sangjin/bess/bessctl/../libbess-python/bess.py", line 158, in _request
    raise self.Error(code, errmsg)
*** Error: Name 'bit_limit' already exists
  BESS daemon response - errno=22 (EINVAL: Invalid argument)
  Command failed: run perftest/pktgen
```

### DIFF
--- a/bessctl/conf/perftest/pktgen.bess
+++ b/bessctl/conf/perftest/pktgen.bess
@@ -3,7 +3,7 @@ import scapy.all as scapy
 pkt_size = int($BESS_PKT_SIZE!'60')
 num_ports = int($BESS_PORTS!'1')
 num_cores = int($BESS_CORES!'1')
-rate_limit = int($BESS_RATELIMIT_MBPS!'0') # default no limit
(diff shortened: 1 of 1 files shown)


melvinw:
LGTM. Thanks!
wsiytzm

PR #459, opened 2017-05-17
Makes TC blocking common to all TCs.

barath:


barath:
I originally added the inline / always_inline when the code was in the .cc
file, and it eventually migrated to .h where it should be inlined by
default but I never removed the inline keywords.

I tried to reason about whether there was any case where you could be
throttled or blocked at the same time, but I realized that (I think) rate
limit traffic classes are the only ones that could ever be both, and in
that case one of the two will happen first and dominate for that time --
either a rate limit TC will get blocked in which case it won't get
scheduled so can't get throttled, but that's fine, or it will get throttled
and then can't have its blocked status changed because its children can't
be scheduled until it is unthrottled, which is also fine.

core/traffic_class.h
```diff
@@ -256,6 +257,9 @@ class TrafficClass {
 
   struct tc_stats stats_;
 
+  // THe tsc time that this should be unblocked by the scheduler.
```
barath: Fixed.

ddiproietto:
gatto

PR #462, opened 2017-05-18
Fix pep8 package name

changlan:
The name is `pep8` in both trusty and xenial: http://packages.ubuntu.com/trusty/python/pep8

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -39,8 +39,14 @@
         - libgtest-dev
         - linux-headers-generic
         - lcov
-        - python-pep8
         - python-autopep8
+
+    - name: Install release-specific packages
+      apt: name={{item}} state=latest update_cache=yes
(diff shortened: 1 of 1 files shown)


sangjinhan:
If I remember correctly, on a clean Xenial machine, `autopep8`didn't work if only Python 2 is available... Could you check with Vagrant?

My comment from #426:
> Note that python-autopep8 package alone is not enough; its dependency package pep8 would only support Python 3, so python-pep8 should be also installed.

changlan:
Yes, you are right. `pep8` after Xenial becomes `python3-pep8` but `python-autopep8`'s package dependency on Xenial doesn't reflect this. I added a check before installing `python-pep8`.

sangjinhan:
tumblr_m08h7qre3k1rnx5oso1_400

PR #463, opened 2017-05-18
Fix queue constraint checking

apanda:
Not only should we not propagate workers for queues, but the constraint
checks themselves need to be different. This patch adds a new check
method.


sangjinhan:
Probably it's good to move non-trivial functions from .h to .cc.

apanda:
@sangjinhan Done

barath:
img_0965

PR #464, opened 2017-05-19
Fix potential gRPC race

sangjinhan:
gRPC servers are internally multi-threaded, so all RPC handlers must be thread-safe. BESS API handlers are not. This means, if there are multiple client connections, the daemon may crash due to race condition.

gRPC does not currently support a knob to limit the number of handler threads (only "idle" handler threads can be limited). Instead, I introduced a mutex to serialize all RPCs over concurrent connections. Since it could be tedious and error-prone to insert a mutex block in every handler, `mutex::lock()` and `mutex::unlock()` have been added to the callback functions called by all RPC messages.

In addition, the gRPC server has been encapsulated as a class object.

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -33,15 +33,8 @@
 #include "utils/time.h"
 #include "worker.h"
 
-using grpc::Server;
-using grpc::ServerBuilder;
-using grpc::ServerContext;
-using grpc::ServerReader;
-using grpc::ServerReaderWriter;
-using grpc::ServerWriter;
 using grpc::Status;
 using grpc::ServerContext;
-using grpc::ServerBuilder;
 
 using bess::priority_t;
 using bess::resource_t;
@@ -1643,11 +1636,11 @@ class BESSControlImpl final : public BESSControl::Service {
 };
 
 // TODO: C++-ify
-static std::unique_ptr<Server> server;
+static std::unique_ptr<grpc::Server> server;
 static BESSControlImpl service;
 
 void SetupControl() {
-  ServerBuilder builder;
+  grpc::ServerBuilder builder;
 
   if (FLAGS_p) {
     std::string server_address = bess::utils::Format("127.0.0.1:%d", FLAGS_p);
@@ -1655,6 +1648,7 @@ void SetupControl() {
     builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
   }
 
+  builder.SetSyncServerOption(grpc::ServerBuilder::MAX_POLLERS, 1);
   builder.RegisterService(&service);
   server = builder.BuildAndStart();
 }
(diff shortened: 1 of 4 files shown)


core/bessctl.cc
```diff
@@ -399,6 +381,8 @@ static void collect_tc(const bess::TrafficClass* c, int wid,
 
 class BESSControlImpl final : public BESSControl::Service {
  public:
+  void set_shutdown_func(std::function<void()> func) { shutdown_func_ = func; }
```
barath: Minor: could func be a const reference?

core/bessctl.cc
```diff
@@ -1640,38 +1632,62 @@ class BESSControlImpl final : public BESSControl::Service {
...
+  std::string addr = bess::utils::Format("%s:%d", host.c_str(), port);
+  LOG(INFO) << "Server listening on " << addr;
 
-  if (FLAGS_p) {
-    std::string server_address = bess::utils::Format("127.0.0.1:%d", FLAGS_p);
-    LOG(INFO) << "Server listening on " << server_address;
-    builder.AddListeningPort(server_address, grpc::InsecureServerCredentials());
+  int alloced_port = 0;
+  builder_->AddListeningPort(addr, grpc::InsecureServerCredentials(),
+                           &alloced_port);
```
barath: Minor: alignment seems funny.

core/bessctl.h
```diff
@@ -1,8 +1,30 @@
...
 
-void SetupControl();
+#include <string>
 
-void RunControl();
+namespace grpc {
+  class ServerBuilder;
+}  // namespace grpc
+
+class ApiServer {
```
barath: Could you add a class comment briefly describing how this is used?

core/bessctl.h
```diff
@@ -1,8 +1,30 @@
...
+ public:
+  ApiServer() : builder_(nullptr) {}
```
barath: Initialize grpc_cb_set_?

core/bessctl.cc
```diff
@@ -1640,38 +1634,58 @@ class BESSControlImpl final : public BESSControl::Service {
...
-void RunControl() {
-  auto serve_func = []() {
-    server->Wait();
-    LOG(INFO) << "Terminating gRPC server";
+void ApiServer::Run() {
+  class ServerCallbacks : public grpc::Server::GlobalCallbacks {
+   public:
+    ServerCallbacks() {}
+    void PreSynchronousRequest(ServerContext*) { mutex_.lock(); }
+    void PostSynchronousRequest(ServerContext*) { mutex_.unlock(); }
```
melvinw: Could we deadlock if a worker barfs in the middle of handling a request, or does `PostSynchronousRequest()` get called regardless?

melvinw: I guess bessd is almost always crashed in that case anyway. This may be a nonissue.

sangjinhan: Any hard crash in a gRPC handler thread will terminate the entire daemon process. Otherwise, the callback function will be called after any soft failures (e.g., not returning Status::OK).

melvinw: Gotcha

melvinw:
Thanks! 
4lyyz8ypufzy

PR #466, opened 2017-05-20
libbess-python: attach_module() to 'priority' or 'weighted_fair'.

ddiproietto:
attach_module() didn't allow attaching to 'priority' or 'weighted_fair'
parents, because it didn't have a way to specify the 'priority' or
'share' parameters.  This commit adds the parameters (they're similar to
what we have in add_tc()).


melvinw:
Thanks! schfunw5ldxy

PR #469, opened 2017-05-22
pcap: Support jumbo frame

changlan:
This PR adds jumbo frame support to the PCAP driver, roughly following this DPDK patch: http://dpdk.org/ml/archives/dev/2015-February/014404.html.

### DIFF
--- a/core/drivers/pcap.cc
+++ b/core/drivers/pcap.cc
@@ -1,5 +1,8 @@
 #include "pcap.h"
 
+#include <algorithm>
+#include <string>
+
 #include "../utils/pcap.h"
 
 CommandResponse PCAPPort::Init(const bess::pb::PCAPPortArg& arg) {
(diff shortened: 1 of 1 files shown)


core/drivers/pcap.cc
```diff
@@ -1,5 +1,8 @@
 #include "pcap.h"
 
+#include <algorithm>
```
sangjinhan: Maybe I am missing something, but the code change doesn't seem to need these headers here...

core/drivers/pcap.cc
```diff
@@ -47,27 +50,27 @@ int PCAPPort::RecvPackets(queue_t qid, bess::Packet** pkts, int cnt) {
...
+      m->set_next(bess::Packet::Alloc());
+      m = m->next();
+
+      // no headroom needed in chained mbufs
+      m->prepend(m->headroom());
+      m->set_data_len(0);
+      m->set_buffer(0);
+
+      copy_len = std::min(caplen, static_cast<int>(m->tailroom()));
+      bess::utils::CopyInlined(m->append(copy_len), packet, copy_len, true);
```
sangjinhan: I think a non-inlined copy is good enough here... We already know that the packet is large, so tiny performance gain would not outweigh the increased code size.

core/drivers/pcap.cc
```diff
@@ -47,27 +50,27 @@ int PCAPPort::RecvPackets(queue_t qid, bess::Packet** pkts, int cnt) {
...
-      break;
-#endif
+    int copy_len = std::min(caplen, static_cast<int>(sbuf->tailroom()));
+    bess::utils::CopyInlined(sbuf->append(copy_len), packet, copy_len, true);
+
+    packet += copy_len;
+    caplen -= copy_len;
+    bess::Packet* m = sbuf;
+
+    while (caplen > 0) {
```
sangjinhan: We need to update the `nb_segs_` field if chained.

core/drivers/pcap.cc
changlan: `std::min` uses `<algorithm>`, and `std::string` is used too.

core/drivers/pcap.cc
changlan: Sure. Fixed.

sangjinhan:
tumblr_m8im7ptqs81rdogh7o1_1280

PR #472, opened 2017-05-23
Account for tasks instead of workers

apanda:
Previously we were ensuring that each worker was propagated exactly once
through a module. This doesn't work since multiple tasks in a worker
might be using the same module. This change fixes this misaccounting.
Also fixes related bugs in queue.cc

### DIFF
--- a/bessctl/conf/testing/test_constraint_checker.bess
+++ b/bessctl/conf/testing/test_constraint_checker.bess
@@ -1,0 +1,42 @@
+# This is taken from queue.bess
+src::Source() \
+        -> queue::Queue() \
+        -> VLANPush(tci=2) \
+        -> Sink()
+
+bess.add_tc('fast', policy='rate_limit', resource='packet', limit={'packet': 9000000})
+bess.attach_module(src.name, 'fast')
+
+bess.add_tc('slow', policy='rate_limit', resource='packet', limit={'packet': 1000000})
+bess.attach_module(queue.name, 'slow')
+ret = bess.check_constraints()
+assert(not ret)
+bess.reset_all()
+
+# From nat.bess -- check that revisiting the same module works correctly.
+nat::NAT(rules=[{'internal_addr_block':'0.0.0.0/0', 'external_addr_block':'192.168.1.1/32'}])
+
+# Swap src/dst MAC
+mac::MACSwap()
+
+# Swap src/dst IP addresses / ports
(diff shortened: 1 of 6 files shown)


apanda:
@barath Tests added

core/module.h
```diff
@@ -335,6 +350,8 @@ class Module {
   std::vector<bess::OGate *> ogates_;
   // Set of active workers accessing this module.
   std::vector<bool> active_workers_;
+  // Set of tasks we have already accounted for when propagating workers.
+  std::vector<const ModuleTask *> visited_tasks_;
```
barath: Maybe use unordered_set or set instead of vector?

barath:
LGTM, pending Sangjin's review.

bessctl/conf/testing/test_constraint_checker.bess
```diff
@@ -0,0 +1,67 @@
+
+def queue_test():
```
sangjinhan: Perhaps we don't need `queue_test()` and `nat_test()` here, since they are already checked in `samples/nat.bess` and `sample/queue.bess`and they are already included in the sanity check?

libbess-python/bess.py
```diff
@@ -210,9 +211,12 @@ def check_constraints(self):
             for module in response.modules:
                 print 'constraints violated for module %s --'\
                     ' please check bessd log' % module.name
+            error = True
         if response.fatal:
             raise self.ConstraintError("Fatal violation of "
                                        "scheduling constraints")
+            error = True
```
sangjinhan: These lines will never be executed since the control flow already left this function. If you want a tri-state return value, how about something like:
* success: no return value
* soft/hard failure: raise ConstraintError("Scheduling violation(s) found", violations=response.violations, modules=response.modules, fatal=response.fatal)

So that the caller (e.g., bessctl) can deal with the error programmatically? The caller may want to make sure that there are even no minor errors, choose its own error message mechanism (e.g., `cli.err()`), etc.

bessctl/conf/testing/test_constraint_checker.bess
apanda: We don't check for soft constraint failures in sanity check, which is why I copied it here.

sangjinhan:
tumblr_m8he779vrg1qze0hyo1_400

PR #473, opened 2017-05-23
Fix the field type of `WorkerStatus`

changlan:


sangjinhan:
tumblr_m8gzng2msn1qargfho1_1280

PR #475, opened 2017-05-25
Extended Utility Codel

TheJStone:
Adds Factory, Generator and Resize methods(as well as a few minor fixups) to Utility Codel in order to allow integration into DRR.

### DIFF
--- a/core/utils/codel.h
+++ b/core/utils/codel.h
@@ -3,12 +3,12 @@
 
 #include <cmath>
 #include <deque>
+#include <functional>
 
 #include <glog/logging.h>
 
 #include "time.h"
 #include "queue.h"
-
 
 namespace bess {
 namespace utils {
@@ -50,6 +50,7 @@ class Codel final: public Queue<T> {
         queue_(),
         drop_func_(drop_func) { }
 
+  // deconstructor that drops all objects still left in the internal queue.  
   virtual ~Codel() {
     Wrapper w;
     while (!queue_.empty()) {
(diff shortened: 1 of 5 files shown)


core/utils/codel.h
```diff
@@ -135,7 +136,12 @@ class Codel final: public Queue<T> {
     return i;
   }
 
-  size_t Capacity() override { return queue_.max_size(); }
+  size_t Capacity() override {
```
apanda: Why is Codel's capacity different from the underlying queue capacity. We have already allocated memory (based on the Resize method below). It would be good to at least add a comment explaining this choice, since the current code is surprising.

core/utils/codel.h
```diff
@@ -148,6 +154,35 @@ class Codel final: public Queue<T> {
 
   size_t Size() override { return queue_.size(); }
 
+  int Resize(size_t new_capacity) override {
+    if (new_capacity <= Size()) {
+      return -1;
+    } else if (new_capacity >= queue_.max_size()) {
+      return -1;
```
apanda: Why does this method fail here, rather than calling `queue_.resize`?

core/utils/codel.h
TheJStone: I switched the underlying queue to deque which has a max size of the memory available to it according to the spec and I didn't see any constructor parameter to limit the deque size. In the use case of having many codels in drr handling a burst, it would be better to restrict queue size rather than have issues with memory although the max_entries can be specified as 0 to default to the memory size. I will add a comment but is allowing restriction not a good idea?

core/utils/codel.h
TheJStone: I changed the underlying queue to be `deque` so the `max_size` is the max number of entries that the queue can have given system limitations.

core/utils/codel.h
apanda: That is OK, just add a comment with the reason you gave above.

apanda: The restriction is fine -- I think it needs to be documented. I am not super sure how one picks `max_size_` -- it seems to depend on the workload, but c'est la vie.

core/utils/codel.h
apanda: Just document it -- it looks strange to have a resize method which mostly does nothing.

barath:
Just wanted to follow up -- Josh, do you think you're pretty much done making the changes Panda asked about?  If so, we can merge this in the next day or two.

TheJStone:
ya, I can finish it up tonight and push. I will ping you when I do.

apanda:
Merged, thanks!
image

PR #477, opened 2017-05-25
Fixes call to std::abs() to address clang 4.0 warning.

barath:


sangjinhan:
What was the error clang complained about? (underflow?)

barath:
modules/measure.cc:83:29: error: taking the absolute value of unsigned type
'unsigned long' has no effect [-Werror,-Wabsolute-value]

sangjinhan:
Oh I see. Might be useful to add this to `common.h`...
```
template <typename T>
T absdiff(const T &lhs, const T &rhs) {
  return lhs > rhs ? lhs - rhs : rhs - lhs;
}
```

sangjinhan:
tumblr_m8o14n0pex1rw872io1_500

PR #479, opened 2017-05-26
Update NAT module

sangjinhan:
This PR revamps the NAT module, with the following changes:
* More compliant to the standards. Most importantly, the new module performs [Endpoint-Independent Mapping](https://tools.ietf.org/html/rfc4787#section-4.1), which is required for NAT hole punching.
* The memory footprint has been reduced (total 64 bytes/entry; used to be 128 bytes/entry) and only proportional to the number of concurrent connections. It doesn't not maintain the list of free IP/port pairs any longer; using a large external IP address pool won't increase memory usage.
* The garbage collection phase has been eliminated. The module won't stop entirely even if IP/port pairs run out.
* Performance improvement (below)

### Local test (unidirectional)
All numbers are in Mpps.

Flows | New | Old
--------- | ----- | ----
1 | 43.8 | 27.2 |
16 | 29.2 | 20.2 |
256 | 30.1 | 20.7 |
4096 | 24.7 | 15.6 |
65536 | 19.4 | 13.2 |

### Local test (bidirectional)
For actual pps, multiple these numbers by 2

Flows | New | Old
--------- | ----- | ----
1 | 23.1 | 13.8 |
16 | 16.7 | 9.9 |
256 | 15.4 | 10.4 |
4096 | 12.6 | 7.7 |
65536 | 9.3 | 6.8 |

### With i40e NICs
Flows | New | Old
--------- | ----- | ----
1 | 16.2 | 12.9 |
16 | 13.8 | 11.3 |
256 | 13.0 | 10.0 |
4096 | 12.2 | 7.1 |
65536 | 11.6 | 6.6 |
1048576 | 6.2 | 3.8 |

### DIFF
--- a/core/modules/nat.cc
+++ b/core/modules/nat.cc
@@ -81,7 +81,8 @@ static inline std::pair<bool, Endpoint> ExtractEndpoint(const Ipv4 *ip,
     }
   }
 
-  return std::make_pair(false, Endpoint{});
+  return std::make_pair(
+      false, Endpoint{.addr = ip->src, .port = be16_t(0), .protocol = 0});
 }
 
 // Not necessary to inline this function, since it is less frequently called
--- a/core/modules/nat.cc
+++ b/core/modules/nat.cc
@@ -44,8 +44,8 @@ CommandResponse NAT::Init(const bess::pb::NATArg &arg) {
 }
 
 static inline std::pair<bool, Endpoint> ExtractEndpoint(const Ipv4 *ip,
-                                                 const void *l4,
-                                                 NAT::Direction dir) {
+                                                        const void *l4,
+                                                        NAT::Direction dir) {
   IpProto proto = static_cast<IpProto>(ip->protocol);
 
   if (likely(proto == IpProto::kTcp || proto == IpProto::kUdp)) {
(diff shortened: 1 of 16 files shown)


core/modules/nat.h
```diff
@@ -10,173 +13,112 @@
...
-#if __SSE4_2__ && __x86_64
-    init_val = crc32c_sse42_u64(bytes.u64[0], init_val);
-    init_val = crc32c_sse42_u64(bytes.u64[1], init_val);
-#else
-    init_val = rte_hash_crc(&bytes.u64[0], sizeof(uint64_t), init_val);
-    init_val = rte_hash_crc(&bytes.u64[1], sizeof(uint64_t), init_val);
-#endif
-    return init_val;
-  }
+// timestamp is only refreshed for forward mappings (rfc4787 REQ-6)
```
sangjinhan: The field name used to be `timestamp` but forgot to update. Thanks!

core/modules/nat.h
```diff
@@ -10,173 +13,112 @@
...
 using bess::utils::be32_t;
-using bess::utils::Ipv4Prefix;
 
-const uint16_t MIN_PORT = 1024;
-const uint16_t MAX_PORT = 65535;
-const uint64_t TIME_OUT_NS = 120ull * 1000 * 1000 * 1000;
+struct alignas(8) Endpoint {
+  be32_t addr;
+  be16_t port;    // identifier if ICMP
+  uint16_t protocol;
```
sangjinhan: Done.

core/modules/nat.cc
```diff
@@ -18,173 +19,199 @@ using IpProto = bess::utils::Ipv4::Proto;
...
-  rules_.clear();
-  flow_hash_.Clear();
+  if (likely(proto == IpProto::kTcp || proto == IpProto::kUdp)) {
+    // UDP and TCP share the same layout for port numbers
+    const Udp *udp = static_cast<const Udp *>(l4);
+    Endpoint ret;
 
-  return CommandResponse();
-}
+    if (dir == NAT::kForward) {
```
sangjinhan: The if statement will disappear at compile time. The functions take `dir` as a template parameter, so that the compiler will generate separate binary code for each forward and reverse direction. Eventually all the functions are inlined as two `ProcessBatch<kForward>()` and `ProcessBatch<kReverse>()`.

barath:
LGTM, will merge after Travis.

barath:
img_1060

PR #480, opened 2017-05-26
Rename attach_module() to attach_task() and add module.attach_task

ddiproietto:
This only makes the interface easier to use from .bess scripts.

Backwards compatibility is preserved for now


sangjinhan:
tumblr_m8y88l1hf21rro1o5o1_1280

PR #484, opened 2017-05-29
Remove redundant python imports

GalSagie:
Remove unused imports from all python files that
are not auto generated


changlan:
image

PR #487, opened 2017-05-29
Add copy command to Ethernet MAC address class

GalSagie:


core/utils/ether.h
```diff
@@ -25,6 +25,8 @@ struct[[gnu::packed]] Ethernet {
     // Returns "aa:bb:00:11:22:33" (all in lower case)
     std::string ToString() const;
 
+    void CopyFrom(const Address &o);
```
barath: Could you add a method comment?

GalSagie: I really thing the method name and parameter are self explanatory,  
I usually prefer in such cases to write less then more.
What do you say?

barath: I agree, in most cases.  Here, since I've seen slightly different behavior in various libraries in their use of CopyFrom, I figured clarity is preferable.

However, looking at it more closely, right now it appears the copy constructor and assignment operator are not deleted, so is this method even needed?

GalSagie: Yes you are right, its not needed, getting used to C++ again :)

GalSagie:
This function is redundant as Ethernet::Address a = b  (where b is another Address object will work just as well)

PR #490, opened 2017-05-30
Migrate sanity_check.sh to Python unittest

sangjinhan:
`sanity_check.sh` is a simple shell script that run some selected BESS scripts to see if anything goes seriously wrong (e.g., crash). It provides a useful means of regression test, and this PR adds a few improvements to it:

* Instead of manually specifying which scripts are to be run, now it tries all scripts in the `bessctl/conf/samples` directory.
  * For this, all scripts in the directory have been rearranged so that:
    * It does not creates physical/virtual ports as they might not be available on the test system.
    * No external dependencies are used (e.g., Docker)
    * Scripts are not interactive (e.g., no artificial time delay)
  * Scripts that do not meet the above conditions have been moved to other directories
* Migrate from a custom shell script to the Python unittest framework
  * Down the road, we will add coverage measurement for Python code.
* The old script took about 6-7 minutes to complete, adding a long delay to the CI build process. Now it takes less than a minute.
* Besides ones in `bessctl/conf/samples`, all other BESS scripts are tested with syntax (and syntactic sugar added).


bessctl/test_sugar.py
```diff
@@ -0,0 +1,100 @@
...
+import unittest
+
+
+class TestSugar(unittest.TestCase):
+    """
+    All scripts in conf/ will be dynamically added here as individual
+    tests (e.g., test_conf_samples_exactmatch_bess) in this class.
+    The tests will just perform nothing but sugar.xform_file() and
+    check if the resulting code is syntactically correct.
+    This is to see if it causes any expcetions during the process.
```
justinemarie: exceptions (sp)

bessctl/test_sugar.py
```diff
@@ -0,0 +1,100 @@
+import fnmatch
+import os
+import sugar
+import unittest
+
+
+class TestSugar(unittest.TestCase):
+    """
+    All scripts in conf/ will be dynamically added here as individual
```
justinemarie: Is this the part that codecov detects? Presumably codecov won't detect the calls out to the shell? (How does the codecov integration work here?)

sangjinhan: Adding Python coverage to codecov will be done in a separate PR (still figuring out)

bessctl/test_sugar.py
sangjinhan: fixed!

bessctl/test_sugar.py
justinemarie: I would guess that shelling out will break it but I guess that can be fixed when you figure it out. Okay just curious, thought that you had figured out how to do it even with the shelling out :-)

sangjinhan: Seems both C++ and Python are working now: https://codecov.io/gh/sangjinhan/bess/branch/travis_test8

justinemarie:
Thanks!
image

PR #492, opened 2017-05-31
Add Python code coverage & Fix C++ coverage fluctuation

sangjinhan:
> NOTE: This PR renames the directory `libbess-python` to `pybess` (see below for details). If you have out-of-tree controllers you will need to update them.

This PR contains two changes regarding code coverage measurement on Codecov. Both Python and C++ coverage results are merged. An example can be seen here: https://codecov.io/gh/sangjinhan/bess/tree/travis_test8

1. Add Python code coverage measurement
   - All Python unittests in the code tree are automatically run. By Python convention, the directory should be a valid Python package (has `__init__.py`), and unittest files should start with `test_`.
   - Currently `bessctl` and `pybess` directories are measured.
     - `libbess-python` directory has been renamed to `pybess`. Python packages must not have a `-` in their name.
     - `bessctl` is now a Python package.
   - In Travis build, [codecov-bash](https://github.com/codecov/codecov-bash) has been replaced with [codecov-python](https://github.com/codecov/codecov-python), as the latter is simpler to use and covers more languages.
   - Auto-generated `*_pb2.py` files are not included in the report.
1. Fix C++ coverage result fluctuation (#388)
   - A few issues have contributed to the incorrect measurement resulting in fluctuation of coverage percentage numbers.
     - `bessd` daemon was not gracefully shutdown but killed by a signal, adding some non-determinism.
     - Code was built with the `-O1` optimization option, which caused incorrect line accounting. Now it is built with `-O0` in DEBUG mode.
       - There were some build errors that prevented `-O0`. Fixed.
       - DPDK causes some build issue with `g++ -O0`. I made a dirty fix in `core/modules/ip_lookup.cc`. I will send a patch to DPDK mainstream...
   - Now the result seems to make more sense: about 63% at the moment. I am not sure if the number will be stable, but let's see.

This PR includes commits from #490. I will rebase this branch to the master once #490 has been merged.


sangjinhan:
Rebased, now ready for review&merge.

sangjinhan:
Patch for 764feea has been submitted to the DPDK upstream.

changlan:
image

Issue #493, opened 2017-06-01
tcpdump gives segmentation fault when using wrong module name


GalSagie:
Log attached: working on fixing this

F0601 10:55:18.417379 55718 debug.cc:359] A critical error has occured. Aborting...
Signal: 11 (Segmentation fault), si_code: 1 (SEGV_MAPERR: address not mapped to object)
pid: 55703, tid: 55718, address: 0x90, IP: 0x67fa10
Backtrace (recent calls first) ---
(0): ./bin/bessd() [0x67fa10]
    enable_hook_for_module(Module const*, unsigned short, bool, bool, bess::GateHookFactory const&, google::protobuf::Any const&) [clone .constprop.513] at /home/gsagie/bess/core/bessctl.cc:118
         115:       }
         116:     }
         117:   } else {
      -> 118:     for (auto& gate : m->ogates()) {
         119:       if (!gate) {
         120:         continue;
         121:       }
(1): ./bin/bessd(_ZN15BESSControlImpl17ConfigureGateHookEPN4grpc13ServerContextEPKN4bess2pb24ConfigureGateHookRequestEPNS4_15CommandResponseE+0x14c) [0x67fe8c]
    BESSControlImpl::ConfigureGateHook(grpc::ServerContext*, bess::pb::ConfigureGateHookRequest const*, bess::pb::CommandResponse*) at /home/gsagie/bess/core/bessctl.cc:1334
      -> 1334:           enable_hook_for_module(it->second, gate_idx, is_igate, use_gate,
(2): ./bin/bessd(_ZNSt17_Function_handlerIFN4grpc6StatusEPN4bess2pb11BESSControl7ServiceEPNS0_13ServerContextEPKNS3_24ConfigureGateHookRequestEPNS3_15CommandResponseEESt7_Mem_fnIMS5_FS1_S8_SB_SD_EEE9_M_invokeERKSt9_Any_dataOS6_OS8_OSB_OSD_+0x47) [0x6d67c7]
[...]

PR #495, opened 2017-06-01
Check module exists before applying tcpdump

GalSagie:
#493 
This patch verify that the module name exists before
applying tcpdump command from python SDK.

### DIFF
--- a/libbess-python/bess.py
+++ b/libbess-python/bess.py
@@ -398,6 +398,9 @@ class BESS(object):
         return self._request('ConfigureGateHook', request)
 
     def tcpdump(self, enable, m, direction='out', gate=0, fifo=None):
+        # Verify the module exists, otherwise raise exception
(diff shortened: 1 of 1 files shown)


sangjinhan:
In my opinion, this patch might be somewhat redundant. #496 will show a correct error message anyway...
```
localhost:10514 $ tcpdump rewrit out 0 -ne
  Running: tcpdump -r /tmp/tmpP5kVKW -ne | less
*** Error: No module 'rewrit' found
  BESS daemon response - errno=2 (ENOENT: No such file or directory)
       query: ConfigureGateHook
   query_arg: {...}
```

GIven that no other commands explicitly check if a module (or any other entities) exists in `commands.py`, I think we can skip this patch...?

GalSagie:
I agree (even that i usually prefer to protect against such cases in the SDK/API level)

PR #496, opened 2017-06-01
Fix segmentation on gate hook configure for invalid module name

GalSagie:
Since there was no return in the code, the method would continue
to the next line and segment as the module doesnt exists.
This change return the error in case the module is not found.
#493


sangjinhan:
Merged, thanks!
tumblr_kzqdnyhxd61qb96vjo1_1280

PR #498, opened 2017-06-01
Update gRPC from v1.1.2 to v1.3.2

sangjinhan:
It was reported that older versions of gRPC are not compiled with newer clang versions...

Also, slightly updated `env/packages.yml` to avoid some redundant downloads (gRPC already has them in its repository).


barath:
photo 4

PR #499, opened 2017-06-03
Modifies task_result to contain block field.

barath:


core/module.h
```diff
@@ -15,19 +15,20 @@
...
+#define INVALID_TASK_ID ((task_id_t)-1)
 #define MAX_NUMA_NODE 16
+#define MAX_TASKS_PER_MODULE 32
 #define UNCONSTRAINED_SOCKET ((0x1ull << MAX_NUMA_NODE) - 1)
 
 struct task_result {
-  uint64_t packets;
+  uint32_t packets;
   uint64_t bits;
+  bool block;
```
barath: Ok good, I was worried about that -- would you prefer first or second?

sangjinhan: Perhaps first...? (so that `packets` and `bits` are grouped together)

barath: Done.

PR #500, opened 2017-06-05
Python 2/3 Compatibility

### LINKED ISSUE #488: Installing BESS on working system with grpc and python3 fails
I am not sure if thats really an issue, but currently if older versions of grpc and python3 exists on
the server the running of bessctl fails.

(I suspect that the reason is we should force python2.7 to install grpcio)

Any thoughts?

changlan:
This should fix #488.

- Modified Python code to be 2/3 compatible 
  - Use Python 3 style `print()` with `from __future__ import print_function`;
  - Use list comprehension instead of `map` and `filter`;
  - Use `items()` instead of `iteritems()` (Yeah, it would become less efficient in Python 2);
  - Use `io` package to `read()` from files, so that the resulting string is Unicode-encoded;
  - Use `raw_input` for Python 2 and `input` for Python 3;
  - Disable implicit relative import and use absolute import instead;
  - Fix the discrepancy of `tokenizer`  between Python 2/3 (`->` is an `OP` token in Python 3);
  - Run `2to3` over `protoc` generated scripts, since those scripts still use implicit relative import;
  - Use raw bytes for non-Unicode literals (e.g `b'\x01'`).
- Add unit test for both Python 2/3.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ before_install:
 install:
   - sudo apt-get install -y python2.7 python3 python3-pip
   - pip2 install grpcio scapy codecov
-  - pip3 install grpcio coverage
+  - pip3 install grpcio scapy-python3 coverage
   - "[[ ${DEBUG:-0} == 0 ]] || sudo apt-get install -y g++-5"  # install gcov-5
   - "[[ ${SANITIZE:-0} == 0 ]] || sudo apt-get install -y llvm-3.8"
   - "[[ $TAG_SUFFIX != _32 ]] || sudo apt-get install -y lib32gcc1"
@@ -53,7 +53,8 @@ script:
   - (cd core && ./all_test)  # TcpFlowReconstructTest requires working directory to be `core/`
   - coverage2 run -m unittest discover -v
   - coverage3 run -m unittest discover -v
-  - bessctl/bessctl -- daemon start -- run testing/run_module_tests
+  - python2 bessctl/bessctl -- daemon start -- run testing/run_module_tests
+  - python3 bessctl/bessctl -- daemon start -- run testing/run_module_tests
 
 after_success:
   - bessctl/bessctl daemon stop # flush out the coverage data
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,9 +52,9 @@ script:
   - ./container_build.py kmod_buildtest
   - (cd core && ./all_test)  # TcpFlowReconstructTest requires working directory to be `core/`
   - coverage2 run -m unittest discover -v
-  - coverage3 run -m unittest discover -v
+  - "[[ ${DEBUG:-0} == 0 ]] || coverage3 run -m unittest discover -v"
   - python2 bessctl/bessctl -- daemon start -- run testing/run_module_tests
-  - python3 bessctl/bessctl -- daemon start -- run testing/run_module_tests
+  - "[[ ${DEBUG:-0} == 0 ]] || python3 bessctl/bessctl -- daemon start -- run testing/run_module_tests"
 
 after_success:
   - bessctl/bessctl daemon stop # flush out the coverage data
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ before_install:
   - sudo apt-get -q update
 
 install:
-  - sudo apt-get install -y python python3 python3-pip
-  - pip install grpcio scapy codecov
+  - sudo apt-get install -y python2.7 python3 python3-pip
+  - pip2 install grpcio scapy codecov
   - pip3 install grpcio coverage
   - "[[ ${DEBUG:-0} == 0 ]] || sudo apt-get install -y g++-5"  # install gcov-5
   - "[[ ${SANITIZE:-0} == 0 ]] || sudo apt-get install -y llvm-3.8"
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,9 @@ before_install:
   - sudo apt-get -q update
 
 install:
-  - sudo apt-get install -y python
+  - sudo apt-get install -y python python3 python3-pip
   - pip install grpcio scapy codecov
+  - pip3 install grpcio coverage
   - "[[ ${DEBUG:-0} == 0 ]] || sudo apt-get install -y g++-5"  # install gcov-5
   - "[[ ${SANITIZE:-0} == 0 ]] || sudo apt-get install -y llvm-3.8"
   - "[[ $TAG_SUFFIX != _32 ]] || sudo apt-get install -y lib32gcc1"
@@ -50,7 +51,8 @@ script:
   - ./container_build.py bess
   - ./container_build.py kmod_buildtest
   - (cd core && ./all_test)  # TcpFlowReconstructTest requires working directory to be `core/`
-  - coverage run -m unittest discover -v
+  - coverage2 run -m unittest discover -v
+  - coverage3 run -m unittest discover -v
   - bessctl/bessctl -- daemon start -- run testing/run_module_tests
 
 after_success:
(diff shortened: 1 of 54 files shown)


bessctl/bessctl
```diff
@@ -1,24 +1,25 @@
 #!/usr/bin/env python2.7
```
sangjinhan: Now all scripts should start with "#!/usr/bin/env python", so that they can run regardless of the installed Python version..?

.travis.yml
```diff
@@ -31,8 +31,9 @@ before_install:
   - sudo apt-get -q update
 
 install:
-  - sudo apt-get install -y python
+  - sudo apt-get install -y python python3 python3-pip
```
sangjinhan: While it is not an issue on our test setup, `python` might be an alias of python3 on other distros. I would be better to specify `python2` explicitly.

.travis.yml
```diff
@@ -31,8 +31,9 @@ before_install:
...
   - pip install grpcio scapy codecov
```
sangjinhan: `pip2`

.travis.yml
```diff
@@ -31,8 +31,9 @@ before_install:
...
+  - pip3 install grpcio coverage
```
sangjinhan: (why not `scapy` for pip3?)

bessctl/bessctl
sangjinhan: (sorry not all scripts, but executable scripts)

bessctl/bessctl
```diff
@@ -141,14 +142,15 @@ def main():
         line_buf = []
 
         for arg in sys.argv[1:]:
+            arg = arg.decode()
```
chris3torek: The elements of `sys.argv` should already be strings (not bytes) in py3k.  Are you sure this is needed?

build.py
```diff
@@ -111,7 +112,7 @@ def check_c_lib(lib):
 def required(header_file, lib_name, compiler):
     if not check_header(header_file, compiler):
         print >> sys.stderr, 'Error - #include <%s> failed. ' \
```
chris3torek: Seems to be missing its conversion...?

bessctl/bessctl
sangjinhan: Why do we convert it into a unicode..? (please add a comment)

container_build.py
```diff
@@ -73,9 +74,9 @@ def do_dist_clean():
 
 def print_usage():
     print >> sys.stderr, \
```
chris3torek: This also still has the `>> sys.stderr` (and is not a function call)...

pybess/protobuf_to_dict.py
```diff
@@ -16,26 +15,25 @@
...
     FieldDescriptor.TYPE_FIXED32: int,
-    FieldDescriptor.TYPE_FIXED64: long,
+    FieldDescriptor.TYPE_FIXED64: int,
     FieldDescriptor.TYPE_SFIXED32: int,
-    FieldDescriptor.TYPE_SFIXED64: long,
+    FieldDescriptor.TYPE_SFIXED64: int,
     FieldDescriptor.TYPE_BOOL: bool,
-    FieldDescriptor.TYPE_STRING: unicode,
+    FieldDescriptor.TYPE_STRING: str,
     FieldDescriptor.TYPE_BYTES: str,
```
sangjinhan: I am not quite sure about this, but shouldn't it be bytes...?

pybess/protobuf_to_dict.py
```diff
@@ -16,26 +15,25 @@
...
-    FieldDescriptor.TYPE_SINT64: long,
+    FieldDescriptor.TYPE_SINT64: int,
     FieldDescriptor.TYPE_FIXED32: int,
-    FieldDescriptor.TYPE_FIXED64: long,
+    FieldDescriptor.TYPE_FIXED64: int,
     FieldDescriptor.TYPE_SFIXED32: int,
-    FieldDescriptor.TYPE_SFIXED64: long,
+    FieldDescriptor.TYPE_SFIXED64: int,
     FieldDescriptor.TYPE_BOOL: bool,
-    FieldDescriptor.TYPE_STRING: unicode,
```
chris3torek: I don't quite understand how this is to work in both py2k and py3k but I also don't quite understand how it works now in py2k in the first place, so I'm just confessing ignorance here.  😄   (In particular the values have to already have been converted somewhere by this point.)

.travis.yml
changlan: `scapy` for `pip3` is broken :( 
And the tests don't need `scapy`.

bessctl/bessctl
changlan: `StringIO` takes unicodes. I changed it to `BytesIO` for clarity (and thus avoiding the conversion).

build.py
changlan: Ah, this seems to be a stale version.

.travis.yml
sangjinhan: Oh... then perhaps `scapy-python3` is the package we need to install (https://github.com/phaethon/scapy)...

Also `bessctl/bessctl -- daemon start -- run testing/run_module_tests` in `.travis.yml` now needs "python2" explicitly, as it does use scapy?

bessctl/bessctl
changlan: Explained in @sangjinhan's comment above.

.travis.yml
sangjinhan: Also we would need to replace `str(scapy packet)` with `bytes(scapy packets)` in BESS scripts... The scapy3k webpage says:

> N.B.! As a difference from scapy for python2, use bytes() instead of str() when converting packet to bytes. Also, most arguments expect bytes value instead of str value except the ones, which are naturally suited for human input (e.g. domain name).

bessctl/bessctl
changlan: Ah, it seems that in Python 2 arguments are merely `str` (not `unicode`), while in Python 3 they are Unicode `str`, so we have to convert them into unicode first.

chris3torek: py3k str is Unicode, yes. (The type name `unicode` no longer exists, it's all just `str`.) py2k str and bytes are the same underlying type. The py2k io.StringIO` does require unicode, so probably this should use `io.BytesIO` on py2k and `io.StringIO` on py3k. Or you can use `six.StringIO` here, if you're importing `six` (I forget if you are).

changlan: Yes, I guess it is easier to just make sure the input are Unicode (already did) instead of handling StringIO and BytesIO separately. I am attempting to avoid extra packages like `six` and `future` to keep the dependency minimal 😄

chris3torek: Sure, but you'll have to not do `.decode()` if it's py3k (either that, or trap `AttributeError` as there is no decode for str objects).

changlan: It is been fixed in https://github.com/NetSys/bess/pull/500/commits/5e4a82d4967565185f0cd1c9860328da1e1b882c

chris3torek: Oh, clever! u' '.join works in both, converting py2k string-as-bytes to unicode, leaving py3k strings alone.

bessctl/bessctl
changlan: Done.

.travis.yml
changlan: Done. `python2` is not a package name, but I changed `python` to `python2` in other places.

.travis.yml
changlan: Done.

.travis.yml
changlan: Done.

pybess/protobuf_to_dict.py
changlan: Done.

.travis.yml
sangjinhan: python2.7...?

changlan: ... Done

sangjinhan:
tumblr_laq3zzq8ph1qe2yaso1_500

PR #501, opened 2017-06-05
add basic tests for buffer module

justinemarie:
This PR adds a crash test and an output test for the buffer module.


sangjinhan:
tumblr_lf6u195mdd1qawjt8o1_1280

sangjinhan:
This PR adds about 3minutes for each CI build, due to the socket timeout mechanism... :-/

PR #504, opened 2017-06-06
Add backpressure queue

vivi:
Modifies the Queue module to support backpressure.

### DIFF
--- a/core/module.cc
+++ b/core/module.cc
@@ -16,6 +16,67 @@
 const Commands Module::cmds;
 
 std::map<std::string, Module *> ModuleBuilder::all_modules_;
+std::map<std::string, Node> ModuleBuilder::module_graph_;
+std::map<std::string, Node> ModuleBuilder::task_graph_;
+
+int ModuleBuilder::AddEdge(const std::string &from, const std::string &to) {
+  auto it = module_graph_.find(from);
+  if (it == module_graph_.end()) {
+    return -EINVAL;
+  }
+  it->second.children().insert(to);
+  UpdateTCGraph();
+  return 0;
+}
+
+int ModuleBuilder::UpdateTCGraph() {
+  int ret = 0;
+  for (auto const &tc : task_graph_) {
+    std::unordered_set<std::string> visited;
+    ret = FindNextTC(tc.first, tc.first, visited);
+    if (ret != 0) {
+      return ret;
+    }
+  }
+  return ret;
+}
+
+int ModuleBuilder::FindNextTC(const std::string &node_name,
+                              const std::string &parent_name,
+                              std::unordered_set<std::string> &visited) {
+  visited.insert(node_name);
+  // While traversing the module graph, if `node` is in the task graph and is
+  // not `parent`, then it must be  the child of `parent`.
+  if (node_name != parent_name &&
+      task_graph_.find(node_name) != task_graph_.end()) {
+    auto parent_it = task_graph_.find(parent_name);
+    auto node_it = task_graph_.find(node_name);
+    if (parent_it == task_graph_.end() || node_it == task_graph_.end()) {
+      return -EINVAL;
+    }
+    parent_it->second.children().insert(node_name);
+    node_it->second.parents().insert(parent_it->second.module());
+    return 0;
+  }
+
+  auto node_it = module_graph_.find(node_name);
+  if (node_it == module_graph_.end()) {
+    return -EINVAL;
+  }
+
+  int ret;
+  for (auto &child_name : node_it->second.children()) {
+    auto it = visited.find(child_name);
+    if (it != visited.end()) {
+      continue;
+    }
+    ret = FindNextTC(child_name, parent_name, visited);
+    if (ret != 0) {
+      return ret;
+    }
+  }
+  return ret;
+}
 
 Module *ModuleBuilder::CreateModule(const std::string &name,
                                     bess::metadata::Pipeline *pipeline) const {
@@ -23,6 +84,12 @@ Module *ModuleBuilder::CreateModule(const std::string &name,
   m->set_name(name);
   m->set_module_builder(this);
   m->set_pipeline(pipeline);
+  module_graph_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
+                        std::forward_as_tuple(m));
+  if (m->IsTask()) {
+    task_graph_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
+                        std::forward_as_tuple(m));
+  }
   return m;
 }
 
@@ -55,6 +122,10 @@ int ModuleBuilder::DestroyModule(Module *m, bool erase) {
 
   if (erase) {
     all_modules_.erase(m->name());
+  }
+  module_graph_.erase(m->name());
+  if (m->IsTask()) {
+    task_graph_.erase(m->name());
   }
 
   delete m;
@@ -402,7 +473,8 @@ int Module::ConnectModules(gate_idx_t ogate_idx, Module *m_next,
   ogate->AddHook(new Track());
   igate->PushOgate(ogate);
 
-  return 0;
+  // Update graph
+  return ModuleBuilder::AddEdge(name(), m_next->name());
 }
 
 int Module::DisconnectModules(gate_idx_t ogate_idx) {
@@ -510,6 +582,40 @@ void Module::RunSplit(const gate_idx_t *out_gates,
   /* phase 3: fire */
   for (int i = 0; i < num_pending; i++)
     RunChooseModule(pending[i], &batches[i]);
+}
+
+void Module::SignalOverload() {
+  if (overload_) {
+    return;
+  }
+
+  std::unordered_set<Module *> *parents = ModuleBuilder::Parents(name());
+  if (parents == nullptr) {
+    LOG(ERROR) << name() << " has no entry in task graph";
+    return;
+  }
+  for (auto const &p : *parents) {
+    ++(p->children_overload_);
+  }
+
+  overload_ = true;
+}
+
+void Module::SignalUnderload() {
+  if (!overload_) {
+    return;
+  }
+
+  std::unordered_set<Module *> *parents = ModuleBuilder::Parents(name());
+  if (parents == nullptr) {
+    LOG(ERROR) << name() << " has no entry in task graph";
+    return;
+  }
+  for (auto const &p : *parents) {
+    --(p->children_overload_);
+  }
+
+  overload_ = false;
 }
 
 #if SN_TRACE_MODULES
--- a/core/module.cc
+++ b/core/module.cc
@@ -28,6 +28,27 @@ bool ModuleBuilder::AddEdge(const std::string &from, const std::string &to) {
   return UpdateTCGraph();
 }
 
+bool ModuleBuilder::RemoveEdge(const std::string &from, const std::string &to) {
+  auto from_it = all_modules_.find(from);
+  auto from_node = module_graph_.find(from);
+  auto to_it = all_modules_.find(from);
+  if (from_it == all_modules_.end() || from_node == module_graph_.end() ||
+      to_it == all_modules_.end()) {
+    return false;
+  }
+
+  from_node->second.RemoveChild(to);
+
+  // We need to regenerate the task graph.
+  for (auto const &task : tasks_) {
+    auto it = all_modules_.find(task);
+    if (it != all_modules_.end()) {
+      it->second->parent_tasks_.clear();
+    }
+  }
+  return UpdateTCGraph();
+}
+
 bool ModuleBuilder::UpdateTCGraph() {
   for (auto const &task : tasks_) {
     std::unordered_set<std::string> visited;
@@ -94,12 +115,13 @@ bool ModuleBuilder::AddModule(Module *m) {
       return false;
     }
   }
-  return all_modules_.insert({m->name(), m}).second &&
-         module_graph_
-             .emplace(std::piecewise_construct,
-                      std::forward_as_tuple(m->name()),
-                      std::forward_as_tuple(m))
-             .second;
+  bool module_added = all_modules_.insert({m->name(), m}).second;
+  bool node_added =
+      module_graph_
+          .emplace(std::piecewise_construct, std::forward_as_tuple(m->name()),
+                   std::forward_as_tuple(m))
+          .second;
+  return module_added && node_added;
 }
 
 int ModuleBuilder::DestroyModule(Module *m, bool erase) {
@@ -128,6 +150,7 @@ int ModuleBuilder::DestroyModule(Module *m, bool erase) {
   if (erase) {
     all_modules_.erase(m->name());
   }
+
   module_graph_.erase(m->name());
   if (m->is_task_) {
     tasks_.erase(m->name());
@@ -479,10 +502,7 @@ int Module::ConnectModules(gate_idx_t ogate_idx, Module *m_next,
   igate->PushOgate(ogate);
 
   // Update graph
-  if (ModuleBuilder::AddEdge(name_, m_next->name_)) {
-    return 0;
-  }
-  return 1;
+  return !(ModuleBuilder::AddEdge(name_, m_next->name_));
 }
 
 int Module::DisconnectModules(gate_idx_t ogate_idx) {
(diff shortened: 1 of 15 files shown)


core/module.h
```diff
@@ -18,6 +19,14 @@ using bess::gate_idx_t;
 #define MAX_NUMA_NODE 16
 #define UNCONSTRAINED_SOCKET ((0x1ull << MAX_NUMA_NODE) - 1)
 
+class Node {
```
barath: Add a class comment.

core/module.h
```diff
@@ -18,6 +19,14 @@ using bess::gate_idx_t;
...
+ public:
+  Node(Module *module) : module_(module), children_(), parents_() {}
```
barath: Add comments for each method (including the constructor and its args) and for each member field.

core/module.h
```diff
@@ -18,6 +19,14 @@ using bess::gate_idx_t;
...
+  Module *module_;
```
barath: The fields should be private.

core/module.h
```diff
@@ -18,6 +19,14 @@ using bess::gate_idx_t;
...
+  std::unordered_set<std::string> children_;
+  std::unordered_set<Module *> parents_;
+};
```
barath: Do you need Node to be copyable?  If not, use DISALLOW_COPY_AND_ASSIGN(Node) in the private section or delete the constructors.

core/module.h
```diff
@@ -137,7 +146,24 @@ class ModuleBuilder {
 
   CommandResponse RunInit(Module *m, const google::protobuf::Any &arg) const;
 
+  static int AddEdge(const std::string &from, const std::string &to);
```
barath: Function comment.

core/module.h
```diff
@@ -137,7 +146,24 @@ class ModuleBuilder {
...
+
+  static std::unordered_set<Module *> *Parents(const std::string &node) {
```
barath: Function comment.

core/module.h
```diff
@@ -137,7 +146,24 @@ class ModuleBuilder {
...
+    auto it = task_graph_.find(node);
+    if (it == task_graph_.end()) {
+      return nullptr;
+    }
+    return &(it->second.parents_);
+  }
+
  private:
+  static int UpdateTCGraph();
```
barath: Function comments for each new function.

core/module.h
```diff
@@ -137,7 +146,24 @@ class ModuleBuilder {
...
+  static int FindNextTC(const std::string &node, const std::string &parent,
+                        std::unordered_set<std::string> &visited);
+
+  static std::map<std::string, Node> task_graph_;
```
barath: Member variable comments.

core/module.h
```diff
@@ -318,6 +345,13 @@ class Module {
 
   virtual CheckConstraintResult CheckModuleConstraints() const;
 
+  virtual bool is_task() { return false; };  // Whether the module has a task.
```
barath: Declare method as const and rename as IsTask():

`virtual bool IsTask() const { return false; };`

core/module.h
```diff
@@ -318,6 +345,13 @@ class Module {
...
+
+  int overload() { return overload_; };  // For testing.
```
barath: Declare method as const.

core/module.h
```diff
@@ -318,6 +345,13 @@ class Module {
...
+
+  void SignalOverload();
```
barath: Method comments.

core/module.h
```diff
@@ -349,6 +383,8 @@ class Module {
   std::vector<const ModuleTask *> visited_tasks_;
 
  protected:
+  std::atomic<int> overload_;
```
barath: Member variable comment.

Also, perhaps rename to children_overload_ or something like that.

core/modules/queue.cc
```diff
@@ -77,6 +78,13 @@ CommandResponse Queue::Init(const bess::pb::QueueArg &arg) {
     prefetch_ = true;
   }
 
+  if (arg.backpressure()) {
+    backpressure_ = true;
+    high_water_ = (uint64_t) (size_ * kHighWaterRatio);
+    low_water_ = (uint64_t) (size_ * kLowWaterRatio);
+    underload_ = true;
```
barath: No need for underload_; just use overload_ = false here.

core/modules/queue.cc
```diff
@@ -101,6 +109,11 @@ std::string Queue::GetDesc() const {
 void Queue::ProcessBatch(bess::PacketBatch *batch) {
   int queued =
       llring_mp_enqueue_burst(queue_, (void **)batch->pkts(), batch->cnt());
+  if (backpressure_ && !overload_ && llring_count(queue_) > high_water_) {
+    SignalOverload();
+    overload_ = true;
```
barath: Might you have SignalOverload() know to set overload_ properly rather than doing it outside?

core/modules/queue.h
```diff
@@ -27,13 +27,26 @@ class Queue final : public Module {
...
   CommandResponse SetSize(uint64_t size);
 
-  struct llring *queue_;
   bool prefetch_;
+  bool backpressure_;
   int burst_;
+  uint64_t size_;  // Capacity of queue
+  uint64_t high_water_;
+  uint64_t low_water_;
+  bool underload_;  // Whether the queue is currently underloaded
```
barath: Eliminate underload_

core/modules/queue.cc
```diff
@@ -134,6 +153,12 @@ struct task_result Queue::RunTask(void *) {
       total_bytes += batch.pkts()[i]->total_len();
   }
 
+  if (backpressure_ && !underload_ && llring_count(queue_) < low_water_) {
+    SignalUnderload();
+    underload_ = true;
+    overload_ = false;
```
barath: Move the setting of overload_ into SignalUnderload()?

core/modules/queue.cc
```diff
@@ -77,6 +78,13 @@ CommandResponse Queue::Init(const bess::pb::QueueArg &arg) {
     prefetch_ = true;
   }
 
+  if (arg.backpressure()) {
+    backpressure_ = true;
+    high_water_ = (uint64_t) (size_ * kHighWaterRatio);
```
barath: I would move this setting of high_water_, low_water_ and underload_ to a separate function, because it will need to be called not just here but also when size is set.

vivi:
@barath changes made!

core/module.h
```diff
@@ -318,6 +384,23 @@ class Module {
 
   virtual CheckConstraintResult CheckModuleConstraints() const;
 
+  /*!
+   * Whether the module overrides RunTask.
+   */
+  virtual bool IsTask() const { return false; };
```
barath: I just realized another way of doing this without a virtual function is to store "is task" as a bool member in Module and then just have a non-virtual accessor and have it be false by default.  Then task-implementing modules can just set it to true in their constructors.

core/module.cc
```diff
@@ -402,7 +473,8 @@ int Module::ConnectModules(gate_idx_t ogate_idx, Module *m_next,
   ogate->AddHook(new Track());
   igate->PushOgate(ogate);
 
-  return 0;
+  // Update graph
+  return ModuleBuilder::AddEdge(name(), m_next->name());
```
barath: You're inside the class here, so no need to access via the accessor.

core/module.cc
```diff
@@ -512,6 +584,40 @@ void Module::RunSplit(const gate_idx_t *out_gates,
     RunChooseModule(pending[i], &batches[i]);
 }
 
+void Module::SignalOverload() {
+  if (overload_) {
+    return;
+  }
+
```
barath: Set `overload_ = true` here

core/module.cc
```diff
@@ -16,13 +16,80 @@
...
+  return ret;
+}
 
 Module *ModuleBuilder::CreateModule(const std::string &name,
                                     bess::metadata::Pipeline *pipeline) const {
   Module *m = module_generator_();
   m->set_name(name);
   m->set_module_builder(this);
   m->set_pipeline(pipeline);
+  module_graph_.emplace(std::piecewise_construct, std::forward_as_tuple(name),
```
barath: Move this adding of entries to AddModule

core/module.h
```diff
@@ -137,7 +165,43 @@ class ModuleBuilder {
...
+   */
+  static int FindNextTC(const std::string &node_name,
+                        const std::string &parent_name,
+                        std::unordered_set<std::string> &visited);
+
+  // A graph of all the modules in the current pipeline.
+  static std::map<std::string, Node> module_graph_;
+
+  // Represents a graph of all modules that are tasks in the pipeline.
+  static std::map<std::string, Node> task_graph_;
```
barath: Move the storage of this information to the modules themselves so that they can access their parents via a local member variable (a vector of parents)?

core/module.h
```diff
@@ -137,7 +165,43 @@ class ModuleBuilder {
 
   CommandResponse RunInit(Module *m, const google::protobuf::Any &arg) const;
 
+  /*!
+   * Connects two modules together in `module_graph_`.
+   */
+  static int AddEdge(const std::string &from, const std::string &to);
```
barath: If the return is just "success" or "failure" then prefer to use a bool and return true upon success.

core/module.cc
```diff
@@ -16,13 +16,80 @@
...
+  if (it == module_graph_.end()) {
+    return -EINVAL;
+  }
+  it->second.children().insert(to);
+  UpdateTCGraph();
+  return 0;
+}
+
+int ModuleBuilder::UpdateTCGraph() {
+  int ret = 0;
```
barath: Omit.

core/module.cc
```diff
@@ -16,13 +16,80 @@
...
+  for (auto const &tc : task_graph_) {
+    std::unordered_set<std::string> visited;
+    ret = FindNextTC(tc.first, tc.first, visited);
```
barath: Perhaps rewrite this and the next few lines as:

`
if (!FindNextTC(...)) {
  return false;
}
`

core/module.h
```diff
@@ -137,7 +165,43 @@ class ModuleBuilder {
...
+   */
+  static int UpdateTCGraph();
+
+  /*!
+   * Finds the next module that implements a task, and updates `task_graph_`
+   * accordingly.
+   */
+  static int FindNextTC(const std::string &node_name,
+                        const std::string &parent_name,
+                        std::unordered_set<std::string> &visited);
```
barath: Per the style guide, arguments that get modified should be passed as pointers rather than references.

vivi:
don't merge yet -- forgot to account for deleting/disconnecting individual modules

core/module.cc
```diff
@@ -84,17 +85,21 @@ Module *ModuleBuilder::CreateModule(const std::string &name,
...
 }
 
 bool ModuleBuilder::AddModule(Module *m) {
-  return all_modules_.insert({m->name(), m}).second;
+  if (m->is_task_) {
+    if (!tasks_.insert(m->name()).second) {
+      return false;
+    }
+  }
+  return all_modules_.insert({m->name(), m}).second &&
```
barath: I would break this up so it reads like a sentence.  E.g.:
`
bool module_added = all_modules_.insert({m->name(), m}).second;
bool graph_updated = ...
return module_added && graph_updated;
`

core/module.cc
```diff
@@ -474,7 +479,10 @@ int Module::ConnectModules(gate_idx_t ogate_idx, Module *m_next,
   igate->PushOgate(ogate);
 
   // Update graph
-  return ModuleBuilder::AddEdge(name(), m_next->name());
+  if (ModuleBuilder::AddEdge(name_, m_next->name_)) {
```
barath: You could just return !ModuleBuilder::AddEdge(...) if I am understanding the return value / logic right.

core/module.h
```diff
@@ -21,17 +22,18 @@ using bess::gate_idx_t;
...
-// Represents a node in a graph.
+// Represents a node in `module_graph_`.
 class Node {
  public:
-  /*!
-   * Creates a new Node that represents `module_`.
-   */
-  Node(Module *module) : module_(module), children_(), parents_() {}
+  // Creates a new Node that represents `module_`.
+  Node(Module *module) : module_(module), children_() {}
```
barath: Do you still need to store the module if the Node itself is held by a module?

core/module.h
```diff
@@ -385,22 +367,35 @@ class Module {
...
-   */
-  virtual bool IsTask() const { return false; };
+  // For testing.
+  int children_overload() const { return children_overload_; };
+  const std::vector<Module *> &parent_tasks() const { return parent_tasks_; };
 
-  int children_overload() { return children_overload_; };  // For testing.
+  // Signals to parent task(s) that module is overloaded.
+  void SignalOverload() {
+    if (overload_) {
```
barath: You may want to add a TODO somewhere that overload and underload signaling only are safe if the module that they are called from is *not* marked as thread safe, indicating that it cannot be invoked by more than one worker.  If more than one worker can call a module then these methods would have to be changed to be thread safe.

core/module.cc
```diff
@@ -16,6 +16,89 @@
 const Commands Module::cmds;
 
 std::map<std::string, Module *> ModuleBuilder::all_modules_;
+std::unordered_map<std::string, Node> ModuleBuilder::module_graph_;
+std::unordered_set<std::string> ModuleBuilder::tasks_;
+
+bool ModuleBuilder::AddEdge(const std::string &from, const std::string &to) {
+  auto it = module_graph_.find(from);
```
barath: Do you also need to check whether `module_graph_` has `to`?

core/module.cc
```diff
@@ -16,6 +16,89 @@
...
+  for (auto const &task : tasks_) {
+    auto it = all_modules_.find(task);
+    if (it != all_modules_.end()) {
+      it->second->parent_tasks_.clear();
+    }
+  }
+  return UpdateTCGraph();
+}
+
+bool ModuleBuilder::UpdateTCGraph() {
```
barath: Perhaps rename "TC" to "Task".

core/module.cc
```diff
@@ -16,6 +16,89 @@
...
+  for (auto const &task : tasks_) {
+    std::unordered_set<std::string> visited;
+    if (!FindNextTC(task, task, &visited)) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool ModuleBuilder::FindNextTC(const std::string &node_name,
```
barath: Rename "TC" to "Task" here as well.

core/module.cc
```diff
@@ -16,6 +16,89 @@
...
+    return true;
+  }
+
+  auto node_it = module_graph_.find(node_name);
+  if (node_it == module_graph_.end()) {
+    return false;
+  }
+
+  for (auto &child_name : node_it->second.children()) {
+    auto it = visited->find(child_name);
```
barath: I think here you can use `if (visited->count(child_name))`...

core/module.cc
```diff
@@ -27,7 +110,18 @@ Module *ModuleBuilder::CreateModule(const std::string &name,
...
 
 bool ModuleBuilder::AddModule(Module *m) {
-  return all_modules_.insert({m->name(), m}).second;
+  if (m->is_task_) {
+    if (!tasks_.insert(m->name()).second) {
+      return false;
+    }
+  }
+  bool module_added = all_modules_.insert({m->name(), m}).second;
+  bool node_added =
```
barath: Need to check that module was added before adding node, e.g.:
`
if (!module_added) {
  return false;
}
`

core/module.cc
```diff
@@ -402,7 +501,8 @@ int Module::ConnectModules(gate_idx_t ogate_idx, Module *m_next,
   ogate->AddHook(new Track());
   igate->PushOgate(ogate);
 
-  return 0;
+  // Update graph
+  return !(ModuleBuilder::AddEdge(name_, m_next->name_));
```
barath: No need for the extra parens.

core/module.h
```diff
@@ -20,6 +22,38 @@ using bess::gate_idx_t;
...
+  Node(Module *module) : module_(module), children_() {}
+
+  // Add a child to the node.
+  bool AddChild(const std::string &child) {
+    return children_.insert(child).second;
+  }
+
+  // Remove a child from the node.
+  void RemoveChild(const std::string &child) {
+    auto it = children_.find(child);
```
barath: Couldn't you just do `children_.erase(child)`?

vivi: [std::vector::erase](http://en.cppreference.com/w/cpp/container/vector/erase) takes an iterator as input

vivi: whoops, wrong part in the code. changed

barath:
LGTM.

barath:
@sangjinhan, did you want to take a quick look, or can we remove the notready label and merge?

barath:
LGTM.  Will wait for Travis.

barath:
image

PR #506, opened 2017-06-08
Fix raw_input and input confusion

changlan:
Fix the `bessctl` crash in Python 2 introduced by the previous PR.


sangjinhan:
tumblr_m9uf9qfm3k1r4u63lo1_400

PR #507, opened 2017-06-08
Update DPDK version to 17.05

sangjinhan:


core/packet.h
```diff
@@ -232,108 +242,86 @@ class alignas(64) Packet {
...
+          uint16_t nb_segs_;  // Number of segments
 
-      uint64_t offload_flags_;  // Offload features.
-
-      // remaining bytes are set on RX when pulling packet from descriptor
-      MARKER rx_descriptor_fields1_;
+          // offset 22:
+          uint16_t _dummy0_;  // rte_mbuf.port
+          // offset 24:
+          uint64_t _dummy1_;  // rte_mbuf.ol_flags
```
apanda: Why mark this as a dummy? Is the intention that we don't allow the use of any offloads? (My reading of this [example](http://dpdk.org/doc/api/examples_2tep_termination_2vxlan_8c-example.html#a49) and other was that this is just a place to request particular offloads.)

sangjinhan: Yes, it was meant to be "do not access". Given that PMD drivers support a variety set of different offloading features (and most vectorized PMDs do not support any), I thought it would be safe to keep it unset for Packet<->rte_mbuf compatibility.

apanda:
Merged, thanks!

image

PR #510, opened 2017-06-09
Optimize RunModuleSplit() and reduce TLS usage

ddiproietto:
My previous attempt at this caused random crashes inside gRPC on travis.

I haven't been able to understand why, or to reproduce the crash locally, but since then we changed some gRPC code.

I've run this on travis 10 times and it hasn't crashed.


sangjinhan:
x3rmp1hjookbxb9u61h2l1qfo1_500

PR #513, opened 2017-06-13
Fix attr_match.bess to use correct arg

vivi:
side note: it's a bit confusing for ExactMatch to use `attribute` and Metadata to use `name`


sangjinhan:
4la9smemungvhc5xotw1zpnoo1_400

PR #514, opened 2017-06-13
Fix metadata endianness

sangjinhan:
While all metadata attributes are supposed to be stored in big endian, `SetMetadata` and `WildcardMatch` modules were incorrectly assuming values are stored in host order.

(module tests will be added for a new test framework)


changlan:
image

PR #519, opened 2017-06-13
Upload release binaries from CI

changlan:
Upload the GCC 6 built binaries (as well as Python scripts) whenever a tag is made.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,5 +64,28 @@ after_failure:
   - more /tmp/bessd.*.INFO.* | cat      # dump all bessd log files
   - more /tmp/sanitizer* | cat          # dump all sanitizer results
 
+before_deploy:
+  - find . -name '*.pyc' -delete
+  - for arch in x86-64 sandybridge haswell; do
+      export CXXARCHFLAGS="-march=$arch";
+      ./container_build.py clean;
+      ./container_build.py bess;
+      ./container_build.py kmod_buildtest;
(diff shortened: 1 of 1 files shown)


sangjinhan:
BESS daemon is built with `g++ -march=native`, so the resulting binary will require same machines instructions from the machine on which it was built. I guess Travis usually runs on GCE with AVX2, but the binary won't run on any pre-Haswell machines...

sangjinhan:
Perhaps in the long-term, we should support binary compatibility with Function Multi-Versioning.
https://lwn.net/Articles/691932/

changlan:
How about building with multiple `march` for legacy architectures? How many architectures we want to support?

sangjinhan:
SSSE3 (minimum HW requirement), AVX, and AVX2 should be sufficient I guess...?

SSSE3 was the lowest spec (Core 2 architecture) we saw in the field. I think most servers are now equipped with AVX2, but there still are lots of Sandy/IvyBridge servers our there only with AVX...

changlan:
Done. Currently it builds for three `-march` types:
- Generic `x86_64`
- `sandybridge`
- `haswell`

changlan:
Sure. Can I update the instruction after this PR is merged to master branch? It is easier for me to generate release binary and test dependencies.

melvinw:
Thanks! ecqhlpw

PR #522, opened 2017-06-15
Explicitly define kMaxBurst

sirspinach:
This is required for proper linking. Previously we were able to get away only with declaration (but without definition) of kMaxBurst, because a compiler optimization ignores this issue when kMaxBurst is used for array sizing.

### DIFF
--- a/core/packet.cc
+++ b/core/packet.cc
@@ -15,6 +15,8 @@
 
 namespace bess {
 
+const size_t PacketBatch::kMaxBurst;
+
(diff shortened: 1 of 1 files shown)


core/packet.cc
```diff
@@ -15,6 +15,8 @@
 
 namespace bess {
 
+const size_t PacketBatch::kMaxBurst;
```
changlan: It is recommended to also initialize the variable in the source file, rather than in the class.

changlan:
image

PR #526, opened 2017-06-16
GCC 7 Support

changlan:
To support newer Ubuntu versions we need to make sure BESS builds with GCC 7 correctly:

- Use `sort -V` to compare `CXX` versions in Makefile
- Explicitly mark fall-throughs
- Misc. fixes


core/Makefile
```diff
@@ -18,10 +18,9 @@ $(shell mkdir -p $(DEPDIR)/hooks >/dev/null)
 # 'clang' or 'g++'
 CXXCOMPILER := $(shell expr $(word 1, $(shell $(CXX) --version)) : '\(clang\|g++\)')
 
-# e.g., 4.9.3 -> 40903. For clang it is always 40201
-CXXVERSION := $(shell $(CXX) -dumpversion | sed -e 's/\.\([0-9][0-9]\)/\1/g' -e 's/\.\([0-9]\)/0\1/g' -e 's/^[0-9]\{3,4\}$$/&00/')
+CXXVERSION := $(shell $(CXX) -dumpversion)
 
-ifeq "$(shell expr $(CXXCOMPILER) = g++ \& $(CXXVERSION) \< 40800)" "1"
+ifneq "$(shell echo -e '$(CXXVERSION)\n4.7' | sort -V | head -n1)" "4.7"
```
sangjinhan: I don't understand how it works. `clang -dumpversion` is supposed to report `4.2.1` no matter what its real version is (this is why the previous code checked whether it is `g++`) first. So in theory build with clang should fail, but somehow it passed CI?

core/utils/cuckoo_map.h
```diff
@@ -180,7 +180,7 @@ class CuckooMap {
   Entry* Find(const K& key, const H& hasher = H(), const E& eq = E()) {
     // Blame Effective C++ for this
     return const_cast<Entry*>(
-        static_cast<const typeof(*this)&>(*this).Find(key, hasher, eq));
+        static_cast<const CuckooMap&>(*this).Find(key, hasher, eq));
```
sangjinhan: Is it safe to do this? If a `DerivedClass` redefines (not overrides) the const version of `Find`, then it should invoke `DerivedClass::Find()`, not `CuckooMap::Find()`

env/Dockerfile.m4
```diff
@@ -21,7 +21,7 @@ RUN cd /build && \
 	git clone -b ${BESS_DPDK_BRANCH} https://github.com/netsys/bess && \
 	cd /build/bess && \
 	setarch ${DPDK_ARCH} ./build.py dpdk && \
-	mv /build/bess/deps/dpdk-17.02 /build/dpdk-17.02 && \
+	mv /build/bess/deps/dpdk-17.05 /build/dpdk-17.05 && \
```
sangjinhan: 👍

core/utils/cuckoo_map.h
changlan: I'm not entirely sure how `typeof` works, but I guess it just statically obtains the type name of its argument? In this case, the type of `*this` would be `CuckooMap` anyway?

core/Makefile
changlan: You are right. This is indeed interesting..

core/utils/cuckoo_map.h
chris3torek: Everything I found says that `typeof` works the way changlan says. You'd have to provide a template that derived classes could instantiate to work around that, I think.

sangjinhan: Oh I see. Perhaps I chose it because it's easier to copy pasta...

core/Makefile
changlan: It turns out that `shell` didn't handle `echo -e` correctly. Now using `printf`.

core/utils/cuckoo_map.h
changlan: Here you go: a copy-pastable and C++11-compliant version (looks a bit ugly though ;)

chris3torek: I found that `std::remove_reference` trick when I was looking at this last night. Does the same as the typeof.  Seems good.  I agree it's uglier, but oh well, that's C++11 for you!

env/packages.yml
```diff
@@ -71,6 +71,9 @@
 
     - name: Compile gRPC and its dependencies
       shell: make -j{{ ansible_processor_vcpus }} chdir=/tmp/grpc
+      environment:
+        - CC: gcc-6
+        - CXX: g++-6
```
sangjinhan: Does it link correctly to BESS, which is built with a compiler other than g++-6? I vaguely remember g++ 4.8 has a different ABI, causing link issues to objects with newer ABIs.

Maybe it's okay for libraries to be built with newer ABIs than the program. Maybe the other way around...

changlan: gRPC in the current docker image is built by g++ 6. Since the CI builds are fine with that, it should be okay to use g++ 6.

sangjinhan:
74561c1bcf64886d70cec9b145cf046c

PR #527, opened 2017-06-19
Update dpdk-devbind.py to DPDK 17.05

sangjinhan:
It seems that 2to3 somehow broke this script:
```
Traceback (most recent call last):
  File "./dpdk-devbind.py", line 36, in <module>
    from builtins import map
ImportError: No module named builtins
```

This PR updates the script with the one in DPDK 17.05 (the previous one was 17.02). The script runs well on both Python 2 and 3 out of the box.


changlan:
image

PR #529, opened 2017-06-20
Add vhost-user sample script

sangjinhan:
This sample script illustrates how vhost-user ports (`PMDPort(vdev='eth_vhost_X', iface=/path/to/Y.sock')`) can be used for VMs with virtio-net devices. It also includes a VM image generator.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,7 @@ coverage.info
 coverage_html
 .coverage
 htmlcov
+
+# VM images and SSH keys
+vm.key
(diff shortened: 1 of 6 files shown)


sangjinhan:
Those are semantically identical. BESS scripts (`*.bess`) support some syntactic sugars, such as `foo -> bar`. Nowadays I am a bit skeptical if it is worth it, though... :(

changlan:
image

PR #530, opened 2017-06-20
Use __attribute__ instead of square backets

changlan:
Unfortunately C does not support square bracket `[[ xxx ]]` for attributes, which
breaks kernel module builds. Falling back to the old `__attribute__` style.

### DIFF
--- a/core/kmod/llring.h
+++ b/core/kmod/llring.h
@@ -66,15 +66,22 @@
 #ifndef _LLRING_H_
 #define _LLRING_H_
 
-#if defined(__has_cpp_attribute)
+#if !defined(FALLTHROUGH) && defined(__has_attribute)
+#if __has_attribute(fallthrough)
+#define FALLTHROUGH __attribute__((fallthrough))
+#endif
+#endif
+
+#if !defined(FALLTHROUGH) && defined(__has_cpp_attribute) &&                   \
+    defined(__cplusplus)
 #if __has_cpp_attribute(fallthrough)
 #define FALLTHROUGH [[fallthrough]]
 #elif __has_cpp_attribute(clang::fallthrough)
 #define FALLTHROUGH [[clang::fallthrough]]
-#else
-#define FALLTHROUGH
 #endif
-#else
+#endif
+
+#ifndef FALLTHROUGH
 #define FALLTHROUGH
 #endif
 
--- a/core/utils/copy.h
+++ b/core/utils/copy.h
@@ -1,15 +1,16 @@
 #ifndef BESS_UTILS_COPY_H_
 #define BESS_UTILS_COPY_H_
 
-#if defined(__has_cpp_attribute)
+#if !defined(FALLTHROUGH) && defined(__has_cpp_attribute) && \
+    defined(__cplusplus)
 #if __has_cpp_attribute(fallthrough)
 #define FALLTHROUGH [[fallthrough]]
 #elif __has_cpp_attribute(clang::fallthrough)
 #define FALLTHROUGH [[clang::fallthrough]]
-#else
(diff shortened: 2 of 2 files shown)


sangjinhan:
Is `copy.h` used by non-C++ code somewhere?

chris3torek:
Yes, my test build on my VM just failed:
```make[1]: Entering directory '/usr/src/linux-headers-4.4.0-66-generic'
Building with KERNELRELEASE = 4.4.0-66-generic
  CC [M]  /home/vagrant/bess/core/kmod/sndrv.o
In file included from /home/vagrant/bess/core/kmod/sn_common.h:54:0,
                 from /home/vagrant/bess/core/kmod/sndrv.c:32:
/home/vagrant/bess/core/kmod/llring.h:72:32: error: missing ')' after "__has_attribute"
 #elif __has_cpp_attribute(clang::fallthrough)
                                ^
/home/vagrant/bess/core/kmod/llring.h:72:33: error:  ':' without preceding '?'
 #elif __has_cpp_attribute(clang::fallthrough)
                                 ^
scripts/Makefile.build:258: recipe for target '/home/vagrant/bess/core/kmod/sndrv.o' failed```

**Comment by chris3torek** [comment]
Oh, duh, that's just llring.h not copy.h. Not sure if copy.h is used elsewhere, will find out.

**Comment by chris3torek** [comment]
Pretty sure copy.h is not used elsewhere.  Unfortunately all the `#if` test variants fail to pass gcc at all, it just cannot handle the `#if` expression with the double colon.

This pattern seems to work:
```#if defined(__has_attribute) && __has_attribute(fallthrough)
#define FALLTHROUGH __attribute__((fallthrough))
#endif

#if !defined(FALLTHROUGH) && defined(__has_cpp_attribute) && \
    __has_cpp_attribute(fallthrough)
#define FALLTHROUGH [[fallthrough]]
#endif

#if !defined(FALLTHROUGH) && defined(__has_cpp_attribute) && \
    defined(__clang__)
#if __has_cpp_attribute(clang::fallthrough)
#define FALLTHROUGH [[clang::fallthrough]]
#endif
#endif

#ifndef FALLTHROUGH
#define FALLTHROUGH
#endif```

**Comment by changlan** [comment]
GCC 4.8 complains about this line: 
```
utils/copy.h:4:48: error: missing binary operator before token "("
 #if defined(__has_attribute) && __has_attribute(fallthrough)
                                                ^
utils/copy.h:9:24: error: missing binary operator before token "("
     __has_cpp_attribute(fallthrough)
```

Probably it is because GCC 4.8 macro preprocessor doesn't recognize `__has_attribute`. Breaking them into two `if` statements works fine.

chris3torek:
Ah, so the copy.h problem is not C vs C++ but rather gcc preprocessor vs clang preprocessor. I can't get llring.h and the kmod C code to build just with the separate `#if` lines though: I needed an explicit `__clang__` test, like the one I put in here.

changlan:
Yes. They are two different issues:

- The `llring.h` problem is about C vs C++: square brackets attributes `[[ xxx ]] ` are invalid in C. They are only included after C++11. 
- The `copy.h` error only occurs with GCC 4.8, as it does not have some builtin macros, such as`__has_attribute` and `__has_cpp_attribute` ...

I updated the patch. It should address those two problems.

chris3torek:
Unfortunately gcc 5.4.1 defines `__has_attribute`. So the llring.h file needs more than just a test for `__has_attribute`, this is where I had to test `__clang__` as well.

chris3torek:
I don't know why codecov is so unhappy about the ether_encap and bpf code but this is working for me now. 👍

sangjinhan:
default-10

PR #531, opened 2017-06-22
Update README and minor tweaks

changlan:
- Update the instruction use binary releases.
- Include the parent directory `bess` in the tarball
- Delete `.pyc` files before tarballing
- Include kmod source code


sangjinhan:
62269c9bb6312f35247579865d895a33

PR #532, opened 2017-06-22
env: Install clang-3.9

ddiproietto:
It's useful to have a more recent version.  For now travis still uses 3.8,
but that may change soon.

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -17,6 +17,7 @@
         - g++-7
         - g++
         - clang-3.8
+        - clang-3.9
         - autoconf
         - automake
         - libtool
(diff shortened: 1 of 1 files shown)


changlan:
Merged, thanks!
image

sangjinhan:
With gRPC (and protobuf) built with g++ 4.8, the ansible script fails on a clean Xenial server. Could you check please?

changlan:
I tried different combinations of `CXX` to build gRPC and BESS on `xenial`:

| BESS \ gRPC 	| 4.8 	| 5 	| 6 	| 7 	|
|:-----------:	|:---:	|:-:	|:-:	|:-:	|
| 4.8 	| x 	| x 	| x 	| x 	|
| 5 	| x 	| v 	| x 	| x 	|
| 6 	| x 	| v 	| v 	| x 	|
| 7 	| x 	| v 	| v 	| x 	|

It seems that 4.8 is not working well here. I wonder if we should stop supporting GCC 4.8.

PR #535, opened 2017-06-28
Remove GCC 4.8 support

changlan:
GCC 4.8 uses a different ABI. As a result, gRPC built by GCC 4.8 doesn't work in any case (#532). In addition, if gRPC is built by GCC 5+, BESS cannot be built by GCC 4.8. This PR ditches the support for GCC 4.8.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -21,8 +21,8 @@ CXXCOMPILER := $(shell expr $(word 1, $(shell $(CXX) --version)) : '\(clang\|g++
 CXXVERSION := $(shell $(CXX) -dumpversion)
 
 ifeq "$(CXXCOMPILER)" "g++"
-ifneq "$(shell printf '$(CXXVERSION)\n4.7' | sort -V | head -n1)" "4.7"
-    $(error g++ 4.8 or higher is required)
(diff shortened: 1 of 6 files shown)


core/Makefile
```diff
@@ -21,8 +21,8 @@ CXXCOMPILER := $(shell expr $(word 1, $(shell $(CXX) --version)) : '\(clang\|g++
 CXXVERSION := $(shell $(CXX) -dumpversion)
 
 ifeq "$(CXXCOMPILER)" "g++"
-ifneq "$(shell printf '$(CXXVERSION)\n4.7' | sort -V | head -n1)" "4.7"
-    $(error g++ 4.8 or higher is required)
+ifneq "$(shell printf '$(CXXVERSION)\n5' | sort -V | head -n1)" "5"
+    $(error g++ 5 or higher is required)
```
sangjinhan: We can recommend `container_build.py` here..?

changlan: Sure.

changlan: Done

barath:
photo on 2014-08-21 at 18 39 2

PR #536, opened 2017-06-29
Add Env variable to support running of only one python test

GalSagie:
Example use case:
run testing/run_module_tests TEST_NAME='nat'

This will only run nat.py test

### DIFF
--- a/bessctl/conf/testing/run_module_tests.bess
+++ b/bessctl/conf/testing/run_module_tests.bess
@@ -9,7 +9,7 @@ import traceback
 from time import gmtime, strftime
 
 # Env variable used to run a single test
-test_name = $TEST_NAME!'none'
+test_name = $TEST_NAME!'*'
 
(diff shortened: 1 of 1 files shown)


bessctl/conf/testing/run_module_tests.bess
```diff
@@ -8,6 +8,9 @@ import sugar
 import traceback
 from time import gmtime, strftime
 
+# Env variable used to run a single test
+test_name = $TEST_NAME!'none'
+
```
apanda: I wonder if it won't be better to just set the default value to `*`, so something like

```
test_name = $TEST_NAME!'*'
```

This has two benefits: one it saves the if block below; two it allows testa named `none.py`.

GalSagie: Good idea, will do!

apanda:
Merged. Thanks!

PR #538, opened 2017-06-29
Add MPLS pop module

GalSagie:
Still TODO:
- Make pop more efficient using AVX instructions
- Save label as metadata for next module
- Currently ignores MPLS stacking


core/modules/mpls_pop.cc
```diff
@@ -0,0 +1,41 @@
...
+  int cnt = batch->cnt();
+
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+
+    Ethernet *eth = pkt->head_data<Ethernet *>();
+
+    if (eth->ether_type != be16_t(Ethernet::Type::kMpls)) {
+      // Currently ignore non MPLS packets
+      continue;
```
barath: It seems a bit odd to me to have the output gate have a mix of processed MPLS packets and unprocessed other packets.  I can see the use case, but wonder if it might be better to have two output gates, one for processed MPLS packets and another for the rest.

core/modules/mpls_pop.cc
```diff
@@ -0,0 +1,41 @@
...
+
+    // TODO(gsagie) convert this to be more efficient using Intel instructions
+    Ethernet::Address src_addr = eth->src_addr;
+    Ethernet::Address dst_addr = eth->dst_addr;
+
+    pkt->adj(4);
+    Ethernet *eth_new = pkt->head_data<Ethernet *>();
+    eth_new->src_addr = src_addr;
+    eth_new->dst_addr = dst_addr;
+    eth_new->ether_type = be16_t(Ethernet::Type::kIpv4);
```
nrdmao: Isn't it rather presumptive to assume that: 1. The MPLS payload is IPv4 and 2. that the label you are popping is indeed the bottom of the label stack? This being said, I am not really sure what use case you are envisioning for this module. Typically, in an MPLS switch, the MPLS label determines the MPLS operation (pop, swap, push) and then subsequent L2 (proto, src/dst mac) next-hop rewrite.

GalSagie: Regarding 2, i have added a TODO and plan to address it with the bottom of stack flag.
Regarding 1, what would you suggest to do here? i have to assume that the next or previous module classified it and know what to do, so maybe the easiest will be to have the next ether type proto be configurable?

What do you think?

core/modules/mpls_pop.cc
GalSagie: This sounds reasonable, i will fix

core/modules/mpls_pop.cc
nrdmao: Yes, probably configurable would be good. Config should allow to either specify the ethertype or, if bottom of stack, completely remove the ethernet encapsulation and expose the entire payload. The MPLS control plane knows what is going on. For example, if this label represents a pseudo wire, the right thing to do is to expose the entire MPLS payload as it represents a fully formed L2 packet.

GalSagie: Ok, that make sense in a generic way (less relevant to what i had in mind) but i will address this and resend this patch, thanks for the review!

nrdmao: Sounds good. What did you have in mind if you don't mind my asking?

core/modules/mpls_pop.cc
```diff
@@ -0,0 +1,60 @@
...
+  int cnt = batch->cnt();
+
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+
+    Ethernet *eth = pkt->head_data<Ethernet *>();
+
+    if (eth->ether_type != be16_t(Ethernet::Type::kMpls)) {
+      // non MPLS packets are sent to different output gate
+      out_gates[i] = 1;
```
sangjinhan: `continue;` here?

bessctl/conf/samples/mpls_test.bess
```diff
@@ -0,0 +1,28 @@
+import scapy.all as scapy
+import random
```
sangjinhan: not used?

bessctl/conf/samples/mpls_test.bess
```diff
@@ -0,0 +1,28 @@
...
+import string
```
sangjinhan: not used?

core/modules/mpls_pop.cc
GalSagie: Good catch, thanks!

bessctl/conf/samples/mpls_test.bess
GalSagie: Removed

bessctl/conf/samples/mpls_test.bess
GalSagie: Removed

GalSagie:
Fixed formatting problems

sangjinhan:
Merged, thanks.

9aecc432dfa9d7725311fa53c2b63313

PR #539, opened 2017-06-29
ansible: Install protoc before gRPC.

sirspinach:
On a clean Ubuntu Xenial server, the Ansible playbook fails on the
"Compile gRPC" step because protoc cannot be found. This patch compiles
and install protoc first so that gRPC compilation can proceed.


sirspinach:
not sure if this is enough to fix the build, please don't push yet.

sangjinhan:
If protobuf is not available, gRPC is supposed to build it and install automatically. If it doesn't do it properly, I think that's the issue we need to fix...

sirspinach:
@chris3torek found a solution!

PR #541, opened 2017-06-30
Add virtio-user example for BESS<->container connection with kernel bypass

sangjinhan:
Same as #529, but for containers this time.

To allow BESS and containers (both DPDK applications) to run in the same domain, `--no-shconf` and `--huge-unlink` options to the EAL options of BESS.

### DIFF
--- a/bessctl/conf/port/vhost/launch_container.py
+++ b/bessctl/conf/port/vhost/launch_container.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+
+import os
+import sys
+import subprocess
+import time
+import shlex
+
+# How many cores we reserve for vSwitches?
+# If set to 2, Containers will run on core 2, 3, 4, ..., skipping core 0-1.
+VM_START_CPU = int(os.getenv('VM_START_CPU', '1'))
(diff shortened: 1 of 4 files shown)


bessctl/conf/port/vhost/launch_container.py
```diff
@@ -0,0 +1,99 @@
...
+    first_cpu = VM_START_CPU + cid * NUM_VCPUS
+    last_cpu = first_cpu + NUM_VCPUS - 1
+    eal_opts = '--file-prefix=vnf{} --no-pci -m 256 -l 0,{}-{}'.format(
+        cid, first_cpu, last_cpu)
+
+    for port_id in range(NUM_VPORTS):
+        sockpath = '{}/vhost_user{}_{}.sock'.format(SOCKDIR, cid, port_id)
+        eal_opts += ' --vdev=virtio_user{},path={}'.format(port_id, sockpath)
+
+    testpmd_opts = '-i --burst=64 --txd=1024 --rxd=1024 --disable-hw-vlan ' \
```
chris3torek: trivial: might be nice to share some of this code with the launch_vm.py code (though it looks like only the testpmd setup itself is the same).

Issue #534, opened 2017-06-27
Scheduler diverges at high rate limits


krfall:
Using a simple src->sink topology with a large rate limit works properly up to some limit, and then then hits a point where it no longer tracks properly.

More specifically, on 2.6GHz cpu, up to a 68Gb/s RL is fine but a setting of 69Gb/s gives about 280Mb/s.

Heres is some "show tc" output for these two cases, respectively:
```
20:36:10.627260          CPU MHz   scheduled        Mpps        Mbps  pkts/sched    cycles/p
--------------------------------------------------------------------------------------------
W0 mylimit              1164.117     3160844     101.147   67970.804      32.000      11.509
W0 !leaf_src1:0         1164.115     3160841     101.147   67970.730      32.000      11.509
--------------------------------------------------------------------------------------------

20:36:29.158229          CPU MHz   scheduled        Mpps        Mbps  pkts/sched    cycles/p
--------------------------------------------------------------------------------------------
W0 mylimit                 9.295       13035       0.417     280.325      32.000      22.282
W0 !leaf_src1:0            9.296       13036       0.417     280.332      32.000      22.283
--------------------------------------------------------------------------------------------
```

ddiproietto:
Maybe this line is the problem?

https://github.com/NetSys/bess/blob/7b5d7a7cd2611ea61f6072b1dbdeb43fb81caeb6/core/traffic_class.h#L504`

`x` is the user input.  If I'm not mistaken, the left shift by `USAGE_AMPLIFIER_POW - 4`, (28) will cause anything greater than `2**36` (68.7G) to overflow

sangjinhan:
Fixed by #542

PR #542, opened 2017-06-30
Fix ratelimit overflow

sangjinhan:
#534 is caused by `uint64_t` overflow while converting resource limit in resource/s to work/cycle, if the input is greater than 2^36/s. This PR is a hot fix to increase the threshold to 2^42/s (e.g., 4.4 Tbps), which should be practically good for now.

### DIFF
--- a/core/traffic_class.h
+++ b/core/traffic_class.h
@@ -497,12 +497,10 @@ class RateLimitTrafficClass final : public TrafficClass {
 
   void TraverseChildren(std::function<void(TCChildArgs *)>) const override;
 
-  // Convert resource units to work units per cycle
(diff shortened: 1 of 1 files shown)


barath:
Does this affect low end rates in any way?

sangjinhan:
No, it should not (in any visible way).

I realized I can just use 128-bit data types to avoid overflow. I'll push a new commit.

sangjinhan:
Done.

changlan:
image

changlan:
`__int128` doesn't exist in 32-bit GCC, causing the 32-bit to fail..

PR #545, opened 2017-07-01
Remove invalid apt-get package in ansible-playbook

sirspinach:
The protoc package doesn't exist, and so the `apt-get remove`
operation that was supposed to wipe previous protobuf installs failed.

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -70,7 +70,7 @@
       git: repo=https://github.com/google/grpc dest=/tmp/grpc accept_hostkey=yes version=v1.3.2
 
     - name: Compile gRPC and its dependencies
-      shell: make -j{{ ansible_processor_vcpus }} chdir=/tmp/grpc
+      shell: make -j{{ ansible_processor_vcpus }} HAS_SYSTEM_PROTOBUF=false chdir=/tmp/grpc
       environment:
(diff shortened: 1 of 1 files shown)


env/packages.yml
```diff
@@ -59,7 +59,7 @@
       become: true
 
     - name: Wipe prior protobuf installs to avoid conflicts
-      shell: apt-get remove -y -f libprotobuf* protobuf-* protoc; sudo rm -f `which protoc` warn=no
+      shell: apt-get remove -y -f libprotobuf* protobuf-*; sudo rm -f `which protoc` warn=no
```
GalSagie: Minor and not related but i think the sudo can be removed

sirspinach: that's right. Thanks!

chris3torek:
I don't know what's happening with the _32 travis builds (one succeeded, one built but died during test, one complained about unauthenticated package) but the changes themselves look good to me. I take it the HAS_SYSTEM_PROTOBUF=false is either necessary or just seems like a good idea

sirspinach:
@chris3torek: HAS_SYSTEM_PROTOBUF=false forces protobuf install in case we missed something or protobuf libs are broken.

justinemarie:
image

PR #548, opened 2017-07-02
Improve MPLS header accessibility

GalSagie:
Make it easier to manipulate and parse MPLS headers

### DIFF
--- a/core/utils/mpls.h
+++ b/core/utils/mpls.h
@@ -1,6 +1,6 @@
 #ifndef BESS_UTILS_MPLS_H_
 #define BESS_UTILS_MPLS_H_
-
+#include <type_traits>
 #include "rte_byteorder.h"
 
 namespace bess {
@@ -58,6 +58,7 @@ struct[[gnu::packed]] Mpls {
   be32_t entry;
 };
 
+static_assert(std::is_pod<Mpls>::value, "not a POD type");
 static_assert(sizeof(Mpls) == 4, "struct Mpls size is incorrect");
 
 }  // namespace utils
--- a/core/utils/mpls.h
+++ b/core/utils/mpls.h
@@ -1,7 +1,7 @@
 #ifndef BESS_UTILS_MPLS_H_
 #define BESS_UTILS_MPLS_H_
-#include <type_traits>
 #include "rte_byteorder.h"
+#include <type_traits>
 
 namespace bess {
 namespace utils {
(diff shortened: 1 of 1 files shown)


core/utils/mpls.h
```diff
@@ -19,18 +19,43 @@ namespace utils {
...
+
+  bool isBottomOfStack() {
+    be32_t mpls_entry = be32_t(entry);
+    return (mpls_entry.value() & MPLS_LS_S_MASK) >> MPLS_LS_S_SHIFT;
+  }
+
+  be32_t entry;
 };
 
 static_assert(sizeof(Mpls) == 4, "struct Mpls size is incorrect");
```
changlan: Would be good to add `static_assert(std::is_pod< Mpls >::value, "not a POD type");`

sangjinhan:
I don't quite understand the change. Aren't bit fields much easier to use?

GalSagie:
I thought bit fields were easier as well, but when i started working on mpls push module, it was harder to add and manipulate with the endiannes problems.

I feel this is more C++...

core/utils/mpls.h
GalSagie: will do

sangjinhan:
How about having a union here, like (untested):
```
union {
  struct {
#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__  
    uint32_t ttl: 8;
    uint32_t bottom_of_stack: 1;
    uint32_t tc: 3;
    uint32_t _label: 20;
...
#endif
  };

  be32_t tag;
}
```
Then only label() and set_label() would be necessary.

Also, please consider replace `MPLS_LS_*` macros with class static constants `kLabelMask`, and so on.

GalSagie:
Ok, addressed all comments.

sangjinhan:
Sorry, a few last nits are below. You don't need to agree with everything...
* You don't need `getTtl()`, `getTc()`, and `isBottomOfStack()` any longer, as they are directly accessible.
  * Accordingly, the constants (except `*Label*` ones) can be removed.
  * `setEntry()`: Is it still necessary?
* The only required field that requires accessor is label, which is nontrivial due to endianness:
  * Please consider rename `label` to `_label`, to indicate the field is not meant to be directly accessible.
  * `getLabel()` should be renamed to either `label()` or `Label()`, as per the Google style guide.
  * A corresponding mutator would be necessary, named either (`set_label()` or `SetLabel()`).
* Please consider renaming `s` to `bottom_of_stack`.
* I personally feel that the comments for the struct fields are somewhat redundant. Number of bits are already in the code, and the field names are self-contained or well-known. They repeat twice for both endians, making the code slightly verbose.
* The body of `label()` can be simply `return (tag.value() & kMplsLabelMask) >> kMplsLabelShift;`, since `tag` is already `be32_t`.

GalSagie:
Thanks for the comments and no need to be sorry :) 
I think they all make perfect sense and will address them soon

GalSagie:
Sorry, i tried this but found that the big/little static endianes compilation is also problematic, reverted to previous work.
Feel free to edit this after it is merged but i found out its more convenient to use this way.

sangjinhan:
f3bd8497e15399201b634714ec5ed390

PR #549, opened 2017-07-03
Make big endian types hashable

sirspinach:


core/utils/endian.h
```diff
@@ -153,4 +153,15 @@ static_assert(be32_t(0x1234).value() == 0x1234, "Something is wrong");
 }  // namespace utils
 }  // namespace bess
 
+namespace std {
+
+template <class T>
```
changlan: nit: `typename` instead of `class`

sirspinach: fixed

changlan:
image

PR #551, opened 2017-07-03
Standardized Field types

justinemarie:
Creates generic "Field" and "FieldData" types to represent data to be written into packets. This addresses inconsistent naming and types across modules that essentially do the same thing (how many "Field" types do we need? :-) and also enables python scripts to supply either byte strings OR integers to insert into any fields (which was inconsistently supported previously). PR also amends WildcardMatch, GenericEncap, and ExactMatch to make use of these new standardized Field types.

RandomUpdate should be extended to do so as well, but also needs to be extended to randomly update metadata in addition to packet-data -- this can be addressed in a later PR.

### DIFF
--- a/bessctl/conf/samples/exactmatch.bess
+++ b/bessctl/conf/samples/exactmatch.bess
@@ -24,7 +24,8 @@ packets = [gen_packet(scapy.UDP, '172.16.100.1', '10.0.0.1'),
 # L4 protocol and source/destination IP addresses
 em::ExactMatch(fields=[{'offset':23, 'size':1},
                        {'offset':26, 'size':4},
-                       {'offset':30, 'size':4}])
+                       {'offset':30, 'size':4}],
+                masks=[{'value_int':0xff},{'value_int':0xff},{'value_int':0xff}])
 
 Source() -> Rewrite(templates=packets) -> em
 
@@ -33,12 +34,12 @@ em:1 -> Sink()
 em:2 -> Sink()
 em:5 -> Sink()   # used as default gate
 
-em.add(fields=[chr(17), aton('172.16.100.1'), aton('10.0.0.1')], gate=0)
-em.add(fields=[chr(6), aton('172.12.55.99'), aton('12.34.56.78')], gate=1)
-em.add(fields=[chr(17), aton('172.12.55.99'), aton('12.34.56.78')], gate=2)
+em.add(fields=[{'value_int':17}, {'value_bin':aton('172.16.100.1')}, {'value_bin':aton('10.0.0.1')}], gate=0)
+em.add(fields=[{'value_int':6}, {'value_bin':aton('172.12.55.99')}, {'value_bin':aton('12.34.56.78')}], gate=1)
+em.add(fields=[{'value_int':17}, {'value_bin':aton('172.12.55.99')}, {'value_bin':aton('12.34.56.78')}], gate=2)
 
 # delete test
-em.add(fields=[chr(17), aton('192.168.1.123'), aton('12.34.56.78')], gate=3)
-em.delete(fields=[chr(17), aton('192.168.1.123'), aton('12.34.56.78')])
+em.add(fields=[{'value_int':17}, {'value_bin':aton('192.168.1.123')}, {'value_bin':aton('12.34.56.78')}], gate=3)
+em.delete(fields=[{'value_int':17}, {'value_bin':aton('192.168.1.123')}, {'value_bin':aton('12.34.56.78')}])
 
 em.set_default_gate(gate=5)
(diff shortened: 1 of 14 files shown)


justinemarie:
I believe that is the case so far, although we have talked in the past about separate libbess-python and bessd releases -- part of the reason we haven't made such a split is that we want the freedom to change the API at will.

bessctl/conf/samples/exactmatch.bess
```diff
@@ -33,12 +33,12 @@ em:1 -> Sink()
 em:2 -> Sink()
 em:5 -> Sink()   # used as default gate
 
-em.add(fields=[chr(17), aton('172.16.100.1'), aton('10.0.0.1')], gate=0)
-em.add(fields=[chr(6), aton('172.12.55.99'), aton('12.34.56.78')], gate=1)
-em.add(fields=[chr(17), aton('172.12.55.99'), aton('12.34.56.78')], gate=2)
+em.add(fields=[{'value_int':17}, {'value_bin':aton('172.16.100.1')}, {'value_bin':aton('10.0.0.1')}], gate=0)
```
sangjinhan: It would be nice if we can directly specify a scalar value for `oneof` variables, as long as there is no type ambiguity. For example, this line can be rewritten as `em.add(fields=[17, aton('172.16.100.1'), aton('10.0.0.1')], gate=0)`, which is shorter and sweeter.

It would require some non trivial changes in `protobuf_to_dict.py`...

protobuf/util_msg.proto
```diff
@@ -0,0 +1,22 @@
+syntax="proto3";
```
sangjinhan: We may drop `_msg` in the filename. As the file extensions are always either `.proto`, `.pb.h`, etc., it is clear that the file is protobuf stuff. Same goes for `bess_msg.proto`, `module_msg.proto`, and `port_msg.proto`.

protobuf/util_msg.proto
```diff
@@ -0,0 +1,22 @@
...
+
+package bess.pb;
+
+/// The Field message represents one field in a packet -- either stored in metadata or in the packet body.
+message Field {
+  oneof position {
+    string attr_name = 1; /// The metadata attribute assigned to store the data
+    uint32 offset = 2; /// The offset in bytes to store the data into
+  }
+  uint32 size = 3; /// The size of the data in bytes
```
sangjinhan: How about renaming to `bytes`?

core/modules/exact_match.cc
```diff
@@ -21,23 +21,24 @@ const Commands ExactMatch::cmds = {
...
+                                        struct EmField *f, int idx) {
   f->size = field.size();
   if (f->size < 1 || f->size > MAX_FIELD_SIZE) {
     return CommandFailure(EINVAL, "idx %d: 'size' must be 1-%d", idx,
                           MAX_FIELD_SIZE);
   }
 
-  if (field.position_case() == bess::pb::ExactMatchArg_Field::kAttribute) {
-    const char *attr = field.attribute().c_str();
+  if (field.position_case() == bess::pb::Field::kAttrName) {
```
justinemarie: That would be a much cleaner syntax but unfortunately protobuf doesn't expose those functions (for whatever reason it actually has them implemented but they are private?!) 
https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#oneof

bessctl/conf/samples/exactmatch.bess
justinemarie: Is this a "this would be nice for the future" or "let's hold on to this PR until Justine implements this"?

protobuf/util_msg.proto
justinemarie: Right now some of our scripts require _msg though...

core/modules/exact_match.cc
```diff
@@ -49,19 +50,20 @@ CommandResponse ExactMatch::AddFieldOne(
   }
```
justinemarie: Fixed.

protobuf/util_msg.proto
justinemarie: Fixed.

core/modules/exact_match.cc
sangjinhan: Oh you are right. This seems to be a proto2 vs proto3 difference... sigh.

bessctl/conf/samples/exactmatch.bess
chris3torek: I was curious and was looking at how to implement such a thing (it's still not quite clear, I think you can inspect for the `.oneofs` or `.oneofs_by_name` field to figure them out) but it wonder if it might also be a bit error prone...

sangjinhan: Either is fine :)

justinemarie:
I think this is ready pending 32-bit bug, and assuming we'll dig into the type-ambiguity in a later PR.

justinemarie:
*bump*

sangjinhan:
Thanks!
bgq582e6fc5888a4

PR #552, opened 2017-07-03
Keep running python unittests after failure

sangjinhan:
At the moment, one failure causes all subsequent test cases to fail. It should not. Example:
https://travis-ci.org/NetSys/bess/jobs/248679890#L1425


changlan:
image

PR #553, opened 2017-07-04
Add prefix_length member variable to Ipv4Prefix

gampel:


changlan:
Thanks for the PR. I am not clear about the purpose of this patch. Isn't `prefix_length` implied by the value of `mask` already?

sangjinhan:
Yeah, perhaps you can add a method using __builtin_ffs() to retrieve the prefix length.

gampel:
Will do, I will add a method using the __builtin

core/utils/ip.cc
```diff
@@ -52,6 +52,7 @@ Ipv4Prefix::Ipv4Prefix(const std::string &prefix) {
   } else {
     mask = be32_t(~((1 << (32 - len)) - 1));
   }
+
```
changlan: redundant linebreak?

core/utils/ip.h
```diff
@@ -67,6 +67,8 @@ struct Ipv4Prefix {
 
   // Returns true if ip is within the range of Ipv4Prefix
   bool Match(const be32_t &ip) const { return (addr & mask) == (ip & mask); }
+  // Returns the prefix length
+  uint32_t prefixLength() const {return ((32 - __builtin_clz(mask.raw_value())));}
```
changlan: Perhaps rename the function to `prefix_length` (accessors may be named like variables as per Google C++ Style Guide)

gampel: Done

core/utils/ip.cc
gampel: will be removed

changlan:
Merged, thanks!
image

PR #554, opened 2017-07-05
Add MAINTAINERS.rst

GalSagie:
As any open source project it is important to understand the people that has access
to push changes to the git repository and the diversity of the community as a whole.
The list was created based on previous commits and might be missing, i also didn't add emails
but think it would be great if everyone add their own emails.


sangjinhan:
Resolved by #578

PR #555, opened 2017-07-05
vagrant: provision VirtualBox VM with 5GB RAM

chris3torek:
The Wiki instructions say "5GB" but the Vagrantfile says "2GB".

When I built way back when, 2GB failed, 5GB worked.

(I've had this on my laptop for quite a while, it was when I rebuilt my vagrant-based VM yesterday I noticed it was time to send a pull request)


changlan:
Nice catch!
image

PR #556, opened 2017-07-06
ip_lookup: work around DPDK out-of-memory bug

chris3torek:
This is just a small hack to try to help diagnose whatever is going wrong
with Travis CI builds (seems to disproportionally affect the _32 builds,
though this does seem to happen randomly with other builds as well).

---

When we call rte_lpm_create() it returns NULL if the
parameters are invalid or if it runs out of memory.
For all but the "out of memory" case it sets rte_errno,
but when it runs out of memory, it fails to set rte_errno
at all.  This leaves whatever per-thread value it had
in it, which can often be zero.  This causes the subsequent
"return CommandFailure(rte_errno, ...)" to return a
failure with a zero error code, which pybess interprets
as success, so that it goes on to use the invalid data
and crashes bess.

While we probably should not crash, and perhaps should
have CommandFailure guarantee a nonzero return code or
otherwise arrange for failures to show up as failures,
we can work around this easily enough for now by pre-setting
ENOMEM as the error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,7 @@ after_success:
 after_failure:
   - more /tmp/bessd.*.INFO.* | cat      # dump all bessd log files
   - more /tmp/sanitizer* | cat          # dump all sanitizer results
+  - fgrep Huge /proc/meminfo /sys/devices/system/node/node*/meminfo
 
 before_deploy:
   - for arch in core2 sandybridge haswell; do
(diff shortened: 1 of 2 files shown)


chris3torek:
The travis-ci 32 bit build failed, with the new check that doesn't crash bess. So that part is good. Still not sure why the builds fail but suspect it has to do with only 512 hugepages reserved in the travis setup. The bess ip_lookup.cc module asks for memory specifically on socket 0; maybe Travis often has too few socket-zero pages available?

chris3torek:
retracting this for now, the travis build failed weirdly and I want to put it on a branch anyway

PR #557, opened 2017-07-06
pybess: Fix protobuf dict parsing to map field

sirspinach:
Currently, map fields in protobuf messages cannot be parsed from
dictionaries because the parsing function expects dictionary
values to also be dictionaries. (Rather than plain old data,
as is found in maps).

Before recursing on dictionary contents, first check that every
value in the dictionary is also a dictionary. Also add unit tests
to confirm proper parsing for both nested and unnested dictionaries.


sangjinhan:
I am a bit hesitant to add a test file in the `protobuf/`. Perhaps you can have the `.proto` file in `pybess/` and dynamically generate `pb.py` file in `test_protobuf_to_dict.py` if import fails?

sirspinach:
@sangjinhan It looks like dynamic protobuf compilation won't work outside of the container b/c protoc is not available. What do you think about adding modifying "coverage2 python -m unittest discover" to run within the container?

sangjinhan:
Oh I see. The unittest must run outside the container though... One alternative would be creating `protobuf/tests` and moving the file (renamed.... `nested_dict.proto`?) into the directory. What we want is not mixing real protobuf files and test files, nothing more :)

sangjinhan:
hu_1446272741_3153844886

PR #559, opened 2017-07-07
Fix removing unrelated packages in ansible script

gampel:
If the expreation contains '*' then it is assumed to be
a POSIX-style regular expression, and it is applied to
all package names in the database.

It is recommanded to anchor the regular expression with
a ^ to avoid removing of unrelated packges.

The the reg expression  "libprotobuf*" include
"libprotobuf-lite9v5" which is needed in many pacakges that
will be removed by that in ubuntu 16.04 like  ubuntu-desktop
ubuntu-docs ubuntu-mono and other.

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -69,7 +69,7 @@
       become: true
 
     - name: Wipe prior protobuf installs to avoid conflicts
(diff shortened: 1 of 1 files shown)


sangjinhan:
Merged, thanks!
ecaeaa3b5f6055b581a66a07cb6c6fdd

PR #562, opened 2017-07-07
Set default GCC version to 5 in the image

changlan:
`./container_build.py` was still using GCC 4 by default, but BESS rejects lower versions than 5. As a result, users had to manually export `CXX=g++-5` even with `./container_build.py`. This PR sets `g++-5` to be the default version, and the image on the Docker Hub is updated as well.

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -14,7 +14,6 @@
         - g++-5
         - g++-6
         - g++-7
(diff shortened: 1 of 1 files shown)


sangjinhan:
bgq582e6fc54e6e9

PR #564, opened 2017-07-10
Do not generate per-packet log in non-verbose mode

sangjinhan:
For TCP stream reconstruction, packet loss or out-of-order arrival is rather common. URLFilter module generates per-packet log messages in such cases, which is too verbose. DLOG(WARNING) messages were replaced with VLOG(1), so as to be printed only with --v=1 or higher.


changlan:
image

PR #565, opened 2017-07-10
Reduce message logging

apanda:
This change removes logging when constraints are not violated, thus
reducing log verbosity.


sangjinhan:
default-3

PR #566, opened 2017-07-10
Allocate hugepages before installing dependencies

sangjinhan:
This may suppress the IPLookup module failure on Travis with 32bit arch....

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,8 @@ env:
     - VER_CXX=g++-7
 
 before_install:
+  # Reserve hugepages as early as possible, to avoid fragmentation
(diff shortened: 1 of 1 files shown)


sangjinhan:
Hi Sami!
(Is she/he Finn?)

chris3torek:
No, she's just "Siam spelled sideways". Sami was an orphan kitten added to the two I was fostering and then ended up adopting ("failed foster", as they call it). To avoid getting too attached, I gave them temporary names: Thing 1 (Alice or Trouble), Thing 2 (whichever one wasn't Thing 1), and Sam-I-Am, all from The Cat in the Hat.  Then Sam-I-Am shortened to Sam-I => Sami....

PR #570, opened 2017-07-13
Adds a queue command to get the queue's occupancy.

barath:


barath:
Done.

vivi:
`Queue::GetDesc` has the same logic, maybe make `QueueCommandGetStatusResponse` contain a string instead?

barath:
Yeah -- unfortunately I want the numbers rather than a string representation.  I thought about changing `GetDesc` but it's so simple that it seemed the extra layer of indirection wasn't worth it.

melvinw:
Thanks! cat

PR #572, opened 2017-07-14
Remove Traverse() feature from TC trees

sangjinhan:
It has been observed a couple of times on Travis that TrafficClass::Traverse() gets stuck in an infinite loop. While its cause is unknown yet, the tree traversal mechanism of Traverse() doesn't seem to be strictly necessary; all it use cases can be replaced with simple iteration over all TCs. I hope this simpler implementation makes the bug less likely to occur...

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -566,13 +566,12 @@ class BESSControlImpl final : public BESSControl::Service {
 
     bess::TrafficClass* root = workers[wid]->scheduler()->root();
     if (root) {
-      bool has_tasks = false;
-      root->Traverse([&has_tasks](bess::TCChildArgs* c) {
-        has_tasks |= c->child()->policy() == bess::POLICY_LEAF;
-      });
-      if (has_tasks) {
-        return return_with_error(response, EBUSY, "Worker %d has active tasks ",
-                                 wid);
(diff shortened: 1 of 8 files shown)


core/bessctl.cc
```diff
@@ -595,57 +594,50 @@ class BESSControlImpl final : public BESSControl::Service {
...
-        collect_tc(c, i, status);
-
-        if (args->parent_type() == bess::POLICY_WEIGHTED_FAIR) {
-          auto ca = static_cast<bess::WeightedFairChildArgs*>(args);
-          status->mutable_class_()->set_share(ca->share());
-        } else if (args->parent_type() == bess::POLICY_PRIORITY) {
-          auto ca = static_cast<bess::PriorityChildArgs*>(args);
-          status->mutable_class_()->set_priority(ca->priority());
+    } else if (wid < 0) {
+      wid = -1;
```
barath: Minor: might it be worth changing -1 to something more descriptive (e.g. a constant / enum)?

barath:
LGTM otherwise.

core/bessctl.cc
sangjinhan: Yup, done.

PR #573, opened 2017-07-14
Add L4Checksum module and add support for IP options to IPChecksum

shinae-woo:
Adding functionality to calculate checksums for IPv4 header with options, UDP.
Adding a module to calculate L4 (UDP/TCP) over IPv4.


apanda:
image

PR #575, opened 2017-07-17
modules/port_out: pick queue using input gate

melvinw:
This should resolve the old TODO.

Credit to @ddiproietto for the idea.

### DIFF
--- a/core/modules/port_out.cc
+++ b/core/modules/port_out.cc
@@ -48,19 +48,21 @@ std::string PortOut::GetDesc() const {
 void PortOut::ProcessBatch(bess::PacketBatch *batch) {
   Port *p = port_;
 
-  /* TODO: choose appropriate out queue */
-  const queue_t qid = 0;
+  const queue_t qid = get_igate();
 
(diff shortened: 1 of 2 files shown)


melvinw:
Not sure whether it makes more sense to pick a default, valid value for `qid` when the input gate is out of bounds or to just drop the batch. I chose to drop, but I'm happy to change if anyone feels strongly enough in the other direction.

sangjinhan:
The original idea (which has never been implemented though) was automatic mapping between cores (workers) <-> TX queues, plus locking if necessary. But I guess we can later add an option to control its default behavior.

Dropping packets for invalid qid sounds good to me

sangjinhan:
Note for future reference: PortOut module is not thread-safe at the moment, as multiple workers may access the same TX queue.

sangjinhan:
defbfb8b739a101db1d3f84e3d04c1a7

llvilanova:
Sorry to resurrect an old issue, but I just wanted to point out that the proto file was not updated to document this (which would eventually reach the wiki).

PR #577, opened 2017-07-17
CI: Run unittests in an arbitrary order

sangjinhan:
... to add more randomness. Done with `--gtest_shuffle`

Some tests in `post_test.cc` were order-dependent and fixed.


barath:
Thanks for doing this.

PR #579, opened 2017-07-17
Fix a bug that ixgbevf does not show any stats for show/monitor port

shinae-woo:
ixgbevf does not support per queue stats.
  It is hard coded since I couldn't find a way to identify whether the driver supports queue-level or port-level stat functionality yet.
  (no flags, no driver-level supported functions, etc).

  BTW, in EC2 instances using ixgbevf, the stats are incorrect.
  EC2 limits pps per EC2 instance, (e.g., 75Kpps for c2.xlarge),
  the stats show more pps than rate limited performance and no drop count.

  It looks like rate limiting happens in host side and hidden from the EC2 instances.

### DIFF
--- a/core/drivers/pmd.cc
+++ b/core/drivers/pmd.cc
@@ -335,8 +335,9 @@ void PMDPort::CollectStats(bool reset) {
 
   port_stats_.inc.dropped = stats.imissed;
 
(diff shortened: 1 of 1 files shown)


sangjinhan:
Merged, thanks!
default-6

PR #582, opened 2017-07-19
Buildtweaks

chris3torek:
This is a series of tweaks to build.py and container_build.py aimed at some usability improvement for development with DPDK updates if and when this is required, and for building with automated build systems (e.g., travis, where we've seen things like travis timeouts—we could work around these by tweaking the travis settings but I like this idea better).


chris3torek:
(moving this to other fork)

PR #583, opened 2017-07-19
Build tweaks

chris3torek:
This is a series of tweaks to build.py and container_build.py aimed at some usability improvement for development with DPDK updates if and when this is required, and for building with automated build systems (e.g., travis, where we've seen things like travis timeouts—we could work around these by tweaking the travis settings but I like this idea better).

### DIFF
--- a/build.py
+++ b/build.py
@@ -188,7 +188,11 @@ def generate_extra_mk():
         fp.write('LDFLAGS += %s\n' % ' '.join(ld_flags))
 
 
-def download_dpdk():
+def download_dpdk(quiet=False):
+    if os.path.exists(DPDK_DIR):
+        if not quiet:
+            print('already downloaded to %s' % DPDK_DIR)
+        return
     try:
         print('Downloading %s ...  ' % DPDK_REPO)
         # Using --depth 1 speeds download, but leaves some traps.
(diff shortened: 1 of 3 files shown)


build.py
```diff
@@ -317,41 +349,44 @@ def update_benchmark_path(path):
...
-    elif args.action == 'clean':
-        do_clean()
-    elif args.action == 'dist_clean':
-        do_dist_clean()
-    elif args.action == 'help':
-        print_usage(parser)
-    else:
-        print('Error - unknown command "%s".' % sys.argv[1], file=sys.stderr)
-        print_usage(parser)
+    cmds[args.action]()
```
changlan: 👍

sirspinach:
Could you change .travis.yml to build in verbose by default?

chris3torek:
I thought I'd leave that to a separate thing (mostly because I'm not sure which builds are timing out in the first place). Which one(s) are?

chris3torek:
(This was discussed elsewhere but I'll add it here since I guess these discussions get archived on GitHub.) It looks like the place to tweak in travis is in another repo that just uses this one.

changlan:
Merged, thanks!
image

PR #584, opened 2017-07-19
Allow module inheritance across shared objects

sirspinach:
Previously, modules could not access (and therefore inherit from)
other modules because the RTLD_GLOBAL flag was not set. This patch
allows inheritance by dlopening module shared objects with the
RTLD_GLOBAL flag and by loading modules in multiple passes.

### DIFF
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -427,12 +427,15 @@ bool LoadPlugins(const std::string &directory) {
       const std::string full_path = *it;
       LOG(INFO) << "Loading module (pass " << pass << "): " << full_path;
       if (!LoadPlugin(full_path)) {
-        LOG(WARNING) << "Error loading module " << full_path << ": "
-                     << dlerror();
+        LOG(INFO) << "Error loading module " << full_path << ": " << dlerror();
       } else {
         remaining.erase(it++);
       }
     }
+  }
+
+  for (auto it = remaining.begin(); it != remaining.end(); ++it) {
+    LOG(WARNING) << "Gave up on loading module " << *it;
   }
 
   return (remaining.size() == 0);
--- a/core/bessd.cc
+++ b/core/bessd.cc
@@ -427,7 +427,7 @@ bool LoadPlugins(const std::string &directory) {
       const std::string full_path = *it;
       LOG(INFO) << "Loading module (pass " << pass << "): " << full_path;
       if (!LoadPlugin(full_path)) {
-        LOG(INFO) << "Error loading module " << full_path << ": " << dlerror();
+        VLOG(1) << "Error loading module " << full_path << ": " << dlerror();
       } else {
         remaining.erase(it++);
       }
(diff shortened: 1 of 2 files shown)


core/bessd.h
```diff
@@ -10,6 +10,7 @@
 namespace bess {
 namespace bessd {
 
+const int kInheritanceLimit = 10;
```
barath: Could you add a comment for this constant?

vivi:
(I would test your module with inheritance extensively before merging this...still not sure if inheritance works 100%) :P

sirspinach:
Is there any concern here with making modules less isolated from each other? (one malicious/buggy module might crash another?)

core/bessd.cc
```diff
@@ -405,6 +406,7 @@ bool UnloadPlugin(const std::string &path) {
 }
 
 bool LoadPlugins(const std::string &directory) {
+  std::list<std::string> failed = {};
```
barath: No need to brace initialize (it will default empty).

core/bessd.cc
```diff
@@ -413,17 +415,31 @@ bool LoadPlugins(const std::string &directory) {
...
                      << dlerror();
-        closedir(dir);
-        return false;
       }
     }
   }
   closedir(dir);
-  return true;
+
+  int iter_cnt = 2;
```
chris3torek: trivial: I'd just call this `pass`

changlan:
Would it be simpler to combine two `while` loops of loading modules by initializing `failed` list with all unloaded modules at the beginning?

sangjinhan:
ERROR would be a better fit than INFO...?

core/bessd.cc
```diff
@@ -409,21 +410,35 @@ bool LoadPlugins(const std::string &directory) {
...
+      const std::string full_path = *it;
+      LOG(INFO) << "Loading module (pass " << pass << "): " << full_path;
       if (!LoadPlugin(full_path)) {
-        LOG(WARNING) << "Error loading module " << full_path << ": "
-                     << dlerror();
-        closedir(dir);
-        return false;
+        LOG(INFO) << "Error loading module " << full_path << ": " << dlerror();
+      } else {
+        remaining.erase(it++);
```
changlan: Would the next item be skipped? (there is a `++it` in `for`)

sirspinach: ah, thank you for catching that!

core/bessd.cc
```diff
@@ -422,14 +422,14 @@ bool LoadPlugins(const std::string &directory) {
...
-       pass++) {
+       ++pass) {
     for (auto it = remaining.begin(); it != remaining.end(); ++it) {
       const std::string full_path = *it;
       LOG(INFO) << "Loading module (pass " << pass << "): " << full_path;
       if (!LoadPlugin(full_path)) {
         VLOG(1) << "Error loading module " << full_path << ": " << dlerror();
       } else {
-        remaining.erase(it++);
+        remaining.erase(it);
```
vivi: I don't think this will work either; the `it` is invalid after erasure. Maybe `for (auto it = remaining.begin(); it != remaining.end();) ... if (!LoadPlugin) .. increment .. else erase(it++)`

sirspinach: Thank you, that's right

chris3torek: I was wondering about this myself. It looks like the standard C++ idiom (as of C++11) is:

```
for (auto i = foo.begin(); i != foo.end();) {
    if (...) ++i; else i = foo.erase(i);
```

sangjinhan:
Sorry, could you please rebase the branch?

sirspinach:
sure

melvinw:
Rebased this. Will open new pr

PR #589, opened 2017-07-20
bessctl: add DumpMempool command

melvinw:
This was useful during some debugging I did today. Others might find it helpful in the future.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -1785,8 +1785,9 @@ def interactive(cli):
     cli.fout = old_fout
     cli.interactive = False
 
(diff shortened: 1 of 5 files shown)


sangjinhan:
Looks great. Thanks!

One minor request: could you change the command to `show system packets`? I expect there would be many more commands in a similar nature to come in the near future; a single `show system ...` command would be easier than remembering individual commands.

melvinw:
You got it. Done

sangjinhan:
05f30c7f3d8cc900e5e94d107edec0ea

PR #590, opened 2017-07-20
Add MPLS push module

GalSagie:


nrdmao:
So with this design you will have an instance of the MPLS Push module for every unique label to be pushed. Which, in some instances, might be 1000's of labels. Does this really scale well in BESS to have 1000's modules and therefore 1000's of connections between modules?

core/modules/mpls_push.cc
```diff
@@ -0,0 +1,52 @@
...
+    bess::Packet *pkt = batch->pkts()[i];
+    Ethernet *eth = pkt->head_data<Ethernet *>();
+
+    Ethernet::Address src_addr = eth->src_addr;
+    Ethernet::Address dst_addr = eth->dst_addr;
+
+    pkt->prepend(4);
+    eth = pkt->head_data<Ethernet *>();
+    eth->src_addr = src_addr;
+    eth->dst_addr = dst_addr;
```
melvinw: To the end of keeping the scope of this module narrow, might you just copy the original ethernet header here and change the ethertype? A sequence of `SetMetadata` and `EtherEncap` modules downstream of this module can take care of changing the addresses if needed.

core/modules/mpls_push.cc
```diff
@@ -0,0 +1,52 @@
...
+void MPLSPush::ProcessBatch(bess::PacketBatch *batch) {
+  int cnt = batch->cnt();
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    Ethernet *eth = pkt->head_data<Ethernet *>();
+
+    Ethernet::Address src_addr = eth->src_addr;
+    Ethernet::Address dst_addr = eth->dst_addr;
+
+    pkt->prepend(4);
```
melvinw: All code south of here should be guarded by a check that this call to `Packet::prepend()` didn't fail.

core/modules/mpls_push.cc
```diff
@@ -0,0 +1,52 @@
...
+}
+
+void MPLSPush::ProcessBatch(bess::PacketBatch *batch) {
+  int cnt = batch->cnt();
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    Ethernet *eth = pkt->head_data<Ethernet *>();
+
+    Ethernet::Address src_addr = eth->src_addr;
+    Ethernet::Address dst_addr = eth->dst_addr;
```
melvinw: Doh, sorry. Ignore my comment below. I didn't notice these lines.

sangjinhan:
@nrdmao That's a great point. At that scale probably we would want:
1. All packets to go thorough a single pipeline, unless their processing flows diverge.
2. Each packet carries its MPLS label as metadata, which is set by some upstream modules.
3. Then MPLSPush module tags packets accordingly on a per-packet basis

(3) is what some existing modules do, such as `IPEncap`, `EtherEncap`, etc. In contrast, some other modules work on a per-"flow" basis, such as `VLANPush` or this module. Ideally, it should be configurable to choose which scheme to use.

At the moment, unfortunately, the per-packet metadata approach is not quite applicable in most cases.  Existing classification modules (`BPF`, `IPLookup`, `ExactMatch`, etc.) support only flow split (packets are distributed across output gates, as a result of classification), but not metadata update.

sangjinhan:
Sorry, could you rebase the branch? Due to changes in #584 there are build errors in this PR...

PR #592, opened 2017-07-25
Trie: Add support for values and prefix nodes

sirspinach:
A prefix key "asdf" can be inserted into the Trie to match all keys that begin with "asdf". `Lookup()` returns the value associated with the most specific match, prioritizing an exact match over the longest prefix.

### DIFF
--- a/core/modules/url_filter.cc
+++ b/core/modules/url_filter.cc
@@ -142,8 +142,7 @@ CommandResponse UrlFilter::Init(const bess::pb::UrlFilterArg &arg) {
                          std::forward_as_tuple(url.host()),
                          std::forward_as_tuple());
     }
-    Trie<std::tuple<>> &trie = blacklist_.at(url.host());
-    trie.Insert(url.path(), {});
+    blacklist_[url.host()].Insert(url.path(), {});
   }
   return CommandSuccess();
 }
--- a/core/modules/url_filter.cc
+++ b/core/modules/url_filter.cc
@@ -1,6 +1,7 @@
 #include "url_filter.h"
 
 #include <algorithm>
+#include <tuple>
 
 #include "../utils/checksum.h"
 #include "../utils/ether.h"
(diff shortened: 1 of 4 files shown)


core/modules/url_filter.cc
```diff
@@ -141,8 +142,8 @@ CommandResponse UrlFilter::Init(const bess::pb::UrlFilterArg &arg) {
                          std::forward_as_tuple(url.host()),
                          std::forward_as_tuple());
     }
-    Trie &trie = blacklist_.at(url.host());
-    trie.Insert(url.path());
+    Trie<std::tuple<>> &trie = blacklist_.at(url.host());
+    trie.Insert(url.path(), {});
```
barath: Perhaps rewrite these two lines as:
`blacklist_[url.host()].Insert(url.path(), {});`

core/utils/trie.h
```diff
@@ -34,26 +40,47 @@ class Trie {
...
 
   Trie() : root_() {}
-  Trie(const Trie& t) : root_(t.root_) { }
+  Trie(const Trie& t) : root_(t.root_) {}
 
-  // Inserts a string into the trie
-  void Insert(const std::string& key);
+  // Inserts a string into the trie, associating the key
+  // with the value.
+  void Insert(const std::string& key, const T val);
```
barath: You should probably have `val` be passed as a const reference since T is likely not a reference type.  (Here and for other methods.)

core/utils/trie_test.cc
```diff
@@ -1,91 +1,190 @@
...
+  EXPECT_TRUE(trie2.MatchPrefix(""));
+  EXPECT_TRUE(trie2.MatchPrefix("1"));
+
+  EXPECT_TRUE(trie2.Lookup("Hello").first);
+  EXPECT_TRUE(trie2.Lookup("Hello y'all").first);
+  EXPECT_TRUE(trie2.Lookup("Hello World").first);
+
+  EXPECT_EQ(trie2.Lookup("Hello").second, 2);
+  EXPECT_EQ(trie2.Lookup("Hello y'all").second, 2);
+  EXPECT_EQ(trie2.Lookup("Hello World").second, 2);
```
nrdmao: The interesting corner cases that arise with respect to the management of the Trie are inserts and deletes which force the creation and destruction of Nodes and the associated processing of the children array. A test that randomly creates thousands of keys of say between 1 and 8 characters and then randomly removes and inserts keys while ensuring that lookups work as expected might more thoroughly prove out these algorithms.

sirspinach: I haven't added deletion to the Trie yet, but I'll remember to write a test like that when I do.

sangjinhan:
62269c9bb6312f35247579865d895a33

PR #593, opened 2017-07-25
Pytweaks

chris3torek:
This fixes a few minor bugs and changes the way the standard test modules are run so that their namespaces behave the way one would expect in a Python module.

In particular, if you write custom test functions, you had to avoid using global (module level) functions and/or classes -- everything had to be within the local scope of each function (i.e., closures).  Now test module code works the same way as other `run <path>` BESS scripts: it gets `exec`-ed in a module with a private namespace, rather than borrowing the local variables of the loader function.

I included an example of the other bug I fixed.  I can't explain why the assertion failure in `pybess/module.py` is missing the _name_ though:

```
    assert False, '%s is not a module' % next_mod
AssertionError:  is not a module
```

(why isn't `next_mod` the string, here?)

### DIFF
--- a/bessctl/conf/testing/run_module_tests.bess
+++ b/bessctl/conf/testing/run_module_tests.bess
@@ -106,20 +106,46 @@ def crash_test(module, num_input_ports, num_output_ports):
     time.sleep(5)
     bess.pause_all()
 
-# Can't just call exec() in the middle of main loop because
-# variables will leak into main script scope
 def load_test(filename):
-    CRASH_TEST_INPUTS = []
-    OUTPUT_TEST_INPUTS = []
-    CUSTOM_TEST_FUNCTIONS = []
+    """
+    Load a test file.  It should modify the (global to it) variables
+    CRASH_TEST_INPUTS, OUTPUT_TEST_INPUTS, and CUSTOM_TEST_FUNCTIONS
+    using .append() to add its tests.  The format of each of these is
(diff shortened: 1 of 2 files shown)


bessctl/conf/testing/run_module_tests.bess
```diff
@@ -106,20 +106,46 @@ def crash_test(module, num_input_ports, num_output_ports):
...
 def load_test(filename):
-    CRASH_TEST_INPUTS = []
-    OUTPUT_TEST_INPUTS = []
-    CUSTOM_TEST_FUNCTIONS = []
+    """
+    Load a test file.  It should modify the (global to it) variables
+    CRASH_TEST_INPUTS, OUTPUT_TEST_INPUTS, and CUSTOM_TEST_FUNCTIONS
+    using .append() to add its tests.  The format of each of these is
+    a bit different; see the wiki at
+    https://github.com/NetSys/bess/wiki/Python-Script-Namespaces
```
sangjinhan: Should be https://github.com/NetSys/bess/wiki/Python-Scripts ?

chris3torek: Oops, yes, I moved the page (before actually creating it) when I made it more general. Should I re-push the pull request with corrected web address, or just fix it later?

sangjinhan:
Merged, thanks!
b4c754c9b93669f25e922a06644bd88e

PR #595, opened 2017-07-27
Fix tcpdump for scattered packet buffers

sangjinhan:
Scattered packet buffers have `total_len` larger than `data_len`, which accounts only the first segment. Since we consider only the first segment has been captured, not the entire packet, only `data_len` bytes must be written to the fifo.

Also, internally `fifo_fd_` stored 0 to represent "no file descriptor is open", but fd 0 might be actually being used. Because of this the class destructor may accidentally close `stdin` or any other file descriptor allocated to 0.


changlan:
image

PR #596, opened 2017-07-27
Fix minor CLI issues

sangjinhan:
* `import plugin PLUGIN_FILE` doesn't correctly support auto completion
* No space between a long command and its description
* Command abbreviation does not work for gate `DIRECTION` type
  * e.g., `tr ena * o[ut]`
* `tcpdump MODULE [DIRECTION] [OGATE]...` should be `[GATE]`
* If one command is a strict prefix of another command, only the latter is shown as the only candidate for auto completion. For example `show worker<tab>` only shows the second command out of the below two commands.
```
show worker               Show the status of all worker threads
show worker WORKER_ID...  Show the status of specified worker threads
```

### DIFF
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -49,8 +49,7 @@ class ColorizedOutput(object):  # for pretty printing
 
 class CLI(object):
 
-    # general command errors
-    class CommandError(Exception):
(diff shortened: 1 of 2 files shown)


changlan:
image

PR #597, opened 2017-07-28
Prevent `bessctl` crash from  bess config file loading errors

shinae-woo:


bessctl/commands.py
```diff
@@ -784,11 +784,15 @@ class Foo(Module):
...
 def _do_run_file(cli, conf_file):
-    if not os.path.exists(conf_file):
+    if not os.path.isfile(conf_file):
         cli.err('Cannot open file "%s"' % conf_file)
         return
 
-    xformed = sugar.xform_file(conf_file)
+    try:
+        xformed = sugar.xform_file(conf_file)
+    except:
```
chris3torek: Should probably list the specific exceptions here, although it's annoying since Python3 uses `FileNotFoundError` if it doesn't find the file, while py2k uses `IOError` (and doesn't have `FileNotFoundError`!).  We probably shouldn't use `os.path.exists` or `os.path.isfile` at all—just try the `io.open`.  Maybe we should put the exception handling in `sugar.py` itself since the testing code uses it too?

shinae-woo: I will :)

shinae-woo:
I catched both `IOError` and `OSError` (parent of `FileNotFoundError`, as there may be other cases, e.g., permission denied).
I kept the try catch statements in `commands.py`, but not in `sugar.py`. 
`sugar.py` doesn't have contexts of CLI environments, but a utility library, so the caller should know what to do with exceptions. In `test_sugar.py`, the tests should check whether the utility raises appropriate exceptions.

chris3torek:
Sounds good, should be fine this way.

sangjinhan:
Instead of `return`, could you do `raise cli.HandledError()` for both error cases? When bessctl is running in non-interactive mode (e.g., `cat commands.txt | bessctl`), bessctl should stop there without running subsequent commands. `HandledError` will cause bessctl to stop processing the command loop in non-interactive mode, but `return` won't.

sangjinhan:
Merged, thanks!
default-9

PR #599, opened 2017-08-01
module tests: Set socket timeout by parameter

sirspinach:
This is useful for reducing the run time of custom tests where lots of drops are expected.


sangjinhan:
zabott2mpjst659mrqjxya7vo1_500

PR #600, opened 2017-08-02
modules/set_metadata: add shift and mask support

melvinw:
This patch changes SetMetadata to optionally shift and mask (in that
order) values when copying metadata values from the contents of a
packet. It also fixes one of the metadata test scripts.

A quick micro benchmark on a 2.1GHz Xeon 2620v4 with the script below shows relatively minor overhead is imposed when `SetMetadata` is not configured to use the new features.

```
ATTR_SIZE = int($ATTR_SIZE!'1')
Source() \
    -> SetMetadata(attrs=[{'name': 'foo', 'size': ATTR_SIZE, 'offset': 0}]) \
    -> MetadataTest(read={'foo': ATTR_SIZE}) \
    -> Sink()
```

### Before

| attr. size | Mpps | cycles/pkt |
| - | - | - |
| 1 | 108.108 | 19.422 |
| 2 | 108.407 | 19.372 |
| 4 | 111.280 | 18.872 |
| 8 | 113.383 | 18.522 |
| 16 | 106.242 | 19.767 |
| 24 | 115.461 | 18.188 |
| 32 | 116.844 | 17.973 |

### After

|attr. size | Mpps | cycles/pkt |
| - | - | - |
| 1 | 103.293 | 20.331 |
| 2 | 103.518 | 20.287 |
| 4 | 109.368 | 19.202 |
| 8 | 109.321 | 19.210 |
| 16 | 109.904 | 19.108 |
| 24 | 111.408 | 18.850 |
| 32 | 114.985 | 18.264 |

### DIFF
--- a/core/modules/set_metadata.cc
+++ b/core/modules/set_metadata.cc
@@ -28,24 +28,87 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <cmath>
+#include <x86intrin.h>
+
 #include "set_metadata.h"
 
 #include "../utils/endian.h"
+#include "../utils/simd.h"
 
 static void CopyFromPacket(bess::PacketBatch *batch, const struct Attr *attr,
                            bess::metadata::mt_offset_t mt_off) {
   int cnt = batch->cnt();
   int size = attr->size;
-
+  int shift = attr->shift;
   int pkt_off = attr->offset;
+  int mt_adj = 0;
+  if (shift < 0) {
+    pkt_off += std::abs(shift);
+  } else if (shift > 0) {
+    mt_adj += shift;
+  }
+  size -= std::abs(shift);
 
   for (int i = 0; i < cnt; i++) {
     bess::Packet *pkt = batch->pkts()[i];
-    char *head = pkt->head_data<char *>();
-    void *mt_ptr;
+    uint8_t *head = pkt->head_data<uint8_t *>(pkt_off);
+    uint8_t *mt_ptr = _ptr_attr_with_offset<uint8_t>(mt_off + mt_adj, pkt);
+    bess::utils::CopySmall(mt_ptr, head, size);
+  }
+}
 
-    mt_ptr = _ptr_attr_with_offset<value_t>(mt_off, pkt);
-    bess::utils::CopySmall(mt_ptr, head + pkt_off, size);
+static void CopyFromPacketWithMask(bess::PacketBatch *batch,
+                                   const struct Attr *attr,
+                                   bess::metadata::mt_offset_t mt_off) {
+  int cnt = batch->cnt();
+  int size = attr->size;
+  mask_t mask = attr->mask;
+  int shift = attr->shift;
+  int pkt_off = attr->offset;
+  int mt_adj = 0;
+  if (shift < 0) {
+    pkt_off += std::abs(shift);
+  } else if (shift > 0) {
+    mt_adj += shift;
+  }
+
+  // Copy data
+  size -= std::abs(shift);
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    uint8_t *head = pkt->head_data<uint8_t *>(pkt_off);
+    uint8_t *mt_ptr = _ptr_attr_with_offset<uint8_t>(mt_off + mt_adj, pkt);
+    bess::utils::CopySmall(mt_ptr, head, size);
+  }
+  size += std::abs(shift);
+
+  // Apply mask
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    int len = size;
+
+    uint64_t *mt64 = _ptr_attr_with_offset<uint64_t>(mt_off + mt_adj, pkt);
+    uint64_t *mask64 = reinterpret_cast<uint64_t *>(mask.bytes);
+    __m128i *mt128 = reinterpret_cast<__m128i *>(mt64);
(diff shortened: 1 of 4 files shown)


melvinw:
Debugging a segfault. Marking not-ready for a bit

core/modules/set_metadata.cc
```diff
@@ -71,6 +134,8 @@ CommandResponse SetMetadata::AddAttrOne(
   size_t size = 0;
   int offset = -1;
   value_t value = value_t();
+  mask_t mask = mask_t();
```
sangjinhan: I am not entirely sure, but the default constructor of `mask_t` is no-op, thus this definition is not different from just `mask_t mask`?

protobuf/module_msg.proto
```diff
@@ -837,6 +837,8 @@ message SetMetadataArg {
       bytes value_bin = 4; /// A binary value to store in the packet (host-order).
     }
     int64 offset = 5; /// An index in the packet data to store copy into the metadata attribute.
+    bytes mask = 6; /// An array of bit masks to apply to each of the bytes copied starting from `offset`. If empty, the mask [0xFF,....,0xFF] will be used.
+    int32 shift = 7; /// The number of bytes to shift the masked value at offset by before storing it in metadata. Positive and negative values represent right and left shifts respectively.
```
sangjinhan: It would be nice if the comments clarify which operation is applied first. Is it `new = (old & mask) >> shift` or `new = (old >> shift) & mask`. The comment is read as the former ("...bytes to shift the masked value..."), but the actual code is more like the latter?

core/modules/set_metadata.cc
```diff
@@ -112,6 +178,16 @@ CommandResponse SetMetadata::AddAttrOne(
     if (offset < 0 || offset + size >= SNBUF_DATA) {
       return CommandFailure(EINVAL, "invalid packet offset");
     }
+    if (do_mask) {
+      memset(mask.bytes, 0xFF, size);
```
sangjinhan: This sentence has no effect at all because of `bess::utils::Copy` below.

core/modules/set_metadata.cc
```diff
@@ -28,24 +28,87 @@
...
+    __m128i a = gather_m128i(mt64, mt64 + 1);
+    __m128i b = gather_m128i(mask64, mask64 + 1);
+    *mt128 = _mm_and_si128(a, b);
+
+    len -= sizeof(__m128i);
+    mt128++;
+    mt64 += 2;
+    mask64 += 2;
+
+    while (len >= int{sizeof(__m128i)}) {
```
melvinw: Doh. Will shrink :)

protobuf/module_msg.proto
melvinw: The latter is correct. Will add a comment

core/modules/set_metadata.cc
```diff
@@ -28,24 +28,87 @@
...
-    bess::utils::CopySmall(mt_ptr, head + pkt_off, size);
+  // Apply mask
+  for (int i = 0; i < cnt; i++) {
+    bess::Packet *pkt = batch->pkts()[i];
+    int len = size;
+
+    uint64_t *mt64 = _ptr_attr_with_offset<uint64_t>(mt_off + mt_adj, pkt);
+    uint64_t *mask64 = reinterpret_cast<uint64_t *>(mask.bytes);
+    __m128i *mt128 = reinterpret_cast<__m128i *>(mt64);
+    __m128i a = gather_m128i(mt64, mt64 + 1);
```
melvinw: Ah, cool. Not sure why I couldn't find `_mm_loadu_si128()` yesterday.

core/modules/set_metadata.cc
```diff
@@ -28,24 +28,87 @@
...
+    bess::utils::CopySmall(mt_ptr, head, size);
+  }
+}
+
+static void CopyFromPacketWithMask(bess::PacketBatch *batch,
+                                   const struct Attr *attr,
+                                   bess::metadata::mt_offset_t mt_off) {
+  int cnt = batch->cnt();
+  int size = attr->size;
+  mask_t mask = attr->mask;
```
melvinw: Nope. Good catch

core/modules/set_metadata.cc
```diff
@@ -28,24 +28,87 @@
...
+    uint8_t *head = pkt->head_data<uint8_t *>(pkt_off);
+    uint8_t *mt_ptr = _ptr_attr_with_offset<uint8_t>(mt_off + mt_adj, pkt);
+    bess::utils::CopySmall(mt_ptr, head, size);
+  }
+  size += std::abs(shift);
 
-    mt_ptr = _ptr_attr_with_offset<value_t>(mt_off, pkt);
-    bess::utils::CopySmall(mt_ptr, head + pkt_off, size);
+  // Apply mask
+  for (int i = 0; i < cnt; i++) {
```
melvinw: Separating the loops had a pretty big (positive) impact on performance iirc (not sure how/why). Will double check and combine if I'm wrong

melvinw:
Thanks for the comments! I'll see if I can come up with a way to use `_mm_bslli_si128()` and `_mm_bsrli_si128()` to handle the shifting, though it's not immediately clear how I'd do that with a carry (maybe I don't need it).

core/modules/set_metadata.h
```diff
@@ -34,13 +34,17 @@
...
+typedef struct { uint8_t bytes[bess::metadata::kMetadataAttrMaxSize]; } value_t;
+typedef struct { uint8_t bytes[bess::metadata::kMetadataAttrMaxSize]; } mask_t;
 
 struct Attr {
   std::string name;
   value_t value;
   int offset;
   int size;
+  bool do_mask;
+  mask_t mask;
```
melvinw: Wow! This actually had a ~3-4 Mpps impact for some of my tests. Good catch

melvinw:
Going to reopen this as a slightly different PR

PR #602, opened 2017-08-04
modules/set_metadata: add shift/mask support

melvinw:
This commit changes SetMetadata to optionally shift and mask (in that
order) values when copying metadata values from the contents of a
packet. It also makes some small optimizations and cleans up parts of
the module.

A quick micro benchmark on a 2.1GHz xeon 2620v4 with the script below shows the new features impose no undue overhead when they aren't used.

```
ATTR_SIZE = int($ATTR_SIZE!'1')
Source() \
    -> SetMetadata(attrs=[{'name': 'foo', 'size': ATTR_SIZE, 'offset': 0}]) \
    -> MetadataTest(read={'foo': ATTR_SIZE}) \
    -> Sink()
```

| Attr. Size | Mpps Before | Mpps After |
| - | - | - |
| 1 | 103.293| 114.835 |
| 2 | 103.518| 114.918 |
| 4 | 109.368| 109.702 |
| 8 | 109.321| 114.845 |
| 16 | 109.904| 113.136 |
| 24 | 111.408| 119.894 |
| 32 | 114.985| 120.66 |

I'm not too sure how useful the shift functions other than `Shift*Small` are since the scope of my interest was limited to `SetMetadata` and it only operates on up to 32 bytes. I'll open another PR in the future (or append to this one if people prefer) with a proper benchmark and possible changes for larger buffers.

### DIFF
--- a/core/modules/set_metadata.cc
+++ b/core/modules/set_metadata.cc
@@ -189,13 +189,15 @@ inline void SetMetadata::DoProcessBatch(bess::PacketBatch *batch,
                                         const struct Attr *attr,
                                         mt_offset_t mt_offset) {
   if (mode == Mode::FromPacket) {
-    bool shift = attr->shift;
+    bool shift = attr->shift != 0;
     if (shift && attr->do_mask) {
       CopyFromPacket<true, true>(batch, attr, mt_offset);
     } else if (shift) {
-      CopyFromPacket<true>(batch, attr, mt_offset);
+      CopyFromPacket<true, false>(batch, attr, mt_offset);
+    } else if (!shift) {
+      CopyFromPacket<false, true>(batch, attr, mt_offset);
     } else {
-      CopyFromPacket<false>(batch, attr, mt_offset);
+      CopyFromPacket<false, false>(batch, attr, mt_offset);
     }
   } else {
     CopyFromValue(batch, attr, mt_offset);
--- a/core/modules/set_metadata.cc
+++ b/core/modules/set_metadata.cc
@@ -28,29 +28,44 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <cmath>
+#include <x86intrin.h>
+
 #include "set_metadata.h"
 
+#include "../utils/bits.h"
 #include "../utils/endian.h"
 
+using bess::metadata::mt_offset_t;
+
+template <bool do_shift = false, bool do_mask = false>
 static void CopyFromPacket(bess::PacketBatch *batch, const struct Attr *attr,
-                           bess::metadata::mt_offset_t mt_off) {
+                           mt_offset_t mt_off) {
   int cnt = batch->cnt();
   int size = attr->size;
-
+  int shift = attr->shift;
   int pkt_off = attr->offset;
 
   for (int i = 0; i < cnt; i++) {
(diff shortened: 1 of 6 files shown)


core/modules/set_metadata.cc
```diff
@@ -112,18 +130,38 @@ CommandResponse SetMetadata::AddAttrOne(
     if (offset < 0 || offset + size >= SNBUF_DATA) {
       return CommandFailure(EINVAL, "invalid packet offset");
     }
+    if (rshift_bytes * 8 != attr.rshift_bits()) {
+      return CommandFailure(EINVAL, "'rshift_bits' must be a multiple of 8");
+    }
+    if (static_cast<size_t>(std::abs(rshift_bytes)) > size) {
```
sangjinhan: I think it should be `>=` and `(..., ...)`, instead of `>` and `[..., ...]`

core/modules/set_metadata.cc
```diff
@@ -112,18 +130,38 @@ CommandResponse SetMetadata::AddAttrOne(
...
+      return CommandFailure(EINVAL, "'rshift_bits' must be in [-%zu,%zu]",
+                            8 * size, 8 * size);
+    }
+    if (do_mask) {
+      memset(mask.bytes, 0xFF, size);
```
sangjinhan: This `memset` doesn't do anything, since it is overwritten by `Copy()` below?

core/modules/set_metadata.cc
```diff
@@ -146,21 +184,39 @@ CommandResponse SetMetadata::Init(const bess::pb::SetMetadataArg &arg) {
...
+    mt_offset_t mt_offset = attr_offset(i);
 
     if (!bess::metadata::IsValidOffset(mt_offset)) {
       continue;
     }
 
-    /* copy data from the packet */
-    if (attr->offset >= 0) {
-      CopyFromPacket(batch, attr, mt_offset);
+    if (attr->offset > 0) {
```
sangjinhan: I think 0 is a valid offset for "copy from packet"

protobuf/module_msg.proto
```diff
@@ -828,6 +828,9 @@ message ReplicateCommandSetGatesArg {
 message SetMetadataArg {
   /**
    * SetMetadata Attribute describes a metadata attribute and value to attach to every packet.
+   * If copying data from a packet buffer, SetMetadata will optionally shift
```
sangjinhan: Perhaps it would be a good idea to clarify that it is logical shift (not arithmetic shift) somewhere

core/modules/set_metadata.cc
```diff
@@ -28,29 +28,44 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <cmath>
+#include <x86intrin.h>
```
sangjinhan: C headers should appear first

core/modules/set_metadata.cc
```diff
@@ -79,10 +97,10 @@ CommandResponse SetMetadata::AddAttrOne(
   }
   name = attr.name();
   size = attr.size();
+  do_mask = attr.mask().length() > 0;
```
sangjinhan: It would be useful if this function checks if mask and shift are specified only for "copy from packet/metadata" mode (for "copy from value" mask&shift are pointless)

sangjinhan:
Merged, thanks!
default-2

PR #603, opened 2017-08-09
set_metadata: make packet mode logic explicit

melvinw:
The existing version is hard to reason about and makes the no shift/no
mask code path unreachable.


sangjinhan:
zabott2mpeliptgtxfc0ke3vo1_500

PR #604, opened 2017-08-10
Revamp plugins

chris3torek:
This series of commits adds support for out-of-tree plugins (and out-of-tree module test code).

It's pretty complex, so it could use close scrutiny.  I have tried to rearrange it nicely into functional commits, so that each commit makes sense on its own (and each one is at least lightly tested).


.gitignore
```diff
@@ -44,12 +45,6 @@ deps/dpdk*
 # OSX
 .DS_Store
 
-# Protobuf
```
sangjinhan: Do you have any particular reason to make this per-directory?

(btw, `sample_plugin/protobuf` also needs .gitignore for these auto-generated files)

chris3torek: At first I thought I needed to keep empty sub-directories (`proto/`), and `.gitignore` would make them not-quite-empty. But I needed the `__init__.py` files anyway, so it turned out I didn't need `.gitignore` in each directory after all.

chris3torek: Hm, I didn't try building the sample_plugin using `--plugin`, maybe I should make sure that still works too.

chris3torek: Needs one slight fix to `sample_plugin/modules/sequential_update.h`:
```
-#include "protobuf/supdate_msg.pb.h"
+#include "pb/supdate_msg.pb.h"
```
The generated protobufs go in the `plugin_pb` directory as usual, though, so there's no need to `.gitignore` them.  I'll add the fix as a separate commit, at least for now.

(Note possible drawback: however many plugins you add, all go into `plugin_pb`.  But they also share one big module namespace in BESS itself, so ... 🤷‍♂️ it's no worse than it is for other reasons.)

PR #605, opened 2017-08-11
allow module .so's to be symbolic links

nrdmao:
If libtool is used to build libraries, then come in the form of .so.0.0.1 for example and then the .so is linked to the version specific .so. With this commit, bessd can load such symlinks. I also added some
logging to add diagnostics on modules loaded and failure to load modules.


changlan:
LGTM. Contributor License Agreement pending.

sangjinhan:
tumblr_mbtq4y62nc1qbe5pxo1_1280

PR #606, opened 2017-08-15
Container build fixes

chris3torek:
To build plugins when doing container builds, we need the container build code to export any outside-bess-repo plugins.

A sanity check (do all plugin directories actually exist) also seems like a good idea, but it's a separate commit since it's not required to make the container build work.

### DIFF
--- a/container_build.py
+++ b/container_build.py
@@ -36,12 +36,55 @@ import sys
 import subprocess
 import os
 import os.path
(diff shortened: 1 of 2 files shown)


sangjinhan:
👍 
ul6cj8gfwlku96dvoj8hamobo1_1280

PR #607, opened 2017-08-15
build.py: fix up the fixup for the DPDK remote

chris3torek:
In commit c3aa0bc5fd6d4b1c5e7f4cc4a7eb15d52ac8eeb1 we did a
shallow clone of the DPDK repository, then did a fixup step
so that it was easy to un-shallow the repository later.

The fixup step was wrong ("git config remote.url.fetch" instead
of "git config remote.origin.fetch").

This only affects anyone intending to update DPDK, but let's
clean it up (and fix the pep8 complaint about the line length).


sangjinhan:
tumblr_mcpa8zbbiz1qze0hyo1_1280

PR #609, opened 2017-08-18
Fix coverage

chris3torek:
The `-E` thing explains why coverage was getting weird.

I put in the correct symlink (as a separate commit) per Sangjin's request.


sangjinhan:
zabott2mpju4nquhmgky2hyio1_1280

PR #613, opened 2017-08-21
drivers/unix_socket: Let clients to freely connect/disconnect

llvilanova:
In order to allow external controllers (connected through a UNIX socket) truly
independent of bessd, client disconnection must always be detected, regardless
of whether the port is used or not. This is done by keeping a thread
around (using epoll).

Before, this driver would only reliably detect disconnects when the port is used
to receive packets. When used to send packets, the port would only detect
disconnects on an actual send, which depends on the processed packets; this led
to only some instances accepting new clients, making external program
reconnection impossible.

### DIFF
--- a/core/drivers/unix_socket.cc
+++ b/core/drivers/unix_socket.cc
@@ -28,6 +28,8 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <sys/epoll.h>
+
 #include "unix_socket.h"
 
 // TODO(barath): Clarify these comments.
@@ -43,48 +45,63 @@
 void UnixSocketPort::AcceptNewClient() {
   int ret;
 
-  for (;;) {
-    ret = accept4(listen_fd_, nullptr, nullptr, SOCK_NONBLOCK);
-    if (ret >= 0) {
+  while (true) {
+    for (;;) {
+      ret = accept4(listen_fd_, nullptr, nullptr, SOCK_NONBLOCK);
+      if (ret >= 0) {
+        // New client while we already have an active one; drop connection
+        if (client_fd_ != kNotConnectedFd) {
+          close(ret);
+        }
+        break;
+      }
+
+      if (errno != EINTR) {
+        PLOG(ERROR) << "[UnixSocket]:accept4()";
+      }
+    }
+
+    recv_skip_cnt_ = 0;
+    client_fd_ = ret;
+
+    // Detect when connection is dropped by client
+    struct epoll_event event;
+    event.events = EPOLLRDHUP;
+    ret = epoll_ctl(epoll_fd_, EPOLL_CTL_ADD, ret, &event);
+    if (ret < 0) {
+      PLOG(ERROR) << "[UnixSocket]:epoll_ctl()";
       break;
     }
 
-    if (errno != EINTR) {
-      PLOG(ERROR) << "[UnixSocket]:accept4()";
+    while (true) {
+      ret = epoll_wait(epoll_fd_, &event, 1, -1);
+      if (ret < 0) {
+        if (errno == EINTR) {
+          continue;
+        } else {
+          PLOG(ERROR) << "[UnixSocket]:epoll_wait()";
+          break;
+        }
+      }
+
+      if (event.events & EPOLLRDHUP) {
+        break;
+      }
     }
-  }
 
-  recv_skip_cnt_ = 0;
-
-  if (old_client_fd_ != kNotConnectedFd) {
-    // Reuse the old file descriptor number by atomically exchanging the new fd
-    // with the
-    // old one.  The zombie socket is closed silently (see dup2).
-    dup2(ret, client_fd_);
-    close(ret);
-  } else {
-    client_fd_ = ret;
+    // Close dropped connection (Send/Recv are resilient)
+    epoll_ctl(epoll_fd_, EPOLL_CTL_DEL, client_fd_, &event);
+    close(client_fd_);
+    client_fd_ = -1;
   }
 }
 
 // This accept thread terminates once a new client is connected.
 void *AcceptThreadMain(void *arg) {
+  // TODO: should make sure it's pinned to a non-worker CPU
   UnixSocketPort *p = reinterpret_cast<UnixSocketPort *>(arg);
   p->AcceptNewClient();
   return nullptr;
-}
-
-// The file descriptor for the connection will not be closed, until we have a
-// new client.
-// This is to avoid race condition in TX process.
-void UnixSocketPort::CloseConnection() {
-  // Keep client_fd, since it may be being used in unix_send_pkts().
-  old_client_fd_ = client_fd_;
-  client_fd_ = kNotConnectedFd;
-
-  // Relaunch the accept thread.
-  std::thread accept_thread(AcceptThreadMain, reinterpret_cast<void *>(this));
-  accept_thread.detach();
 }
 
 CommandResponse UnixSocketPort::Init(const bess::pb::UnixSocketPortArg &arg) {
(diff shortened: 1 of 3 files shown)


llvilanova:
I'm not sure how this patch affects the failed test. Any ideas?

sangjinhan:
Seems like the failure happens with 32bit build and it is reproducible. Trying to debug it...

llvilanova:
Is there some way for me to easily test the failing experiment on that build?

Also, is it possible that `DeInit` is being called before `Init`? (I've checked this never happens when `Init` returns a command error) That could trigger a file descriptor `close` before properly setting the descriptors (they should all be zero from the constructor).

llvilanova:
I just did a cleanup to ensure there's no threads or FDs left behind when a module is destroyed, but it still shows the same failure.

If you give me a command line to reproduce the failing build and specific test I'll give it a shot in the debugger.

sangjinhan:
I don't know how to reliably reproduce the issue locally. What I did was using Travis with my private fork of this repo. The issue seems to happen only with 32bit builds.

What I realized is that `bessctl/conf/testing/module_tests/vlan.py` opens lots of (more precisely, 450) UnixSocketPort. Given that each port now opens three (not two) file descriptors, my theory is that it may exceed the file descriptor limit of 1024. For processes launched by root, maybe this limit only applies to 32bit programs. Could you update the file as below?

```diff
-        return [VLANSplit(), 1, 150, expected]
+        return [VLANSplit(), 1, 30, expected]

-    OUTPUT_TEST_INPUTS.append(output_test([1, 100, 77, -1, 149, 50, 100, -1]))
-    OUTPUT_TEST_INPUTS.append(output_test([100, 77, -1, 149, 50, 100, -1, 33, 70]))
-    OUTPUT_TEST_INPUTS.append(output_test([100, 77, -1, 149, 50, 100, -1, 33, 70], True))
+    OUTPUT_TEST_INPUTS.append(output_test([1, 17, -1, 29, 10, 13, 7]))
+    OUTPUT_TEST_INPUTS.append(output_test([1, 17, -1, 29, 10, 13, 7], True))
```

sangjinhan:
Two requests regarding file descriptor usage:
* Instead of using `epoll`, could you switch to `select` or `poll`?
* Instead of using `pipe`, you can terminate the thread with `pthread_cancel` or `pthread_kill`
  * Let's have a separate function for the thread, rather than a large lambda block.

llvilanova:
Can't use `poll` because wakeup conditions apply to all FDs (therefore the thread will be woken up too often - i.e., every time there's data on the client socket, not only when it's closed, since we need to wait for available reads on the accept socket).

Can't use `select` either because socket disconnect can only be detected together with available read events (i.e., will wake up too often when the client socket is readable, not only when it's closed).

So, is it OK to keep it with `epoll`?

sangjinhan:
I am not sure if I understood correctly. You can add `client_fd_` only to `exceptfds` for `select()`, or set it only with `POLLRDHUP` for `poll()`? Just like `epoll`, you should be able to specify what events you want to monitor for individual file descriptors. I vaguely remember that all `select/poll/epoll` are implemented with the same backend code in the kernel.

Maybe I am missing something here. If you find `epoll` is the only way to reliably detect disconnection of `client_fd_`, please go ahead.

llvilanova:
Ooops, sorry. I read `poll` too quickly and misunderstood it; yes it's a suitable replacement to `epoll_wait`.

For the record, the limitations on `select` still stand: there is no equivalent to `EPOLLHUP`/`POLLHUP` in `select`, and one must instead add a FD into `readfds` and check if a read returns zero (overkill for this FD monitoring thread).

core/drivers/unix_socket.cc
```diff
@@ -38,53 +41,63 @@
...
-  }
+  struct pollfd fds[2];
+  memset(fds, 0, sizeof(fds));
+  fds[0].fd = listen_fd_;
+  fds[0].events = POLLIN;
+  fds[1].events = POLLRDHUP;
 
-  recv_skip_cnt_ = 0;
+  while (true) {
+    fds[1].fd = client_fd_;
```
sangjinhan: Would you add a comment that negative file descriptors are ignored by `poll()` for clafirication?

core/drivers/unix_socket.cc
```diff
@@ -38,53 +41,63 @@
...
-// This accept thread terminates once a new client is connected.
-void *AcceptThreadMain(void *arg) {
-  UnixSocketPort *p = reinterpret_cast<UnixSocketPort *>(arg);
-  p->AcceptNewClient();
-  return nullptr;
+    } else if (res < 0) {
+      if (errno == EINTR) {
+        continue;
+      } else {
+        PLOG(ERROR) << "[UnixSocketPort]:epoll_wait()";
```
sangjinhan: Feel free to ignore this: you don't need to add `[UnixSocketPort]:` to the log message since the filename will be automatically added by `PLOG`. Also, `epoll_wait()` -> `ppoll()`

core/drivers/unix_socket.cc
```diff
@@ -38,53 +41,63 @@
...
+        LOG(WARNING) << "[UnixSocketPort]: Ignoring additional client\n";
+        close(fd);
+      } else {
+        client_fd_ = fd;
+      }
+
+    } else if (fds[1].revents & (POLLRDHUP | POLLHUP)) {
+      // connection dropped by client
+      int fd = client_fd_;
+      client_fd_ = -1;
```
sangjinhan: `kNotConnectedFd` in place of `-1`

core/drivers/unix_socket.cc
```diff
@@ -38,53 +41,63 @@
...
-// This is to avoid race condition in TX process.
-void UnixSocketPort::CloseConnection() {
-  // Keep client_fd, since it may be being used in unix_send_pkts().
-  old_client_fd_ = client_fd_;
-  client_fd_ = kNotConnectedFd;
-
-  // Relaunch the accept thread.
-  std::thread accept_thread(AcceptThreadMain, reinterpret_cast<void *>(this));
-  accept_thread.detach();
+static void AcceptThreadHandler(int sig) {
```
sangjinhan: Minor: You can just omit the argument name (just `static void AcceptThreadHandler(int) {`) to indicate the argument will be unused.

core/drivers/unix_socket.cc
```diff
@@ -130,32 +141,54 @@ CommandResponse UnixSocketPort::Init(const bess::pb::UnixSocketPortArg &arg) {
...
+  struct sigaction sa;
+  memset(&sa, 0, sizeof(sa));
+  sa.sa_handler = AcceptThreadHandler;
+  if (sigaction(SIG_THREAD_EXIT, &sa, NULL) < 0) {
+    DeInit();
+    return CommandFailure(errno, "sigaction(SIG_THREAD_EXIT) failed");
+  }
+
+
+  accept_thread_ = std::thread([&]() {
```
sangjinhan: `[this]` can be used in place of `[&]`.

core/drivers/unix_socket.cc
```diff
@@ -28,6 +28,9 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <signal.h>
```
sangjinhan: `#include <poll.h>` should appear first.

core/drivers/unix_socket.h
```diff
@@ -32,6 +32,7 @@
 #define BESS_DRIVERS_UNIXSOCKET_H_
 
 #include <assert.h>
+#include <atomic>
```
sangjinhan:
```
#include <poll.h>

#include <atomic>
#include <cassert>
#include <cerrno>
#include <cstdio>
```

The Google C++ Style Guide says all C++ headers should appear after C headers. Each group needs to be alphabetically sorted.

Do we need `cassert` and `cstdio`? Also, `poll.h` and `cerrno` seem only necessary in the .cc file, not in the header.

sangjinhan:
Merged, thanks!
tumblr_m44vnvr00w1rtuomto1_1280

Issue #612, opened 2017-08-20
Command "show driver <name>" does not work


llvilanova:
If I run `show driver <name>` in `bessctl`, I get the following error for `show driver PMDPort` (same error for any driver, existing or not):

<pre>
*** Error: No driver 'P' found
  BESS daemon response - errno=2 (ENOENT: No such file or directory)
       query: GetDriverInfo
   query_arg: {'driver_name': 'P'}
</pre>

Looks like the name is being truncated to the first letter of the string argument somewhere.

chris3torek:
fixed by 1b9db04 (need to figure out how to make commits auto-close issues on github)

PR #614, opened 2017-08-21
Fix minor CLI issue (#612)

chris3torek:
(driver name was mis-classified as `name` instead of `name+`)


chris3torek:
Wow, codecov is picky!  Minus less than .01%, and it's just a random chance one I think (not sure how one gets EINTR there). :roll_eyes:

changlan:
Merged, thanks!

image

PR #615, opened 2017-08-22
Allocate more memory with --no-huge

ddiproietto:
This fixes BESS with `--no-huge`, which can be useful for testing.

Also, this PR refactors DPDK args creation and uses more of C++.  I think it's a little bit more readable this way.


sangjinhan:
tumblr_m45vycszfz1qze0hyo1_1280

PR #616, opened 2017-08-24
module tests: pause workers after custom tests

chris3torek:
When we run custom tests, the custom-test loop reads:

    for each test:
        pause workers
        reset pipeline
        run test

so that each one starts with a fresh context.  But custom
tests, especially if they fail, may leave workers running.

Meanwhile the main loop expects not to need to pause, it just
does a pipeline reset:

    for each module:
        reset pipeline
        (run all the module tests)

To keep the number of rpc's down do one extra "pause" after
all custom tests if and only if there are any custom tests.
It would be simpler to do a pause at the top of the "for each
module" loop, though...

(While here, trim out unrelated now-irrelevant comment about
scope, fixed in c524bc8f85a9fa32d49654d59e2f223498731a2b.)

### DIFF
--- a/bessctl/conf/testing/run_module_tests.bess
+++ b/bessctl/conf/testing/run_module_tests.bess
@@ -169,7 +169,6 @@ def run_tests():
 
         print("-- Running Module Tests for %s --" % filename.split("/")[-1])
 
-       # Wrap in a function to make sure scope doesn't leak.
         CRASH_TEST_INPUTS, OUTPUT_TEST_INPUTS, CUSTOM_TEST_FUNCTIONS = load_test(
             filename)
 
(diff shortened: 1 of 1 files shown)


chris3torek:
Kevin had what I think is a good suggestion: we should have an RPC to do both (pause-all-reset), perhaps an optional pause-all flag in the reset RPC. This would cut the number of RPCs in half for the usual case. I'm not sure about backwards-compatiblity issues though.

sangjinhan:
Perhaps the easiest way is moving the `bess.pause_all()` at the beginning of `for test in CUSTOM_TEST_FUNCTIONS:` loop to the end? At least it is what crash/output tests seem to do... I think the intended invariant for each per-type test loop is that the pipeline has been already paused before start, and after end.

sangjinhan:
tumblr_mbwd5z9swb1qhwmnpo1_1280

PR #618, opened 2017-08-24
Makefile: Use pkg-config if available.

ddiproietto:
Debian buster(testing) and presumably ubuntu 17.10 package all the
dependencies required to build BESS (except dpdk).

New versions of GRPC (packaged in debian) require libc-ares and if BESS
is statically linked (by default it is), `-lcares` must be passed to the
linker.

Instead of forcing a new dependency for all users (even the ones that don't
need it), this commit makes use of pkg-config, which will automatically
include the indirect dependencies (such as libc-ares).

For users that have an old grpc version (such as `container_build.py`),
pkg-config is still optional.  I'd like to make it required, but a new
container image must be generated, and that's a decision we can make
later.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -34,6 +34,7 @@ MAKEFLAGS += --no-builtin-rules
 
 CXX ?= g++
 PROTOC ?= protoc
+PKG_CONFIG ?= pkg-config
 
 VERBOSE ?= 0
 
@@ -57,6 +58,8 @@ $(error g++ 5 or higher is required. Use container_build.py if newer g++ is not 
 endif
 endif
 
+HAS_PKG_CONFIG := $(shell command -v $(PKG_CONFIG) 2>&1 >/dev/null && echo yes || echo no)
+
 RTE_SDK ?= $(abspath ../deps/dpdk-17.05)
 RTE_TARGET ?= $(shell uname -m)-native-linuxapp-gcc
 DPDK_LIB ?= dpdk
@@ -73,6 +76,28 @@ else ifeq ($(words $(MAKECMDGOALS)),1)
 	$(error DPDK is not available. \
 		Make sure $(abspath $(RTE_SDK)) is available and built)
 	endif
+endif
+
+# We always want these libraries to be dynamically linked even when the
+# user requests a static build.  These needs to be listed first, so even
+# if they're listed again in the static section, they will be dynamically
+# linked.
+ALWAYS_DYN_LIBS := -lpthread -ldl
+# These libraries are not supported by pkg-config.
+ALWAYS_LIBS := -lpcap
+# If pkg-config is available, we just need a list of the dependecies.
+PKG_CONFIG_DEPS := libglog gflags protobuf grpc++ libunwind zlib
+# If pkg-config is not available, we need to list the libs we depend on.
+NO_PKG_CONFIG_LIBS := -lglog -lgflags -lprotobuf -lgrpc++ -lunwind -lz
+# If pkg-config is not available and we're static linking, we also need
+# the indirect dependecies.  This is annoying, because they may change
+# in future versions.
+NO_PKG_CONFIG_LIBS_INDIRECT := -lgrpc -lssl -lcrypto -llzma
+
+ifeq ($(HAS_PKG_CONFIG), yes)
+    PKG_CFLAGS = $(shell $(PKG_CONFIG) --cflags $(PKG_CONFIG_DEPS))
+else
+    PKG_CFLAGS =
 endif
 
 # Plugins get to look in $COREDIR through -I $(COREDIR).  Let them also
@@ -93,7 +118,7 @@ CXXFLAGS += -std=c++11 -g3 -ggdb3 $(CXXARCHFLAGS) \
 	    -isystem $(dir $<).. -isystem $(COREDIR)/modules \
 	    -D_GNU_SOURCE \
 	    -Werror \
-	    -Wall -Wextra -Wcast-align
+	    -Wall -Wextra -Wcast-align $(PKG_CFLAGS)
 
 PERMISSIVE := -Wno-unused-parameter -Wno-missing-field-initializers \
 	      -Wno-unused-private-field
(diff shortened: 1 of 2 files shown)


core/Makefile
```diff
@@ -75,6 +78,28 @@ else ifeq ($(words $(MAKECMDGOALS)),1)
...
+PKG_CONFIG_DEPS := libglog gflags protobuf grpc++ libunwind zlib
+# If pkg-config is not available, we need to list the libs we depend on.
+NO_PKG_CONFIG_LIBS := -lglog -lgflags -lprotobuf -lgrpc++ -lunwind -lz
+# If pkg-config is not available and we're static linking, we also need
+# the indirect dependecies.  This is annoying, because they may change
+# in future versions.
+NO_PKG_CONFIG_LIBS_INDIRECT := -lgrpc -lssl -lcrypto -llzma
+
+ifeq ($(HAS_PKG_CONFIG), yes)
+    PKG_CFLAGS = $(shell $(PKG_CONFIG) --cflags $(PKG_CONFIG_DEPS))
```
chris3torek: Can this be statically defined (with `:=`)?

ddiproietto: Thanks for the suggestion! I did that here and in the other places I was shelling out to call `pkg-config`

core/Makefile
```diff
@@ -75,6 +78,28 @@ else ifeq ($(words $(MAKECMDGOALS)),1)
 	endif
 endif
 
+# We always want these libraries to be dynamically linked even when the
+# user requests a static build.  These needs to be listed first, so even
+# if they're listed again in the static section, they will be dynamically
+# linked.
+ALWAYS_DYN_LIBS := -lpthread -ldl
```
sangjinhan: Is it necessary to add `-lpthread` here? The `-pthread` option already implies this...

core/Makefile
```diff
@@ -112,20 +137,28 @@ LDFLAGS += -rdynamic -L$(DPDK_LIB_DIR) -Wl,-rpath=$(DPDK_LIB_DIR) -pthread
...
-LIBS += -Wl,-non_shared \
+LIBS += $(ALWAYS_DYN_LIBS) -Wl,-non_shared \
 	-Wl,--whole-archive -l$(DPDK_LIB) -Wl,--no-whole-archive \
 	$(LIBS_ALL_SHARED) \
-	-lglog -lgflags -lprotobuf -lgrpc++ -lgrpc \
-	-lssl -lcrypto -lunwind $(LIBS_LZMA) -lpcap -lz \
+	$(PKG_LIBS) $(ALWAYS_LIBS) \
 	$(LIBS_DL_SHARED) \
-	-ldl
+	$(ALWAYS_DYN_LIBS)
```
sangjinhan: Just checking: Adding `$(ALWAYS_DYN_LIBS)` twice was intended?

ddiproietto: It was intended.

DPDK depends on `-ldl`, so `-ldl` must be specified after DPDK.

But then again, we want to dynamically link with `-ldl`.  If one of the other libs depend on `-ldl` it will be listed in the static section.  We must make sure that we include it at the beginning of the cmdline, in the dynamic section

core/Makefile
ddiproietto: I believe it's necessary.

grpc includes `-lpthread` in its statically required libraries.  If we don't specify `-lpthread` before (in the dynamic section), -lpthread will be linked statically, and that's not well supported apparently

sangjinhan:
Great, thanks!
tumblr_m7wfqrq8gw1qznh2ao1_1280

Codeacious:
This pull request appears to have broken static builds for me on Ubuntu 16.04 LTS (kernel 4.4.0-93-generic):
```
/usr/local/home/rotor/bess/deps/dpdk-17.05/build/lib/librte_pmd_ark.a(ark_ethdev.o): In function `check_for_ext':
/usr/local/home/rotor/bess/deps/dpdk-17.05/drivers/net/ark/ark_ethdev.c:199: warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libdl.a(dlopen.o): In function `dlopen':
(.text+0x5): undefined reference to `__dlopen'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libdl.a(dlclose.o): In function `dlclose':
(.text+0x1): undefined reference to `__dlclose'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libdl.a(dlsym.o): In function `dlsym':
(.text+0x5): undefined reference to `__dlsym'
/usr/lib/gcc/x86_64-linux-gnu/5/../../../x86_64-linux-gnu/libdl.a(dlerror.o): In function `dlerror':
(.text+0x1): undefined reference to `__dlerror'
collect2: error: ld returned 1 exit status
Error: port_test
g++ -o port_test port_test.o gtest-all.o gtest_main.o bess.a -rdynamic -L/usr/local/home/rotor/bess/deps/dpdk-17.05/build/lib -Wl,-rpath=/usr/local/home/rotor/bess/deps/dpdk-17.05/build/lib -pthread -static-libstdc++ -lpthread -ldl -Wl,-non_shared -Wl,--whole-archive -ldpdk -Wl,--no-whole-archive -L/usr/local/lib -lglog -lprotobuf -pthread -lpthread -lz -lgrpc++ -lgrpc -lssl -ldl -lcrypto -ldl -lunwind -llzma -lz -lpcap -lgflags -Wl,-call_shared -lpthread -ldl -libverbs
Makefile:394: recipe for target 'port_test' failed
```
The full output is much longer, but are duplicates of this error on other files.

The issue seems to be that `pkg-config --libs --static grpc++` returns `-L/usr/local/lib -lgrpc++ -lgrpc -lz -lssl -ldl -lcrypto -ldl`, meaning g++ attempts to use symbols from the static library `libdl.a` instead of `libdl.so` when linking DPDK. Manually running the linking command without the additional `-ldl` flags returned by `pkg-config` solves the issue.

An easy, but hacky, way to do this in the Makefile is to use `sed`:
`PKG_LIBS := $(shell $(PKG_CONFIG) --static --libs $(PKG_CONFIG_DEPS) | sed -e s/-ldl//g)`
But there's probably better method to solve this issue.

ddiproietto:
Thanks for reporting this and sorry for the breakage.

I wasn't able to reproduce it on ubuntu 16.04, but your suggestion makes sense to me (I like it better than listing twice the libraries).

I submitted PR #626, I removes `-ld` from the static libraries, I hope it fixes the problem for you.

Thanks

PR #619, opened 2017-08-25
[trivial] .gitignore: ignore Vagrant temporaries

chris3torek:
Restore dot-file ignore lost in the general gitignore
reduction in bdfb282260791ac78372c5149de38d50e1cf32ab.


melvinw:
Thanks! cat

PR #620, opened 2017-08-25
modules: Thread safety checker fixes.

ddiproietto:
VLANSplit is stateless and can be run from multiple threads.

Queue contains a multiproducer ring.  It is single consumer, but since
the module registers only one task that shouldn't be a problem.

PortInc supports more than one task, and that needs to be taken into
account.


melvinw:
Thanks! hcufecc

PR #621, opened 2017-08-27
Make path for module tests configurable

sangjinhan:
This commit introduces a new env variable `TEST_PATH`, to specify custom module test directory other than `conf/testing/module_tests`.

### DIFF
--- a/bessctl/conf/testing/run_module_tests.bess
+++ b/bessctl/conf/testing/run_module_tests.bess
@@ -38,9 +38,6 @@ import sugar
 import traceback
 from time import gmtime, strftime
 
-# Env variable used to run a single test
-test_name = $TEST_NAME!'*'
-
 ## HANDY UTILS AND GLOBAL VARS ##
 SOCKET_PATH = '/tmp/bess_unix_'
 SCRIPT_STARTTIME = strftime("%Y-%m-%d-%H-%M-%S", gmtime())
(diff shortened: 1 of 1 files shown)


bessctl/conf/testing/run_module_tests.bess
```diff
@@ -156,15 +153,19 @@ def monitor_task(module, wid):
 def run_tests():
     DID_TESTS_FAIL = False
 
-    path = os.path.dirname(os.path.realpath(sys.argv[0]))
+    # Env variable used to run a subset of tests
+    test_name = $TEST_NAME!'*'
+
+    default_path = os.path.join(os.path.dirname(os.path.realpath(sys.argv[0])),
```
chris3torek: It might be good to use `os.path.abspath` rather than `os.path.realpath` in case something changes directories at some point, but this should be fine for now.

bessctl/conf/testing/run_module_tests.bess
```diff
@@ -156,15 +153,19 @@ def monitor_task(module, wid):
...
+                                'conf/testing/module_tests')
+    test_path = $TEST_PATH!default_path
```
chris3torek: At some point we might want to modify this to allow more than one path name component (split at `:` like `$PATH`, perhaps) so that if you have several plugin directories you can point to all of them at once.  But again, this should be fine for now.

melvinw: Unless you want to address this one in this PR @sangjinhan, I'll merge now

melvinw:
Thanks! catsitweird

PR #622, opened 2017-08-27
Add qid to QueueOut::GetDesc()

sangjinhan:
Unlike QueueInc module, `show pipeline` does not show which `qid` a QueueOut module is associated with. This commit fixes it.


melvinw:
Thanks! cat

PR #623, opened 2017-08-27
Fix RoundRobin crash with zero output gate

sangjinhan:
If RoundRobin module has no output gate, it crashes due to division by zero. In this case, the "right" behavior should be just dropping packets.

Also, this commit replaces the modulo(%) operation with a simple `if` branch. It saves about 22 cycles per batch or packet, depending on the operation mode of RoundRobin.


melvinw:
Thanks! cat

PR #624, opened 2017-08-28
Print out BESS envvar warnings to stderr

sangjinhan:
`Environment variable XXX is not set. Using default value YYY` better be sent to stderr, so that external scripts can read stdout for "real" output.


melvinw:
Thanks! cat

PR #627, opened 2017-08-30
bessctl: take DNS names and IPs in daemon connect

melvinw:
`daemon connect` currently accpets `name` tokens for its HOST argument.
`name` is overly restrictive, so this commit add a new token called
`host` which matches DNS names and IPv4 addresses.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -386,7 +386,7 @@ def get_var_attrs(cli, var_token, partial_word):
             var_desc = 'bess daemon command-line options (see "bessd -h")'
 
         elif var_token == '[HOST]':
-            var_type = 'name'
+            var_type = 'host'
             var_desc = 'host address'
 
         elif var_token == '[TCP_PORT]':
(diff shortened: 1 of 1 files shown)


bessctl/commands.py
```diff
@@ -476,6 +477,13 @@ def bind_var(cli, var_type, line):
                 raise cli.BindError('"wid" must be a positive number')
         val = sorted(list(set(val)))
 
+    elif var_type == 'host':
+        dns = re.match(r'^[a-zA-Z0-9][a-zA-Z0-9\-.]*$', val)
```
chris3torek: This is overly forgiving, e.g., it accepts `host.dom..ain`, but I don't think a full-name-compliance-checker is warranted here unless there's one handy in a library (is there?). Presumably we don't care to allow Unicode IDN input 😺

(pointless thoroughness: there are some sample Python validators in https://stackoverflow.com/questions/2532053/validate-a-hostname-string)

bessctl/commands.py
```diff
@@ -476,6 +477,13 @@ def bind_var(cli, var_type, line):
...
+        ip = re.match(r'^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$', val)
```
chris3torek: What, no IPv6? 😈

sangjinhan:
tumblr_mbt2tzisfz1qznh2ao1_1280

PR #626, opened 2017-08-30
Makefile: Fix build on ubuntu 16.04 with pkg-config.

ddiproietto:
Instead of listing twice the libraries that we want to link dynamically
with, we can simply filter those out from the result of pkg-config.

This should fix the build in ubuntu 16.04 (although I wasn't able to
reproduce the problem).

Reported and suggested by @Codeacious

PR #629, opened 2017-09-01
Launch BESS daemon per testcase, not every test

sangjinhan:
https://github.com/NetSys/bess/pull/626#issuecomment-326513133


ddiproietto:
LGTM as well, thanks for doing this

melvinw:
Thanks! cat

PR #630, opened 2017-09-02
worker: add thread-local instance of Random

melvinw:
@sangjinhan and @ddiproietto pointed out that this may be useful

### DIFF
--- a/core/worker.cc
+++ b/core/worker.cc
@@ -49,6 +49,7 @@
 #include "opts.h"
 #include "packet.h"
 #include "scheduler.h"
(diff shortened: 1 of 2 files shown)


core/worker.cc
```diff
@@ -49,6 +49,7 @@
 #include "opts.h"
 #include "packet.h"
 #include "scheduler.h"
+#include "utils/random.h"
```
sangjinhan: This line seems redundant.

sangjinhan:
tumblr_mbskalxgn81qhwmnpo1_1280

PR #632, opened 2017-09-04
Mark get_summary in measure.cc as thread safe

apanda:
`get_summary` reads a value, and should be thread safe to execute (the
values might be off though).


barath:
Is it possible that the return value will not just be stale but total junk?

apanda:
My reading of the function was no -- things like number of packets, and size of packets will be correct. The histogram might report some inaccuracies, but they will be small.

barath:
Ok, LGTM in that case.

sangjinhan:
tumblr_mbsqv4nt6y1qexe0no1_1280

PR #637, opened 2017-09-06
Kill container by name, but not by python subprocessor

shinae-woo:
Container daemons are not children of `docker run`, but children of docker daemon. Therefore, killing `docker run` does not always guarantee killing container daemons

### DIFF
--- a/bessctl/conf/port/vhost/launch_container.py
+++ b/bessctl/conf/port/vhost/launch_container.py
@@ -58,6 +58,7 @@ VERBOSE = int(os.getenv('VERBOSE', '0'))
 
 SOCKDIR = '/tmp/bessd'
 IMAGE = 'nefelinetworks/bess_build'
+CONTAINER_NAME = 'nefeli_bessd'
 
 
 def launch(cid):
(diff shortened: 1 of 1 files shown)


chris3torek:
Aha, reading up on docker I see you're right. But `docker kill <name>` is a bit severe (sends `SIGKILL`), would it be reasonable to use `docker stop` which sends `SIGTERM` first, then waits 10 seconds by default, then resorts to `SIGKILL`?

(I will need to do the same in my test code)

sangjinhan:
For the sake of demonstration, as an example script, personally I would prefer simplicity over completeness. As `testpmd` does not make any side effects I guess SIGKILL should be fine...?

chris3torek:
OK, in that case I'd say this looks fine.

bessctl/conf/port/vhost/launch_container.py
```diff
@@ -115,14 +130,11 @@ def main(argv):
...
     except KeyboardInterrupt:
         pass
-    finally:
-        for proc in procs:
-            print('Terminating container (pid=%d)' % proc.pid)
-            proc.kill()
-            proc.wait()
 
-    return 0
+    for cid in range(num_containers):
```
sangjinhan: Minor: whether to have this loop in the `finally` block or not would exhibit slightly different behaviors, in case the script has not been interrupted by Ctrl+C but something else; the containers won't be terminated.

shinae-woo: Right

shinae-woo:
@chris3torek For me, SIGKILL would be fine, too. (Waiting 10 sec would be bothersome in my specific 'development' environments), But if you need more graceful termination, please fix them :) As now, they are `--rm` containers, my first thought was who cares? But soon later, in some environments (scale-in in my case), someone may need graceful teardown such as serving all flows in queues.

bessctl/conf/port/vhost/launch_container.py
```diff
@@ -99,6 +100,20 @@ def launch(cid):
...
+
+    cmd = 'docker kill {name}'.format(name=CONTAINER_NAME + str(cid))
+
+    if VERBOSE:
+        print(cmd)
+
+    try:
+        proc = subprocess.check_call(shlex.split(cmd), stdout=subprocess.PIPE)
+    except subprocess.CalledProcessError:
+        pass
```
chris3torek: Noticed in passing: you could just use `subprocess.call`, since we throw out the exception raised in `check_call` here.

chris3torek:
`SIGKILL` suffices for me too but it's exactly that kind of thing (draining a queue) that I was thinking about.

I like the extra exception-case cleanup with the code in `finally`. (There's some other cleanup that my own code needs, e.g., removing the host-side `/tmp/bessd` directory, that makes it more complicated and I haven't made it all work for every case yet.)

sangjinhan:
Thanks!
tumblr_m6v0t66uvj1qh66wqo1_1280

bessctl/conf/port/vhost/launch_container.py
shinae-woo: + 1

PR #639, opened 2017-09-06
Check if kernel headers installed

shinae-woo:
- DPDK and BESS kernel modules have dependencies on it

This is motivated by to make DPDK and BESS run on my AWS instances as they only support virtual functions (e.g., ixgbevf), which only supports igb_uio driver. So after short discussion with @sangjinhan, I checked if kernel header installed to configure DPDK correctly and also for the BESS kernel module.

+ I would prefer not to maintain deps/dpdk-*_common_linuxapp file, But I don't have enough history regarding this matter.


build.py
```diff
@@ -203,6 +205,28 @@ def set_config(filename, config, new_value):
             fp.write(line)
 
 
+def is_kernel_header_installed():
+    try:
+        cmd = 'dpkg-query -W -f \'${Status}\' linux-headers-%s' \
```
nrdmao: Won't this change make build.py ubuntu dependent, which up to this point it has not been? 

Can this be driven from the KERNELDIR shell variable as is done in core/kmod/Makefile. If KERNELDIR is set or if 
/lib/modules/$(uname -r)/build exist, then build the modules.

Note, I do like the existing method whereby a local build environment can edit dpdk-17.05-common_linux as they see fit for their environment. I use IGB_UIO but I do not use KNI and would prefer not to build it.

shinae-woo: You are right. No dependency would be definitely better. I saw so many helper scripts using apt-get, but I have noticed that they are not in the critical path.

sangjinhan:
Merged, thanks.
tumblr_m06u5dabdl1rqfgw9o1_1280

PR #640, opened 2017-09-11
Run pip install with --user

sangjinhan:
Travis stopped working with `pip install`, as virtualenv is not used for
non-Python repos any longer. This patch runs pip with `--user` option
explicitly (`sudo pip install` seemed also working, but it is generally
not recommended).


chris3torek:
image

PR #641, opened 2017-09-11
module: fix RunCommand thread saftey check

melvinw:
(`active_workers_` is a bitmap, not a vector of worker IDs)


sangjinhan:
Thanks!
tumblr_m6u9ncz7lf1qddbvio1_1280

PR #642, opened 2017-09-12
Implement a simple bessctl command to capture packet data and metadata

ddiproietto:
The new `tcpdump_md` command is similar to `tcpdump`, but:
* It displays a textual representation of the metadata
* It doesn't call external programs, because they don't support showing additional metadata for each packet.
* It doesn't display a human readable representation of the packet data, just a hexdump, since it doesn't rely on external tools.

To implement this new command, a new hook is implemented (`pcapng`), which is similar to the `tcpdump` hook, but
* Uses the pcapng format, which includes optional metadata.
* Adds a textual representation of the metadata info.
Hopefully this hook can be useful beyond the tcpdump_md command.  It has been tested with wireshark and it produces a textual representation of the metadata.

### DIFF
--- a/core/hooks/pcapng.cc
+++ b/core/hooks/pcapng.cc
@@ -34,6 +34,8 @@
 #include <sys/uio.h>
 #include <unistd.h>
 
+#include <limits>
+
 #include <glog/logging.h>
 
 #include "../message.h"
@@ -58,12 +60,37 @@ T PadSize(T a, T b) {
   return RoundUp(a, b) - a;
 }
 
+// Return the single hex digit representing `nibble`.  If it cannot be
+// represented, return the char 'X'.
+char NibbleToHD(char nibble) {
+  if (nibble >= 0 && nibble <= 9) {
+    return nibble + '0';
+  } else if (nibble >= 10 && nibble <= 15) {
+    return nibble - 10 + 'A';
+  } else {
+    return 'X';
+  }
+}
+
+// Represent the buffer `src` (long `len` bytes) as an hex string.  Put
+// the result into `dst` (which must be at least `len * 2` bytes long).
+void BytesToHexDump(const void *src, size_t len, char *dst) {
+  for (size_t i = 0; i < len; i++) {
+    char byte = static_cast<const char *>(src)[i];
+    dst[i * 2] = NibbleToHD((byte >> 4) & 0xF);
+    dst[i * 2 + 1] = NibbleToHD(byte & 0xF);
+  }
+}
+
 }  // namespace
 
 const std::string Pcapng::kName = "pcapng";
 
 Pcapng::Pcapng()
-    : bess::GateHook(Pcapng::kName, Pcapng::kPriority), fifo_fd_(-1) {}
+    : bess::GateHook(Pcapng::kName, Pcapng::kPriority),
+      fifo_fd_(-1),
+      attrs_(),
+      attr_template_() {}
 
 Pcapng::~Pcapng() {
   if (fifo_fd_ >= 0) {
@@ -71,9 +98,40 @@ Pcapng::~Pcapng() {
   }
 }
 
-CommandResponse Pcapng::Init(const bess::Gate *,
+CommandResponse Pcapng::Init(const bess::Gate *gate,
                              const bess::pb::PcapngArg &arg) {
+  Module *m = gate->module();
+  std::string tmpl;
   int ret;
+
+  size_t i = 0;
+  for (const auto &it : m->all_attrs()) {
+    tmpl += it.name + " = ";
+
+    size_t tmpl_offset = tmpl.size();
+
+    tmpl += std::string(it.size * 2, 'X') + " ";
+
+    if (tmpl.size() > std::numeric_limits<uint16_t>::max()) {
+      // Doesn't fit in the option string.
+      break;
+    }
+
+    attrs_.emplace_back(Attr{.md_offset = m->attr_offset(i),
+                             .size = it.size,
+                             .tmpl_offset = tmpl_offset});
+    i++;
+  }
+
+  if (!tmpl.empty() && tmpl.back() == ' ') {
+    tmpl.pop_back();
+  }
+
+  if (tmpl.size() > std::numeric_limits<uint16_t>::max()) {
+    tmpl.resize(std::numeric_limits<uint16_t>::max());
+  }
+
+  attr_template_ = std::vector<char>(tmpl.begin(), tmpl.end());
 
   fifo_fd_ = open(arg.fifo().c_str(), O_WRONLY | O_NONBLOCK);
   if (fifo_fd_ < 0) {
@@ -129,14 +187,35 @@ void Pcapng::ProcessBatch(const bess::PacketBatch *batch) {
   gettimeofday(&tv, nullptr);
 
   uint64_t ts = tv.tv_sec * 1000000 + tv.tv_usec;
+  uint16_t comment_size = static_cast<uint16_t>(attr_template_.size());
 
   for (int i = 0; i < batch->cnt(); i++) {
     bess::Packet *pkt = batch->pkts()[i];
 
+    Option opt_comment = {
+        .code = Option::kComment, .len = comment_size,
+    };
+
+    for (const Attr &attr : attrs_) {
+      const char *attr_data = ptr_attr_with_offset<char>(attr.md_offset, pkt);
+      if (attr_data != nullptr) {
+        BytesToHexDump(attr_data, attr.size, &attr_template_[attr.tmpl_offset]);
+      } else {
+        auto string_it = attr_template_.begin() + attr.tmpl_offset;
+        std::fill(string_it, string_it + attr.size * 2, 'X');
+      }
+    }
+
+    Option opt_end = {
+        .code = Option::kEndOfOpts, .len = 0,
+    };
+
     EnhancedPacketBlock epb = {
         .type = EnhancedPacketBlock::kType,
-        .tot_len = static_cast<uint32_t>(sizeof(epb) + sizeof(uint32_t) +
-                                         RoundUp(pkt->head_len(), 4)),
+        .tot_len = static_cast<uint32_t>(
+            sizeof(epb) + sizeof(uint32_t) +
+            RoundUp<uint32_t>(pkt->head_len(), 4) + sizeof(opt_comment) +
+            RoundUp<uint32_t>(comment_size, 4) + sizeof(opt_end)),
         .interface_id = 0,
         .timestamp_high = static_cast<uint32_t>(ts >> 32),
         .timestamp_low = static_cast<uint32_t>(ts),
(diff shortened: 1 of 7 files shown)


ddiproietto:
The python code is very far from perfect, I appreciate any comments on how to improve it

core/utils/pcapng.h
```diff
@@ -0,0 +1,100 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef BESS_UTILS_PCAPNG_H_
+#define BESS_UTILS_PCAPNG_H_
+
+namespace bess {
+namespace utils {
+namespace pcapng {
+
+struct SectionHeaderBlock {
```
barath: Could you add a comment for these types and expand the field descriptions (or point to some place where these are coming from)?

barath:
Thanks for doing this -- looks very useful.

core/utils/pcapng.h
ddiproietto: Good idea, done, thanks

bessctl/pcapngdump.py
```diff
@@ -0,0 +1,142 @@
...
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import os
+import binascii
+import struct
+
+# Read from file `f` exactly `size` bytes.  Might return a buffer
```
melvinw: style nit: could you make this and other function comments into docstrings (see https://www.python.org/dev/peps/pep-0257/)?

bessctl/pcapngdump.py
```diff
@@ -0,0 +1,142 @@
...
+    shb = read_exactly(pcapng_file, 16)
+
+    if len(shb) < 16:
+        outfile.write("Unexpected EOF while reading SHB\n")
+        return
+
+    struct_endianness = '<'
+    block_type, bom = struct.unpack(struct_endianness + 'IxxxxIxxxx', shb)
+
+    if block_type != 0x0A0D0D0A:
```
melvinw: Could you define this and other the other magic numbers below as constants?

bessctl/pcapngdump.py
```diff
@@ -0,0 +1,142 @@
...
+# smaller than `size` if EOF or error.
+def read_exactly(f, size):
+    ret = b''
+    while True:
+        ret += f.read(size)
+        size = size - len(ret)
+        if size == 0:
+            break
+    return ret
+
```
melvinw: could you run this through `autopep8`?

bessctl/pcapngdump.py
ddiproietto: Makes sense, done thanks!

bessctl/pcapngdump.py
ddiproietto: Sorry, I forgot about that, it's done now

bessctl/pcapngdump.py
ddiproietto: Done, thanks

bessctl/pcapngdump.py
```diff
@@ -0,0 +1,192 @@
...
+
+def read_exactly(f, size):
+    """
+    Read from file f exactly size bytes.
+    Might return a buffer smaller than size if EOF or error.
+    """
+    ret = b''
+    while True:
+        ret += f.read(size)
```
chris3torek: Efficiency is probably not an issue here but if it is the usual trick is to read into a list of strings/bytestrings and then concatenate the list once at the end (to avoid O(n**2) appends)... it looks like size is generally <= 16 though, right?

bessctl/pcapngdump.py
```diff
@@ -0,0 +1,192 @@
...
+        size = size - len(ret)
+        if size == 0:
+            break
+    return ret
+
+
+def round_up(a, b):
+    """Return a rounded up to the nearest multiple of b"""
+    return ((a + (b - 1)) / b) * b
```
chris3torek: for py3k use `// b` (integer divide)

core/hooks/pcapng.cc
```diff
@@ -0,0 +1,254 @@
...
+    return nibble + '0';
+  } else if (nibble >= 10 && nibble <= 15) {
+    return nibble - 10 + 'A';
+  } else {
+    return 'X';
+  }
+}
+
+// Represent the buffer `src` (long `len` bytes) as an hex string.  Put
+// the result into `dst` (which must be at least `len * 2` bytes long).
```
chris3torek: It seems like there ought to be some library function for this (especially, a fancy version like `hexdump -C` output), but I don't know of any... oh well, this will do fine.

bessctl/pcapngdump.py
ddiproietto: Ah good point. `size` actually depends on how big is a block and some of them are variable length.

I implemented your suggestion.

Also there was a bug in my implementation: on every iteration the code was subtracting the total length of ret, whiile it should have been subtracting only the length of the buffer returned on this iteration.

In practice read always returns what you ask, because this is not a tcp stream, but I wish python had a function for this (I'm not aware of one)

Thanks!

bessctl/pcapngdump.py
ddiproietto: Thanks for mentioning py3k.  I fixed this and another py3k problem (a bytes vs string problem)

sangjinhan:
I wonder if `tshark` can be used to print out packets and their metadata for the client side. I remember once @bknuggetsfan worked on metadata-included packet dump, but I am not sure if `tshark` was usable out of the box...

bessctl/pcapngdump.py
```diff
@@ -67,18 +67,19 @@ def read_exactly(f, size):
...
-    ret = b''
+    ret = []
     while True:
-        ret += f.read(size)
-        size = size - len(ret)
+        buf = f.read(size)
+        ret.append(buf)
+        size = size - len(buf)
         if size == 0:
             break
```
chris3torek: Hm, reading this again, what happens at hard EOF? `f.read` returns 0, so `size` does not decrement.  Needs an `if buf == b'': break` to handle that. Then you can change the loop to be `while size:` and get rid of the second test/break.

ddiproietto: You're right, I wasn't handling it correctly.  There was another big problem that prevented EOF from happening: the fifo was set in non blocking mode.  For low packet rates this meant that `f.read` raised EAGAIN immediately, ending the capture before it could even start.

Both these problems should now be fixed, thanks!

ddiproietto:
I don't like having that much python code either, that's why I went with pcapng originally, I was hoping that some command line tool would be able to parse the packet and print the metadata.

tshark seems to be able to print the metadata with
`tshark -r file.pcapng -T fields -e frame.comment`
but it doesn't print anything else from the packet here, unless it's listed explicitly in the command line.

If there was a way to use tshark to display the usual textual representation of each packet and the comment, we should definitely use it

sangjinhan:
I found that there are two ways you can add `frame.comment` to the usual output. Unfortunately neither is perfect, but seems somewhat satisfactory for certain scenarios.
1. `tshark -r file.pcapng -z proto,colinfo,frame.comment,frame.comment`
  * The `-z proto,colinfo` option adds custom fields to the end of each summary line (in ther form of `frame.comment == "..."`)
     * Actually I have tested it with `frame.number`, not `frame.comment`... but it should work as well..?
  * problems: only works with `-T text`. Also not compatible with `-V` or `-x`.
1. `tshark -r file.pcapng -T fields -e _ws.col.Info -e frame.comment`
  * _ws.col.Info looks like the typical one-line summary text.
  * problems: some columns that are included with `-T text` are missing, such as frame number and timestamp. I guess we need to add such fields explicitly.

I think it is worth trying to utilize tshark, given that it also enables specifying capture filters. It is too bad that tshark is not as quite flexible as we want it to be...

ddiproietto:
Thanks for your input!

The `-z proto,colinfo` is exactly what I was looking for.  It it unfortunate that it might not play nicely with other options, so I'm including it only if there are no custom options.  If there are custom options, I think it's the user's job to display what's needed.

I rebased it (since most of the fixes were useless) and pushed a new version

core/hooks/pcapng.h
```diff
@@ -0,0 +1,72 @@
...
+#include "../message.h"
+#include "../module.h"
+
+// Pcapng dumps copies of the packets seen by a gate (data + metadata) in
+// pcapng format.  Useful for debugging.
+class Pcapng final : public bess::GateHook {
+ public:
+  Pcapng();
+
+  ~Pcapng();
```
sangjinhan: Shouldn't it be virtual? The base class GateHook has a virtual destructor.

protobuf/bess_msg.proto
```diff
@@ -473,6 +473,20 @@ message TcpdumpArg {
...
 }
 
+/// Enable/Disable pcapng tapping at an input/output gate.
+///
+/// Once the tap is installed, all packets going through the gate will be
+/// captured and sent in pcapng format to the specified named pipe (FIFO).
+/// Unlike the Tcpdump hook, this also dumps a textual metadata representation,
+/// in the form of a comment to the Enhanced Packet Block. Thus you can run
+/// `tcpdump -r <path to FIFO>` or save the stream in a file.
+/// This feature may afffect performance.
```
sangjinhan: Typo :)

core/hooks/pcapng.h
ddiproietto: Thanks, it's better to mark it explicitly.  Fixed here and on `hooks/tcpdump.h`

protobuf/bess_msg.proto
ddiproietto: Oops, fixed here and ond `TcpdumpArg`

ddiproietto:
As you pointed out, `-z proto,colinfo...` collides with `-V`. If I run tshark with both of them I get an error:

```
localhost:10514 $ tshark wm in 0 -V -z proto,colinfo,frame.comment,frame.comment
  Running: tshark -r /tmp/tmpJrWNoB -V -z proto,colinfo,frame.comment,frame.comment
tshark: the proto,colinfo tap doesn't work if the INFO column isn't being printed.
localhost:10514 $ 
```

You're right, some options are very simple and compatible (I often want to use `-c`). Perhaps we can whitelist them?

Or perhaps we can by default to `-V`, which prints the comment on its own?

I'm fine with any approach :)

sangjinhan:
Thanks for checking. I think then the option should be explicitly given by the user (copypasta should be easy enough since the default option is shown with help) if any other options are given; inspecting the argument list would be not quite reliable...

sangjinhan:
👍 
tumblr_m06t95a6ul1qzuix4o1_1280

PR #643, opened 2017-09-12
ip_checksum: Handle VLAN packets.

ddiproietto:
trafficgen udp mode now supports generating packet with a vlan tag.
This commit changes the ip_checksum module to correctly handle those
packets.

As a result of this change, `ip_checksum` is now slower for untagged
packets.  In this very simple benchmark we measure the amount of traffic
generated by trafficgen udp mode with untagged packets before and after the
change on a single core:

BEFORE: 19.919 Mpps
AFTER: 19.645 Mpps

### DIFF
--- a/core/modules/ip_checksum.cc
+++ b/core/modules/ip_checksum.cc
@@ -37,12 +37,38 @@
 void IPChecksum::ProcessBatch(bess::PacketBatch *batch) {
   using bess::utils::Ethernet;
   using bess::utils::Ipv4;
+  using bess::utils::Vlan;
+  using bess::utils::be16_t;
 
   int cnt = batch->cnt();
 
   for (int i = 0; i < cnt; i++) {
     Ethernet *eth = batch->pkts()[i]->head_data<Ethernet *>();
-    Ipv4 *ip = reinterpret_cast<Ipv4 *>(eth + 1);
+    void *data = eth + 1;
+    Ipv4 *ip;
+
+    be16_t ether_type = eth->ether_type;
+
+    if (ether_type == be16_t(Ethernet::Type::kQinQ)) {
+      Vlan *qinq = reinterpret_cast<Vlan *>(data);
+      data = qinq + 1;
+      ether_type = qinq->ether_type;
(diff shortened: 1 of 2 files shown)


melvinw:
Idea for a future change: I bet this sort of logic will crop up in a lot of modules, so it would be nice if `bess::utils::Ethernet` exposed a function that dealt with all of this and returned a pointer to the L3 header.

core/modules/ip_checksum.cc
```diff
@@ -37,12 +37,28 @@
...
 
   for (int i = 0; i < cnt; i++) {
     Ethernet *eth = batch->pkts()[i]->head_data<Ethernet *>();
-    Ipv4 *ip = reinterpret_cast<Ipv4 *>(eth + 1);
+    Ipv4 *ip;
+
+    // Calculate checksum only for IPv4 packets
+    if (eth->ether_type == be16_t(Ethernet::Type::kIpv4)) {
+      ip = reinterpret_cast<Ipv4 *>(eth + 1);
+    } else if (eth->ether_type == be16_t(Ethernet::Type::kVlan)) {
```
melvinw: Could you check for double tagged packets, too?

ddiproietto: Ok, done, thanks!

shinae-woo:
Looks good to me as well.

If performance matters, you may also want to consider switch-case with a conditional break. I am not quite sure whether some difference will be made by the compiler, but like personal curious..

```
switch(eth->ether_type) {
case Ethernet::Type::kQinQ:
        // recalcuate data
	if (next_ether_type is not kVlan)
		break;
case Ethernet::Type::kVlan:	
        // recalcuate data
	if (next_ether_type is not kVlan)
		break;
case Ethernet::Type::Ipv4:
       ip = reinterpret_cast<Ipv4 *>(data);
       ip->checksum = CalculateIpv4Checksum(*ip);
default:
	break;
}
```

ddiproietto:
I like the switch code!

I tried it, but it appears to be slightly slower (19.37), not sure why

Thanks

shinae-woo:
Thank you by Hodoo ><

PR #644, opened 2017-09-19
Add a new primitive "shared object" to bess core

sangjinhan:
Often multiple modules need to share the same "object". One straightforward way to do this would be (and its issues):

1. Module A creates/owns an object (so loading/unloading order matters), 
1. other modules find the module (module lookup by name is not supposed to be used by modules), and
1. the object is referenced by the modules somehow, e.g., via an A's method (the modules now need to know A's class definition)

To work around these issues, this PR adds a simple primitive to share arbitrary objects among modules (or ports or whatever). It's designed to be minimal with a single API:

```
shared_ptr<TypeFoo> foo_obj = bess::shared_objects::Get<TypeFoo>("foo_name");
```

`Get()` takes an arbitrary string as the name of an object, then:
* If the object with the specified name already exists, it returns a `shared_ptr` to the object.
* If not, it creates an instance of TypeFoo (of course with its constructor, if any).
  * Use `Lookup()` helper function if you don't want this behavior.
* Just like any other `shared_ptr` objects, the returned object is destroyed when none has a reference.
* Each type (`TypeFoo` above) has its own namespace, i.e., another type `TypeBar` may have a different object also named `foo_name`. This is to avoid potential type errors.
* `Get()` and `Lookup()` themselves are thread-safe.
  * You are still responsible for the thread safety of objects, though.

### DIFF
--- a/core/shared_obj.h
+++ b/core/shared_obj.h
@@ -71,18 +71,21 @@ class SharedObjectSpace {
  public:
   // By default, the Get() method creates a new object with its default
   // constructor (a constructor without any arguments). To change this behavior,
-  // you can use the optional argument "constructor", a callable object
+  // you can use the optional argument "creator", a callable object
   // (function pointer, functor, lambda function, etc.) that creates the object
   // in a way you'd like, e.g., to create an object with a non-default
   // constructor, or to reuse an already existing object.
+  //
+  // NOTE: the creator function will be called with the global mutex being held,
+  //       so it's not a good idea for the creator to block on something.
   template <typename T>
   std::shared_ptr<T> Get(
       const std::string &name,
-      std::function<std::shared_ptr<T>()> constructor = DefaultConstructor<T>) {
+      std::function<std::shared_ptr<T>()> creator = DefaultConstructor<T>) {
     SharedObjectKey key = std::make_pair(std::type_index(typeid(T)), name);
 
-    static std::mutex mutex;
-    std::lock_guard<std::mutex> lock(mutex);
+    static std::recursive_mutex mutex;
+    std::lock_guard<decltype(mutex)> lock(mutex);
 
     auto it = obj_map_.find(key);
     if (it != obj_map_.end()) {
@@ -93,9 +96,9 @@ class SharedObjectSpace {
       }
     }
 
-    // "constructor" returns an empty shared_ptr if object allocation
+    // "creator" returns an empty shared_ptr if object allocation
     // failed or no object should be newly made.
-    std::shared_ptr<T> new_object = constructor();
+    std::shared_ptr<T> new_object = creator();
     if (new_object) {
       obj_map_[key] = std::weak_ptr<T>(new_object);
     }
@@ -115,13 +118,14 @@ class SharedObjectSpace {
     return std::shared_ptr<T>(new T());
   }
 
+ private:
+  using SharedObjectKey = std::pair<std::type_index, std::string>;
+
+  // Used by Lookup()
   template <typename T>
   static std::shared_ptr<T> LookupOnly() {
     return std::shared_ptr<T>();
   }
-
- private:
-  using SharedObjectKey = std::pair<std::type_index, std::string>;
 
   std::unordered_map<SharedObjectKey, std::weak_ptr<void>, PairHasher> obj_map_;
 };
--- a/core/shared_obj.h
+++ b/core/shared_obj.h
@@ -1,0 +1,133 @@
+// Copyright (c) 2017, Nefeli Networks, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// * Neither the names of the copyright holders nor the names of their
+// contributors may be used to endorse or promote products derived from this
+// software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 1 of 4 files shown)


core/shared_obj.h
```diff
@@ -0,0 +1,83 @@
...
+namespace bess {
+
+using SharedObjectKey = std::pair<std::type_index, std::string>;
+
+struct PairHasher {
+  template <typename T1, typename T2>
+  std::size_t operator()(const std::pair<T1, T2> &p) const noexcept {
+    return std::hash<T1>{}(p.first) + std::hash<T2>{}(p.second);
+  }
+};
```
barath: Might it be worth putting `PairHasher` in a header in `utils/`?

core/shared_obj.h
```diff
@@ -0,0 +1,83 @@
...
+
+extern std::unordered_map<SharedObjectKey, std::weak_ptr<void>, PairHasher>
+    shared_objects;
+
+template <typename T>
+std::shared_ptr<T> GetSharedObject(const std::string &name,
```
barath: Could you add a function comment?

core/shared_obj.h
```diff
@@ -0,0 +1,83 @@
...
+                                   bool lookup_only = false) {
```
barath: Might it be useful to allow the caller to pass an already-constructed `T*` or arguments for construction of a `T`?  (Since `T` may not have a zero-arg constructor.)

sangjinhan: Ah good point. Probably I would add a `std::function` argument to be used for object construction, rather than directly give it constructor arguments. Preparing arguments or a constructed instance a priori might be expensive/inapplicable in some cases...

core/shared_obj.h
sangjinhan: 👍

core/shared_obj.h
sangjinhan: I thought the hash combining used here is not reliable enough (with the simple `+` operation) to be placed in `utils/`. I will find a better approach and place it in utils.

barath: Sounds good.

core/shared_obj.h
barath: Sounds good.

core/shared_obj.h
```diff
@@ -0,0 +1,83 @@
...
+  SharedObjectKey key = std::make_pair(std::type_index(typeid(T)), name);
+
+  static std::mutex mutex;
+  std::lock_guard<std::mutex> lock(mutex);
+  auto it = shared_objects.find(key);
+
+  if (it == shared_objects.end() || it->second.expired()) {
```
chris3torek: I can't quite figure out why the unit tests are sometimes (?!) failing, but according to http://en.cppreference.com/w/cpp/memory/weak_ptr/lock the `lock` itself is atomic, so I think you can just use `.lock` here and test for failure, rather than an explicit `.expired` test...?

sangjinhan: The failure was due to the fact I used `unordered_map::insert()` as if it was `insert_or_assign()`. `insert()` doesn't overwrite an existing item, and some random reordering of gtest unittests triggered bug.

I think `lock()` doesn't need to be atomic here, since everything is protected by a giant mutex. Anyway your suggestion should make a much cleaner control flow. Updating the code...

core/shared_obj.h
```diff
@@ -0,0 +1,133 @@
...
+// by default; you should use any synchronization mechanism for objects as
+// necessary.
+class SharedObjectSpace {
+ public:
+  // By default, the Get() method creates a new object with its default
+  // constructor (a constructor without any arguments). To change this behavior,
+  // you can use the optional argument "constructor", a callable object
+  // (function pointer, functor, lambda function, etc.) that creates the object
+  // in a way you'd like, e.g., to create an object with a non-default
+  // constructor, or to reuse an already existing object.
```
chris3torek: Nice.  Should perhaps warn users that this holds the lock during the creation, though, so their creator should not block on anything that might also require a shared creation.

core/shared_obj.h
```diff
@@ -0,0 +1,133 @@
...
+    return Get<T>(name, LookupOnly<T>);
+  }
+
+  template <typename T>
+  static std::shared_ptr<T> DefaultConstructor() {
+    return std::shared_ptr<T>(new T());
+  }
+
+  template <typename T>
+  static std::shared_ptr<T> LookupOnly() {
```
barath: I'm not sure I understand this function.  If it's only for internal use, maybe it should be private?

sangjinhan: Oh you are right, it's meant to be used by `Lookup()` only. I'll make it private.

core/shared_obj.h
sangjinhan: Good point. I'll add a comment here.

core/shared_obj.h
chris3torek: Yeah, it's just the magic empty creator that `Lookup` uses (`Get` will always call a creator so we just use this empty one for that case).

core/shared_obj.h
sangjinhan: * Renamed "constructor" to your term "creator" to avoid confusion with object constructor.
* Realized the creator might want to create another shared object. Replaced `mutex` with `recursive_mutex`.

core/shared_obj.h
```diff
@@ -71,18 +71,21 @@ class SharedObjectSpace {
...
   template <typename T>
   std::shared_ptr<T> Get(
       const std::string &name,
-      std::function<std::shared_ptr<T>()> constructor = DefaultConstructor<T>) {
+      std::function<std::shared_ptr<T>()> creator = DefaultConstructor<T>) {
     SharedObjectKey key = std::make_pair(std::type_index(typeid(T)), name);
 
-    static std::mutex mutex;
-    std::lock_guard<std::mutex> lock(mutex);
+    static std::recursive_mutex mutex;
```
chris3torek: Oh, I didn't notice that you were using a non-recursive mutex earlier!  This is better, one can make a shared object using another shared object in the same thread now. (Not sure we'll ever need this but it's generally inexpensive anyway, and I often think all mutexes should default to recursive.)

shinae-woo:
TGIF 

image

PR #645, opened 2017-09-21
fix FlowGen documentation typo

chris3torek:
The actual core/modules/flowgen.cc code reads:

  if (ip_src_range_ == 0 && ip_dst_range_ == 0 && port_src_range_ == 0 &&
      port_dst_range_ == 0) {
    /*randomize ports anyway*/
    port_dst_range_ = 20000;
    port_src_range_ = 20000;
  }

but the comment in module_msg.proto said "2000".  On the assumption
that the code is more correct than the documentation, fix the
documentation.


sangjinhan:
Merged, thanks!
tumblr_m6ry1fqzcj1qzv52ko1_1280

PR #647, opened 2017-09-22
Automatically pause workers on certain commands

nrdmao:
The following changes are being proposed to automatically pause workers when commands requiring that workers be paused are executed.

This change alleviates the application from having to pause/resume workers and the associated maintenance of the worker state. This issue more particularly effects asynchronous clients.

This change is also expected to improve the performance of bessd when configuration changes are made under traffic load as workers are paused only when the message is processed. As is currently implemented, workers are paused from the processing of PauseAll to the subsequent ResumeAll, which could be a significant period of time given the gRPC round trips between controller and bessd.

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -469,10 +469,7 @@ class BESSControlImpl final : public BESSControl::Service {
   Status ResetAll(ServerContext* context, const EmptyRequest* request,
                   EmptyResponse* response) override {
     Status status;
-
-    if (is_any_worker_running()) {
-      return return_with_error(response, EBUSY, "There is a running worker");
-    }
+    WorkerPauser wp;
 
     LOG(INFO) << "*** ResetAll requested ***";
 
@@ -538,10 +535,8 @@ class BESSControlImpl final : public BESSControl::Service {
   }
 
   Status ResetWorkers(ServerContext*, const EmptyRequest*,
-                      EmptyResponse* response) override {
-    if (is_any_worker_running()) {
-      return return_with_error(response, EBUSY, "There is a running worker");
-    }
+                      EmptyResponse*) override {
+    WorkerPauser wp;
     destroy_all_workers();
     LOG(INFO) << "*** All workers have been destroyed ***";
     return Status::OK;
@@ -615,9 +610,7 @@ class BESSControlImpl final : public BESSControl::Service {
 
   Status ResetTcs(ServerContext*, const EmptyRequest*,
                   EmptyResponse* response) override {
-    if (is_any_worker_running()) {
-      return return_with_error(response, EBUSY, "There is a running worker");
-    }
+    WorkerPauser wp;
 
     if (!TrafficClassBuilder::ClearAll()) {
       return return_with_error(response, EBUSY, "TCs still have tasks");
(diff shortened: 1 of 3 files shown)


core/bessctl.cc
```diff
@@ -87,6 +87,28 @@ static inline Status return_with_errno(T* response, int code) {
...
+      workers_paused = false;
+    }
+  }
+  ~WorkerPauser() {
+    if (workers_paused) {
+      resume_all_workers();
+    }
+  }
+ private:
+  bool workers_paused;
```
chris3torek: Based on the Google C++ style guide, this should be named `workers_paused_`

core/bessctl.cc
```diff
@@ -87,6 +87,28 @@ static inline Status return_with_errno(T* response, int code) {
...
 
+
+// This class is used as a resource manager to automatically pause workers if
+// running and then restarts workers if they were previously paused.
+class WorkerPauser {
+ public:
+  explicit WorkerPauser() {
+    if (is_any_worker_running()) {
+      workers_paused = true;
+      pause_all_workers();
```
ddiproietto: The other place in this file that calls `pause_all_workers()` also prints a log message.

I think the log message is useful, can you make sure it gets printed even in this case, perhaps by moving the log message inside pause_all_workers()?

core/bessctl.cc
```diff
@@ -87,6 +87,28 @@ static inline Status return_with_errno(T* response, int code) {
...
+    } else {
+      workers_paused = false;
+    }
+  }
+  ~WorkerPauser() {
+    if (workers_paused) {
+      resume_all_workers();
```
ddiproietto: Same here, a log message would be nice.

core/bessctl.cc
```diff
@@ -1244,18 +1248,7 @@ class BESSControlImpl final : public BESSControl::Service {
...
-      if (m1->num_active_workers()) {
-        return return_with_error(response, EBUSY, "Module '%s' is in use",
-                                 m1_name);
-      }
-      if (m2->num_active_workers()) {
-        return return_with_error(response, EBUSY, "Module '%s' is in use",
-                                 m2_name);
-      }
-    }
-
```
ddiproietto: This had a special case: a couple of modules that weren't used by any workers could be connected without any pausing.  Do you think it's worth preserving?

core/bessctl.cc
nrdmao: will fix.

core/bessctl.cc
nrdmao: I wonder if we should use a conditional log, like VLOG(1) or something. I am concerned about cluttering the log file with this information. I am also expecting that this auto-pause could happen frequently and the LOG(INFO) call might prove to be expensive.

core/bessctl.cc
nrdmao: Yes, let me look into that a bit more. Indeed one could imagine an application building up a module topology while the workers are running and then making a final connection with minimal worker disruption.

nrdmao:
Regarding: _before this change, a controller could try to apply some changes without pausing (for example, connect two new modules) and receive an error if this wasn't possible._ from @ddiproietto 

I am open to suggestion. I was thinking that if this behavior was required we could add a flag to all the affected gRPC messages to indicate if workers should be auto-paused or if an error should be returned indicating that workers are running.

core/bessctl.cc
ddiproietto: You're right, it could get pretty verbose. VLOG(1) sounds good to me.

ddiproietto:
You're right, we can always add a flag if we need to support that kind of behavior in the future.

sangjinhan:
A corner case might arise when some workers are already paused at the beginning, while other workers are active (one can selectively pause workers with `pause_worker()`, instead of using `pause_all()`). Then in the destructor, only newly paused workers should be woken up.

nrdmao:
@sangjinhan yes, I saw the same issue. I will push another change.

shinae-woo:
Thank you!

image

PR #649, opened 2017-09-23
Add missing dependency files in Makefile

sangjinhan:
In the line in Makefile `-include $(wildcard $(DEPDIR)/**/*.d)`, `**` is a shell feature and not supported by Makefile. As a result, this clause won't include `$(DEPDIR)/*.d` files, missing dependencies for the files in `core/` (not in its sub directories). Incremental building behaves incorrectly.


chris3torek:
image

PR #650, opened 2017-09-23
Do not consume Ethernet headers in loopback mode

sangjinhan:
In the kernel module, all received packets are passed to `eth_type_trans()`. Its side effect is that packet Ethernet headers are decapsulated; not a desired behavior for loopback mode, where one would expect all sent packets return intact.

### DIFF
--- a/core/kmod/sn_netdev.c
+++ b/core/kmod/sn_netdev.c
@@ -258,12 +258,9 @@ static void sn_disable_interrupt(struct sn_queue *rx_queue)
 	rx_queue->rx.rx_regs->irq_disabled = 1;
 }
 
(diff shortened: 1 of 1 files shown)


sangjinhan:
This bug caused incorrect behavior of the `port/latency` example (`Measure` module do not find a marker in packets updated by `Timestamp`, due to offset shift by 14 bytes).

shinae-woo:
LGTM.

One quick question: When is `sn_process_rx_metadata()` supposed to return rather than zero?

sangjinhan:
I think it used to return error values regarding some offloading features. Thank you for making it `void`.

melvinw:
Thanks! cat

Issue #352, opened 2017-03-24
Add keyword argument support to `bessctl`


apanda:
This is based on looking at #351 -- currently `bessctl` seems to only support positional arguments, which is a problem when multiple arguments to the same command are optional (in particular requiring one to decide between what is more likely to be omitted and ordering arguments to allow that). We should add support for some sort of keyword arguments, this would fix this problem. The challenge of course is making sure that the commands don't become too verbose as a result.

PR #651, opened 2017-09-26
bessctl: Add GRAPHEASY_OPTS to show/monitor pipeline.

ddiproietto:
Similar to TCPDUMP_OPTS, this lets the user provide additional options
to the graph-easy invocation (or even pipe the graph-easy output to other
programs).

I find this change useful because sometimes the graph is too complicated to
be shown on the terminal.  With this commit

```
localhost:10514 $ show pipeline --as dot | dot -Tsvg -o graph.svg
```

outputs the graph to a file on disk.


ddiproietto:
If this feels to hacky, we could change the commands to explicitly supports output to an image, but I think keyword arguments would really help in that case (#352)

shinae-woo:
image

PR #653, opened 2017-09-28
Make module aligned to 64B

shinae-woo:
* plus, fix a bug `operator new()` in Packet class 
* `alinged_alloc()` is available from C++17, so I use legacy `mem_alloc_ex()`.


melvinw:
Thanks! cat

PR #654, opened 2017-09-28
Fill leaf info in `TrafficClass` protobuf From `ListTcs` gRPC

ddiproietto:
It makes sense to fill this info, it helps correlating the leaf with a module.

Reported by @nlsun


melvinw:
Thanks! cat

PR #655, opened 2017-09-28
Prevent zombie processes of tcpdump, tshark

shinae-woo:
* tcpdump / tshark becomes zombie processes when bess daemon responses with error codes
* handling bess error codes for processing cleanups

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -844,7 +844,7 @@ def _do_run_file(cli, conf_file):
 
         # Mimic python's error reporting style
         cli.err('\n  File "%s", line %d\n    %s\n    %s\nSyntaxError: %s' %
-                (conf_file, e.lineno, e.text, ' '*(e.offset - 1) + '^', e.msg))
+                (conf_file, e.lineno, e.text, ' ' * (e.offset - 1) + '^', e.msg))
         raise cli.HandledError()
     except Exception as e:
         cli.err('Fail to compile bess config file (%s): %s ' % (conf_file, e))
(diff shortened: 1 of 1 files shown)


chris3torek:
I think we could get the close/unlink/system sequence into one place, fairly cleanly ... move the first "try/finally" pair inside the second one, so that the second "finally" always runs, and use a variable to know whether to call `hook_fn(False, module_name, direction, gate)`, maybe.  (I haven't looked closely at the change yet)

shinae-woo:
@chris3torek @ddiproietto Yeah, I have thought about putting them in one place and I agree with the pros part of it! Meanwhile, having the same behavior (i.e., raise `cli.bess.Exception` back to the caller to make the error handling consistent with others cli methods) was not straightforward to me and the control flow become complex. I am not python expert. Any help?

chris3torek:
I have not tested this _at all_ but I pushed an attempt to unify all the cleanup handling under one code path so that it's at least readable (if a bit verbose), to [my fork here](https://github.com/chris3torek/bess/tree/tcpdump-intr).  The strategy is to use a bunch of tiny little cleanup functions, and make a list of them to call in a `finally` clause.  Then we can just write mostly-normal Python code and know that, no matter what else happens, every cleanup will happen in the order they occur in the list.  Each one uses a variable to decide whether to do anything and what to do.

I made use of the fact that `cleanup_resume` happens towards the end to do lazy pause/resume around the hooks, too.

shinae-woo:
Thanks! @sangjinhan also pointed out #647, we don't need to pause/resume as well.
I will push another commit!

bessctl/commands.py
```diff
@@ -1856,46 +1856,33 @@ def _capture_module(cli, module_name, direction, gate, opts, program, hook_fn):
...
-    try:
         proc.wait()
     except KeyboardInterrupt:
         # kill all descendants in the process group
         os.killpg(proc.pid, signal.SIGTERM)
+    except cli.bess.Error:
+        unhook = False
+        # kill all descendants in the process group
+        os.killpg(proc.pid, signal.SIGTERM)
+        raise
```
chris3torek: I'm not sure if this is a good or bad idea, but to collect zombies we should call `proc.wait()` after any `os.killpg(proc.pid, ...)`. The reason it might be a bad idea is that if the process refuses to die, `proc.wait` could wait forever.

shinae-woo: Right. umm.. At least, users can catch what is happening.

sangjinhan:
Merged, thanks!
tumblr_m6nh71i1yx1qzci09o1_1280

PR #661, opened 2017-10-03
add resume hooks

melvinw:
It would be nice if module writers had a way to perform tasks at resume-time without having to modify the `ResumeAll()` and `ResumeWorker()` RPCs. This set of commits introduces `ResumeHooks` to provide that functionality.


melvinw:
Not entirely sure if the interface I came up with this is the cleanest-- definitely willing to rework it if anyone has an opinion on the matter. The `ResumeHookUser` stuff is there mostly to avoid having to scan the entire map of modules on resume, as it can grow to be quite large.

core/resume_hook.cc
```diff
@@ -0,0 +1,141 @@
...
+// Worker resume hooks -------------------------------------------------------
+template <>
+std::map<std::string, ResumeHookFactory<Worker>>
+    &ResumeHookFactory<Worker>::all_resume_hook_factories_holder(bool reset) {
+  // Maps from hook names to hook factories. Tracks all hooks (via their
+  // ResumeHookFactorys).
+  static std::map<std::string, ResumeHookFactory<Worker>>
+      all_resume_hook_factories;
+
+  if (reset) {
```
chris3torek: I can't figure out where reset is ever true (but this is the same as the gate hook code, where I also can't figure out where this is ever true, so at least I'm consistently missing something 😺 )

core/gate.cc
```diff
@@ -72,7 +72,7 @@ int Gate::AddHook(GateHook *hook) {
   }
 
   hooks_.push_back(hook);
-  std::sort(hooks_.begin(), hooks_.end(), GateHookComp);
+  std::sort(hooks_.begin(), hooks_.end());
```
chris3torek: (see later comment about sort-right-after-insert ... I wound up going through this backwards)

core/bessctl.cc
```diff
@@ -1376,6 +1463,107 @@ class BESSControlImpl final : public BESSControl::Service {
...
+        disable_resume_hook(user, request->hook_name());
+        VLOG(1) << "Disabled global resume hook '" << request->hook_name()
+                << "'";
+      }
+      return Status::OK;
+    }
+
+    uint64_t worker_filter = request->workers();
+    while (worker_filter) {
+      Worker* w = workers[wid];
```
chris3torek: The documentation says that the filter is a bitmap, but we don't actually test the specific bit (`worker_filter & 1`) here...

core/bessctl.cc
```diff
@@ -1241,10 +1327,10 @@ class BESSControlImpl final : public BESSControl::Service {
...
       propagate_active_worker();
-      if (m1->num_active_workers() || m2->num_active_workers() ) {
-	WorkerPauser wp; // Only pause when absolutely required
-	ret = m1->ConnectModules(ogate, m2, igate);
-	goto done;
+      if (m1->num_active_workers() || m2->num_active_workers()) {
+        WorkerPauser wp;  // Only pause when absolutely required
+        ret = m1->ConnectModules(ogate, m2, igate);
+        goto done;
       }
```
chris3torek: (general comment: I like having the whitespace fixes as a separate commit - maybe we should just get that one in and done with first)

core/bessctl.cc
```diff
@@ -531,17 +562,72 @@ class BESSControlImpl final : public BESSControl::Service {
...
                    EmptyResponse*) override {
-    LOG(INFO) << "*** Resuming ***";
     if (!is_any_worker_running()) {
       attach_orphans();
     }
+
+    ResumeHookUser<Worker>* global_hooks =
+        bess::ResumeHookUser<Worker>::get_user_hooks(nullptr);
+
+    // Enable metadata offset computation by default
```
chris3torek: I wish there were a good way to get this to happen at constructor time, when we build the metadata resume hook... not sure if there is though, we might get into constructor ordering issues.

core/resume_hook.cc
melvinw: `reset` is here to facilitate testing. `ModuleBuilder::all_module_builders_holder` has a similar flag that I could have sworn used to be used in some tests, but I can't find them anymore either

core/bessctl.cc
melvinw: Doh. Thanks for catching-- will fix

core/resume_hooks/README.md
```diff
@@ -0,0 +1,17 @@
+# Resume Hooks
+Resume hooks allow you to run arbitrary code immediately before a worker is
+resumed. They can be configured with the `ConfigureResumeHook()` RPC.
+
+The API is simple, similar to that of `GateHook`. All you need to do is:
+
+- Define `YourHook::kName`. This must be unique across all other resume hooks.
+- Define `YourHook::kPriority`. Ties are broken by hook name in increasing lexographical order.
+- Define `void YourHook::Run()`.
+
```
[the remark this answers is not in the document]
melvinw: I may be able to move this to the `ADD_MODULE_RESUME_HOOK`/`ADD_WORKER_RESUME_HOOK` macros. Will update the docs either way. Thanks

core/gate.cc
melvinw: Good point. Switched to `std::set` over `std::vector` here since this isn't on the fast path.

core/bessctl.cc
melvinw: Yeah, I thought about putting this logic in the [Pipeline constructor](https://github.com/NetSys/bess/blob/master/core/metadata.h#L148), but it's not clear how to handle re-adding the hook after a `daemon reset`.

core/bessctl.cc
melvinw: Ah, whichever commit touched these lines last must not have been clang-formated correctly. We have a pre-commit hook for this, but it only really works if everyone installs it :)

core/resume_hooks/README.md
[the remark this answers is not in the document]
melvinw: Fixed along with some other minor things in this doc

core/bessctl.cc
chris3torek: Whoops, the `continue` skips the `worker_filter >>= 1` and `wid++` lines 😮

melvinw: Oh, yikes! Fixed and rebased

melvinw:
Marking this not-ready while I refactor things

sangjinhan:
Did some offline discussion for higher bandwidth. One reminder for later: resume hooks should be called also when pause/resume is done implicitly (by WorkerPauser)

melvinw:
Rebased and updated to reflect our conversation, @sangjinhan . I also added a possible implementation of a thread-safe `PortOut` as an example of how to implement pre-resume hooks for modules.

core/resume_hook.cc
```diff
@@ -0,0 +1,70 @@
...
+
+#include <map>
+#include <set>
+#include <string>
+#include <utility>
+
+namespace bess {
+
+std::set<std::unique_ptr<ResumeHook>> global_resume_hooks;
+std::set<Module *> event_modules;
```
sangjinhan: Perhaps should be in `event.cc`, since Resume is just one particular type of events?

core/bessctl.cc
```diff
@@ -1147,6 +1156,10 @@ class BESSControlImpl final : public BESSControl::Service {
       response->set_name(module->name());
     }
 
+    if (module->OnEvent(bess::Event::PreResume) != ENOTSUP) {
```
sangjinhan: This event is somewhat spurious, since `PreResume` event is not really happening here. I think it should be fine for most module handlers, but how about,
1. By default, we add all new modules to `event_modules` (without `if` here).
1. `ENOTSUP` is checked in the loop over `event_modules`.

core/resume_hook.h
```diff
@@ -0,0 +1,136 @@
...
+
+  virtual ~ResumeHook() {}
+
+  const std::string &name() const { return name_; }
+
+  uint16_t priority() const { return priority_; }
+
+  virtual void Run() = 0;
+
+  bool operator<(const ResumeHook &rhs) {
```
sangjinhan: This method can be `const`

core/worker.cc
```diff
@@ -403,20 +406,59 @@ bool detach_tc(bess::TrafficClass *c) {
...
+WorkerPauser::~WorkerPauser() {
+  if (workers_paused_.size()) {
+    run_global_resume_hooks();
+  }
+
+  std::set<Module *> modules_run;
   for (int wid : workers_paused_) {
+    for (Module *m : bess::event_modules) {
+      if (!modules_run.count(m) && m->active_workers()[wid]) {
+        m->OnEvent(bess::Event::PreResume);
```
sangjinhan: For each module, the event seems to be triggered twice (here and `run_global_resume_hooks` above)

core/worker.cc
```diff
@@ -403,20 +406,59 @@ bool detach_tc(bess::TrafficClass *c) {
   return remove_tc_from_orphan(c);
 }
 
+void run_global_resume_hooks() {
```
sangjinhan: `resume_hook.cc` might be a better place for this function (not sure)

protobuf/bess_msg.proto
```diff
@@ -499,6 +499,17 @@ message ConfigureGateHookRequest {
 }
 
 //  -------------------------------------------------------------------------
+//  Resume hooks
+//  -------------------------------------------------------------------------
+
+// Enable/Disable a global ResumeHook.
+message ConfigureResumeHookRequest {
+  string hook_name = 1;         /// Name of the hook
+  bool enable = 5;              /// Installs the hook if True, else uninstalls
```
sangjinhan: Since the protobuf has not been released, we can do 1, 2, 3, ... for the member index?

pybess/bess.py
```diff
@@ -495,6 +495,15 @@ def _configure_gate_hook(self, hook, module,
         request.arg.Pack(arg)
         return self._request('ConfigureGateHook', request)
 
+    def _configure_resume_hook(self, hook, arg, enable=None):
```
sangjinhan: Since the method is meant to be used externally, the method name should not begin with `_`.

(`enable=True` should be fine...?)

core/worker.cc
```diff
@@ -403,20 +406,59 @@ bool detach_tc(bess::TrafficClass *c) {
...
+        VLOG(1) << "*** Pausing Worker " << wid << " ***";
+        pause_worker(wid);
+      }
     }
   }
 }
 
-WorkerPauser:: ~WorkerPauser() {
+WorkerPauser::~WorkerPauser() {
+  if (workers_paused_.size()) {
```
sangjinhan: nit: `!workers_paused_.empty()`

core/resume_hook.cc
```diff
@@ -0,0 +1,70 @@
...
+
+std::map<std::string, ResumeHookFactory>
+    &ResumeHookFactory::all_resume_hook_factories_holder(bool reset) {
```
sangjinhan: We don't need `reset` here..? (I think this idion was used for gtest somewhere else, but not quite necessary here)

core/module.h
```diff
@@ -287,6 +288,14 @@ class alignas(64) Module {
   virtual struct task_result RunTask(void *arg);
   virtual void ProcessBatch(bess::PacketBatch *batch);
 
+  /*
+   * If a derived Module overrides OnEvent and returns ENOTSUP for a particular
+   * Event `e` it will be invoked for each instance of the derived Module
+   * whenever `e` occours. See `event.h` for details about the various event
+   * types and their semantics/requirements when it comes to modules.
+   */
+  virtual int OnEvent(bess::Event) { return ENOTSUP; }
```
sangjinhan: To be consistent with other code, please consider using `-ENOTSUP`. The remaining C-style code expects int functions return negative numbers to report errors.

core/resume_hooks/README.md
```diff
@@ -0,0 +1,12 @@
+# Resume Hooks
+Resume hooks allow you to run arbitrary code immediately before a worker is
+resumed. They can be configured with the `ConfigureResumeHook()` RPC.
+
+The API is simple, similar to that of `GateHook`. All you need to do is:
+
+- Define `YourHook::kName`. This must be unique across all other resume hooks.
+- Define `YourHook::kPriority`. Ties are broken by hook name in increasing lexographical order.
```
sangjinhan: It may be worth mentioning that lower numbers have higher priority

core/resume_hook.cc
melvinw: Yeah, it's carryover from ModuleBuilder to facilitate unit testing-- will remove.

core/worker.cc
melvinw: Ah, good catch. Will fix

protobuf/bess_msg.proto
melvinw: Definitely

pybess/bess.py
melvinw: Ah, yup-- but I think we still need the `is None` check below since the bessctl command parser sets unpresent args to `None`

core/module.h
melvinw: Sure thing

core/worker.cc
melvinw: Agreed. I'll need to rework this slightly. Moving as is introduces an awkard circular dependency between `resume_hook.h` and `resume_hooks/setup_metadata.h`

core/bessctl.cc
melvinw: Good point. Will change

melvinw:
Comments addressed. Not sure if [this weirdness](https://github.com/NetSys/bess/pull/661/files#diff-9e373311c9fa90b6c7c3412dfc306367R48) is okay, though

core/resume_hook.h
```diff
@@ -0,0 +1,137 @@
...
+  virtual ~ResumeHook() {}
+
+  const std::string &name() const { return name_; }
+
+  uint16_t priority() const { return priority_; }
+
+  virtual void Run() = 0;
+
+  bool operator<(const ResumeHook &rhs) const {
+    return priority_ < rhs.priority_ && name_ < rhs.name_;
```
nrdmao: Don't you want something like?:
return priority_ < rhs.priority_ || (priority_ == rhs.priority_ && name_ < rhs.name_);

nrdmao: same issue in gate.cc

core/worker.cc
```diff
@@ -407,16 +410,31 @@ WorkerPauser::WorkerPauser() {
...
     }
   }
 }
 
-WorkerPauser:: ~WorkerPauser() {
+WorkerPauser::~WorkerPauser() {
+  if (!workers_paused_.empty()) {
+    bess::run_global_resume_hooks(false);
+  }
+
```
nrdmao: I noticed in my testing that there should actually be a call made to attach_orphans() made before workers are resumed as is done in when the ResumeAll command is processed. Otherwise, if there are no subsequent calls made to ResumeAll when no workers are running, orphans will remain unattached. So, you can either add the call to attach_orphans() here or I can provide a fix after this code merges.

core/resume_hook.h
chris3torek: I thought @melvinw had used `std::tie` to do this comparison, but maybe that was a different commit. But yes, when you can `std::tie` the various fields and compare, that's the way to go.

core/worker.cc
melvinw: Great catch. I can fold the change into this PR

core/resume_hook.h
melvinw: Ah, yup. After reading the definition of `operator<` on `std::tuple`, `std::tie` is definitely what we should use here. I switched to the current definition because I thought tuple comparisons were just `lhs[0] < rhs[0] && ... && lhs[n] < rhs[n]` and this was more explicit. Turns out tuples are actually ordered by `(lhs[0] < rhs[0] || !(rhs[0] < lhs[0])) && ...`

melvinw: Fixed

core/worker.cc
melvinw: Fixed

sangjinhan:
(rebased)

sangjinhan:
Merged, thanks! Sorry for the delay.

a-dar-tuesday-32

PR #662, opened 2017-10-04
Support bess daemon generated traffic class name

shinae-woo:
* traffic class names with prefix '!'

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -317,7 +317,7 @@ def get_var_attrs(cli, var_token, partial_word):
                 pass
 
         elif var_token == 'TC...':
-            var_type = 'name+'
+            var_type = 'tcname+'
             var_desc = 'one or more traffic class names'
             try:
                 var_candidates = [getattr(c, 'class').name
@@ -439,7 +439,7 @@ def split_var(cli, var_type, line):
             head = line[:pos]
             tail = line[pos:]
 
-    elif var_type in ['wid+', 'name+', 'map', 'pyobj', 'opts']:
+    elif var_type in ['wid+', 'name+', 'tcname+', 'map', 'pyobj', 'opts']:
         head = line
         tail = ''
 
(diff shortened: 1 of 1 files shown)


ddiproietto:
Thanks for the fix!

I've realized that we also need to handle `:` in the later characters, as the leaf tasks include that in their name.

Do you think we should enhance the exception message to include `!` (and eventually `:`) in the regex displayed to the user?

chris3torek:
Is this exclusive to traffic class names?  If so, should we separate them out into their own `tcname+` token type?

shinae-woo:
@ddiproietto Well, I don't know.  '!' and ':' is not intended for being used by users.  And bess daemon uses the character to prevent conflict between user-provided names.

shinae-woo:
@chris3torek Yes, only for traffic class name. As is, it is only for monitoring, not for creating a new name.

ddiproietto:
@shinae-woo Ok

chris3torek:
OK, so, we could add a `tcname+` and do:

    elif var_type in ('name+', 'tcname+'):
        regexp = 'r'^[_a-zA-Z][\w]*$' if var_type == 'name+' else r'^!?[_a-zA-Z][\w:]*$'
        val = sorted(list(set(head.split())))  # collect unique items
        ... same as before ...

(or similar but change the string passed to `cli.BindError` if you want to show that `!` and `:` are allowed).

chris3torek:
(I'll add that I don't feel strongly about this. I think you can simplify the optional `!` using `!?` rather than `(!)?`, you'd only need the parens here if you wanted to retrieve it as a group later.)

shinae-woo:
Added traffic_class_name as @chris3torek suggested. This PR would be a just quick fix; a better solution would be refactoring the way of creating traffic class name by daemon to prevent conflict without '!'. One simple way might be, incrementally increasing the postfix indexes (`:0, :1, :2`) as of now, and check for duplication.

sangjinhan:
Merged, thanks!
default-4

Issue #625, opened 2017-08-29
test framework could use more predefined functions


chris3torek:
Some module tests work by injecting some packet(s) using a port-and-socket pair generated by `gen_socket_and_port`.  See, e.g., `bessctl/conf/testing/module_tests/nat.py` or `bessctl/conf/testing/module_tests/etherencap.py`.  As long as the code has the form:

    sock.send(data)
    return_data = sock.recv(2048)

or similar, we're fine.  But suppose we want to specifically test, e.g., that a packet with some attributes gets routed to some particular output gate:

    def get_pkts_for_ogate(module, ogate):
        info = bess.get_module_info(module)
        for gate in info.ogates:
            if gate.ogate == ogate:
                return gate.pkts
        return 0

    sock.send(pkt)
    assert get_pkts_for_ogate(module, 0) == 0
    assert get_pkts_for_ogate(module, 1) == 1

Now there is a potential problem: the `sock.send` call in Python waits for the socket data to get to the kernel, but getting into the kernel does not mean that the data have even arrived at the receiving `UnixSocketPort` in bessd (much less been dispatched).

What's needed here is a way to wait for the `UnixSocketPort` module's igate packet count to go up.  This means we need to sample it once before the `sock.send`, and in a loop afterward.  This code should probably be in the test framework so that any test can use it easily.

While at it, the test framework probably should define a function similar to `get_pkts_for_ogate` above, but generalized to handle both input and output gates.  Rather than getting just the packet count it should get the IGate or OGate data from <module, direction, gate_num>.  Then `get_pkts_for_ogate(module, 0)` becomes `get_gate_stats(module, 'ogate', 0).pkts`.  But that's less important than hiding the "sample the count, then inject packet(s), then wait for packet to get dispatched" test case.

(If there's a better way overall to do this, that would be even better)

sangjinhan:
Once we have a wide range of such helper functions, it would be beneficial to switch to the Python unittest framework, rather than using our custom one. At the moment writing a module test is somewhat arcane and exotic.

shinae-woo:
I am taking a look. Letting me know if you have any wishlist added into it.

sangjinhan:
#663 introduced mechanisms to reliably feed packets to modules and dispatch the results. Thank you all 👍

PR #663, opened 2017-10-04
New module testing framework

shinae-woo:
Module testing framework including improvement suggestions in #625.

The major changes are
- Adopting python unittest module
- Checking module scheduling completion by checking traffic class statistics (No more sleep and blocking system call, shorter test completion time) 

I did not heavily fix each module test, but more focusing on porting existing tests into the new framework. It would be great if the tests are gradually improved as modules are updated.

Reporting two issues
1. To get the benefit of syntactic sugar, the tests are still called by `bessctl`. Each module test is called as `exec()`, we can not directly get the return value from unittest.  In the case of test failure, we need to forcibly exit the tests by returning a non-zero value.  This can be further improved by moving sugar semantics (+ more functionality) from `bessctl` to 'pybess' so that native unittest.py can be directly used. Separate PR required.

2. All tests in the same file share the same execution environments (by `exec()`). This may become problematic when we use sugared statements which automatically add global variables. When we use `src::Source()` multiple times for different test functions in a single file,  the second src::Source() call will fail due to conflict module names, since `src` variable is global, not local.


bessctl/test_utils.py
```diff
@@ -0,0 +1,243 @@
...
+            raise AssertionError('"%s" != "%s"',
+                                 pkt_str(pkt1), pkt_str(pkt2))
+
+    def assertNotSamePackets(self, pkt1, pkt2):
+        if pkt_str(pkt1) == pkt_str(pkt2):
+            raise AssertionError('"%s" == "%s"',
+                                 pkt_str(pkt1), pkt_str(pkt2))
+
+    def assertBessAlive(self):
+        self.bess = BESS()
```
sangjinhan: Recreating (and replacing) a `BESS` instance may be somewhat disruptive. We can either:
1. using existing `self.bess`, we check with a dummy RPC (e.g., `GetVersion`), or
2. we create a new BESS instance, but we do not overwrite self.bess with it.

bessctl/conf/testing/run_module_tests.py
```diff
@@ -0,0 +1,85 @@
...
+
+def main():
+    any_failure = 0
+
+    try:
+        run_cmd('%s daemon start' % bessctl)
+    except CommandError:
+        raise Exception('bess daemon could not start')
+
+    for file_name in glob.glob(os.path.join(test_dir, "*.py")):
```
nemethf: Wouldn't this mean that the feature of testing just a single module disappears?  I've found that very useful.

shinae-woo: I propose a new one! Please check a `assertBessAlive()` in `test_util.py` and see sample code in module test code.

sangjinhan: Maybe I misunderstand, but I guess he meant the configurable `TEST_PATH` and `TEST_NAME` environment variable  here, not about testing "a single module" vs "a chain of modules"

shinae-woo: Oh, sorry. it was 'my' misunderstand. @nemethf, you can simply run a single test within `bessctl`. For example, simply do `run testing/xxx.py`, then the test results will show up on the shell.

shinae-woo:
@sangjinhan, Of course, I will do it. I see that way is much better for maintenance.

shinae-woo:
Reminder for myself, in an offline discussion, @chris3torek pointed out that the current `run_module()` is not fully equivalent with existing  `CRASH_TEST_INPUTS`  as now we use random packets, but previously they use predefined (I would say `hard-coded) set of packets. Using random packets are more appropriate for the most case for checking crash rather than correct functionality, but I also quite agree that it would be better if there is a way that users can specify a set of packet template to run. I will extend `run_module()` interface accordingly.

bessctl/test_utils.py
shinae-woo: Good point.

bessctl/test_utils.py
```diff
@@ -0,0 +1,243 @@
...
+    if not pkt:
+        return str('').encode('HEX')
+    else:
+        return str(pkt).encode('HEX')
+
+
+class BessModuleAssertion():
+
+    def assertSamePackets(self, pkt1, pkt2):
+        if pkt_str(pkt1) != pkt_str(pkt2):
```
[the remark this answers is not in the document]
shinae-woo: `assertEquals()` are not available unless inheriting `unittest.Test` as you said. But they show the same behavior (raising same `AssertionError`), and we can also customize assert error messages per error type in this way :)

bessctl/conf/testing/run_module_tests.py
nemethf: Oh, that's great!  (And sorry for the noise.)

sangjinhan:
Merged, thanks!
c817c23ed5406b8cbfdf17e95ed206af

PR #664, opened 2017-10-04
Add clang 4 and 5 to the build container

sangjinhan:
Unlike gcc, the clang versions installed in the build container are quite old. This PR adds clang 4.0 and 5.0 to the compiler collection.

A few notes:
* clang-4.0 and higher are not directly provided by Ubuntu, so the `apt.llvm.org` repo was added.
* It seems that the apt server returns 404 for clang 5.0 binary for i386 arch at the moment. Because of this issue the 32bit build container does not install clang5 for now.
* The 3rd-party `ioft/i386-ubuntu:trusty` base image does not accept the root certificate (w/ 1024bit key) for the `*.googlecode.com` any longer, which is necessary to download submodules of gRPC. I am not sure why we were using that image in the first place... but switched to `i386/ubuntu`, which is managed by Docker.

### DIFF
--- a/env/packages.yml
+++ b/env/packages.yml
@@ -1,7 +1,16 @@
 - hosts: all
   tags: package
   tasks:
-    - apt_repository: repo='ppa:ubuntu-toolchain-r/test' state=present
+    - apt_repository: repo='ppa:ubuntu-toolchain-r/test/ubuntu'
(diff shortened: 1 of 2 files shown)


shinae-woo:
image

PR #665, opened 2017-10-06
state saving and restoring prototypes

chris3torek:
Some modules (like ExactMatch) make it difficult to load up a BESS instance with a full state all at once, or to inspect how they were initially constructed.  To help out, we add `GetArg` and `GetMut` (get generally-immutable-state from the argument used to construct a module, and get the entire generally-mutable state that can be set piecewise in some module).  A similar `SetMut` command sets the mutable state all at once.

Wiki documentation is yet to come based on whether this gets taken in this form.  (Left undone here is the notion of saving, or not, any dynamic state, such as internal NAT tables.  The overall idea is that if for whatever reason one must shoot down an existing BESS instance—or this has already happened because one crashed—and use saved state to create a new one, it would be much easier.)  Not all modules need to support these commands; the plan is to add them when they're convenient.

The current test for exact-match mutable state is a bit sloppy and will be full of redundancies with similar tests for other modules, once the test is more comprehensive, but I think it's better right now to wait for the new test framework.

### DIFF
--- a/bessctl/conf/testing/run_module_tests.bess
+++ b/bessctl/conf/testing/run_module_tests.bess
@@ -37,6 +37,7 @@ import time
 import sugar
 import traceback
 from time import gmtime, strftime
+from pybess import protobuf_to_dict as pb_conv
 
 # Env variable used to run a single test
 test_name = $TEST_NAME!'*'
@@ -134,6 +135,7 @@ def load_test(filename):
         'CRASH_TEST_INPUTS': [],
         'OUTPUT_TEST_INPUTS': [],
         'CUSTOM_TEST_FUNCTIONS': [],
+        'protobuf_to_dict': pb_conv.protobuf_to_dict,
     }
     # pre-import all the actual creator modules directly
     namespace.update(__bess_creators__)
--- a/bessctl/module_tests/exact_match.py
+++ b/bessctl/module_tests/exact_match.py
@@ -29,6 +29,7 @@
 
 import socket
 from test_utils import *
+from pybess import protobuf_to_dict as pb_conv
 
 
 class BessExactMatchTest(BessModuleTestCase):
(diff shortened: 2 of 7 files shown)


chris3torek:
(copying offline request to comment) "... e.g. GetMutable instead of GetMut"

core/modules/exact_match.cc
```diff
@@ -43,6 +43,12 @@ static inline int is_valid_gate(gate_idx_t gate) {
 }
 
 const Commands ExactMatch::cmds = {
+    {"GetArg", "EmptyArg", MODULE_CMD_FUNC(&ExactMatch::GetArg),
+     Command::THREAD_UNSAFE},
+    {"GetMut", "EmptyArg", MODULE_CMD_FUNC(&ExactMatch::GetMut),
+     Command::THREAD_UNSAFE},
```
ddiproietto: Can we mark these as thread safe?

chris3torek: Are we guaranteed one thread at a time executing commands?

melvinw: Yup (https://github.com/NetSys/bess/blob/e11f78916cbe4e4139ad14ea5f31e83ba72b88e2/core/bessctl.cc#L1640)

core/modules/exact_match.cc
```diff
@@ -43,6 +43,12 @@ static inline int is_valid_gate(gate_idx_t gate) {
 }
 
 const Commands ExactMatch::cmds = {
+    {"GetArg", "EmptyArg", MODULE_CMD_FUNC(&ExactMatch::GetArg),
```
sangjinhan: Could you change the names of commands to `lower_case`, just to be consistent with other commands? Since the command names are used as python method names, the existing code avoids `CamelCase` because of pep8...

chris3torek: Sure ... I guess the protobuf names stay CamelCased but should they be, e.g., `ExactMatchConf`?  `ExactMatchConfiguration`?  (Whatever their pattern is, it affects the names retained in the protobuf loader python module.  If that gets complicated maybe I'll use a regexp instead of `.endswith` and literals...)

sangjinhan: Ah, forgot about the `*Arg` constraint. I think `get_initial_arg` also sounds good.

sangjinhan:
Oopsie. Could you please rebase the PR? Sorry I shouldn't have been late this much...

chris3torek:
I've completely rebased this (on the new test framework, with the new exact match table header, etc).  It could use another scan, at least...

bessctl/conf/testing/run_module_tests.bess
```diff
@@ -134,6 +135,7 @@ def load_test(filename):
         'CRASH_TEST_INPUTS': [],
         'OUTPUT_TEST_INPUTS': [],
         'CUSTOM_TEST_FUNCTIONS': [],
+        'protobuf_to_dict': pb_conv.protobuf_to_dict,
```
chris3torek: Side note: this is no longer used, since the exact match test has been converted to the new framework. We could drop the changes to this file entirely now.

bessctl/module_tests/exact_match.py
```diff
@@ -121,6 +122,41 @@ def test_exactmatch_with_metadata(self):
             self.assertEquals(len(pkt_outs[i]), 1)
             self.assertSamePackets(pkt_outs[i][0], test_packet_in)
 
+    def test_exactmatch_selfconfig(self):
+        "make sure get_initial_arg and [gs]et_runtime_config work"
```
chris3torek: It turns out that this docstring gets printed by the unit test code. Delete?

core/utils/exact_match_table.h
```diff
@@ -309,6 +309,10 @@ class ExactMatchTable {
   // Returns the ith field.
   const ExactMatchField &get_field(size_t i) const { return fields_[i]; }
 
+  typename EmTable::iterator begin() { return table_.begin(); }
+
+  typename EmTable::iterator end() { return table_.end(); }
```
chris3torek: I had to make these non-`const` iterators, and I'm not 100% sure why. Should I be using `cbegin` and `cend` here?

barath: You could use `cbegin` or `cend`, but then they would still technically be non-`const` as they would be `const_iterator`s, but either way is fine I think.

barath: (Perhaps we should add a `const_iterator` to cuckoo map?)

chris3torek: I've looked into adding `const_iterator` to cuckoo map, and it seems excessively annoying (the two methods seem to be: write the iterator twice, once as non-const and once as const; or, write a template in place of the two versions of the iterator, then expand it twice to make the non-const and const iterators).

sangjinhan:
Merged, thanks! Sorry for the delay.

15da4ac3904495d45

PR #666, opened 2017-10-06
pull table and key making logic out of ExactMatch

melvinw:
The way the `ExactMatch` module builds its table and lookup keys seems like a good tool to have for future use in other modules, so this PR moves much of that code into a new util called `ExactMatchTable` and reduces the `ExactMatch` module to an instance of `ExactMatchTable<gate_idx_t>`.

Using `bessctl/conf/samples/exactmatch.bess` as a benchmark, the impact on performance seems negligible on my 2.1 GHz Xeon E5-2620 (0.5 Mpps slower)

### DIFF
--- a/core/modules/exact_match.cc
+++ b/core/modules/exact_match.cc
@@ -55,63 +55,36 @@ const Commands ExactMatch::cmds = {
 
 CommandResponse ExactMatch::AddFieldOne(const bess::pb::Field &field,
                                         const bess::pb::FieldData &mask,
-                                        struct EmField *f, int idx) {
-  f->size = field.num_bytes();
-  if (f->size < 1 || f->size > MAX_FIELD_SIZE) {
-    return CommandFailure(EINVAL, "idx %d: 'size' must be 1-%d", idx,
-                          MAX_FIELD_SIZE);
+                                        int idx) {
+  int size = field.num_bytes();
+  uint64_t mask64 = 0;
+  if (mask.encoding_case() == bess::pb::FieldData::kValueInt) {
+    mask64 = mask.value_int();
+  } else if (mask.encoding_case() == bess::pb::FieldData::kValueBin) {
+    bess::utils::Copy(reinterpret_cast<uint8_t *>(&mask64),
+                      mask.value_bin().c_str(), mask.value_bin().size());
   }
 
+  Error ret;
   if (field.position_case() == bess::pb::Field::kAttrName) {
-    const char *attr = field.attr_name().c_str();
-    f->attr_id = AddMetadataAttr(attr, f->size,
-                                 bess::metadata::Attribute::AccessMode::kRead);
-    if (f->attr_id < 0) {
-      return CommandFailure(-f->attr_id, "idx %d: add_metadata_attr() failed",
-                            idx);
+    ret = table_.AddField(this, field.attr_name(), size, mask64, idx);
+    if (ret.first) {
+      return CommandFailure(ret.first, "%s", ret.second.c_str());
     }
   } else if (field.position_case() == bess::pb::Field::kOffset) {
-    f->attr_id = -1;
-    f->offset = field.offset();
-    if (f->offset < 0 || f->offset > 1024) {
-      return CommandFailure(EINVAL, "idx %d: invalid 'offset'", idx);
+    ret = table_.AddField(field.offset(), size, mask64, idx);
+    if (ret.first) {
+      return CommandFailure(ret.first, "%s", ret.second.c_str());
     }
   } else {
-    return CommandFailure(EINVAL, "idx %d: must specify 'offset' or 'attr_name'",
-                          idx);
-  }
-
-  bool force_be = (f->attr_id < 0);
-
-  if (mask.encoding_case() == bess::pb::FieldData::kValueInt) {
-    if (!bess::utils::uint64_to_bin(&f->mask, mask.value_int(), f->size,
-                                    bess::utils::is_be_system() || force_be)) {
-      return CommandFailure(EINVAL, "idx %d: not a correct %d-byte mask", idx,
-                            f->size);
-    }
-  } else if (mask.encoding_case() == bess::pb::FieldData::kValueBin) {
-    if (mask.value_bin().size() != (size_t)f->size) {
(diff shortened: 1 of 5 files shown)


melvinw:
found a bug in metadata mode. marking not-ready for now

melvinw:
Alrighty. Metadata mode works now

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,443 @@
...
+      err_ = Format("idx %d: 'size' must be in [1,%d]", idx, MAX_FIELD_SIZE);
+      return EINVAL;
+    }
+
+    if (mt_attr_name.length() > 0) {
+      f->attr_id = m->AddMetadataAttr(mt_attr_name, f->size,
+                                      metadata::Attribute::AccessMode::kRead);
+      if (f->attr_id < 0) {
+        err_ = Format("idx %d: add_metadata_attr() failed", idx);
+        return -f->attr_id;
```
chris3torek: I was confused for a bit, but I see AddMetadataAttr returns -errno and we return +errno here, OK.

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,443 @@
...
+        return EINVAL;
+      }
+    }
+
+    int force_be = (f->attr_id < 0);
+
+    if (field.mask == 0) {
+      /* by default all bits are considered */
+      f->mask =
+          (f->size == 8) ? 0xFFFFFFFFFFFFFFFFULL : (1ULL << (f->size * 8)) - 1;
```
chris3torek: This one is minor, but I see a couple of cases of "make a bit mask for value" that are like this scattered throughout bess, and they all have the "if 0 then all-1-bits else (1(maybe-ULL) << k) - 1" pattern.  It's short but maybe deserves a dinky little header file for it?

core/utils/exact_match_table_test.cc
```diff
@@ -0,0 +1,152 @@
...
+using bess::utils::ExactMatchField;
+using bess::utils::ExactMatchKey;
+using bess::utils::ExactMatchRuleFields;
+using bess::utils::ExactMatchTable;
+
+TEST(EmTableTest, AddField) {
+  ExactMatchTable<uint8_t> em;
+  int err = em.AddField(0, 4, 0, 0);
+  ASSERT_EQ(0, err);
+  ASSERT_EQ(1, em.num_fields());
```
chris3torek: Below this one can use `EXPECT_EQ` -- it looks like `ASSERT_EQ` suppresses the rest of the tests, while `EXPECT_EQ` logs the failure and keeps going, so as long as there's no `err` here we should be able to test all the other parts.

core/modules/exact_match.cc
```diff
@@ -120,128 +93,75 @@ CommandResponse ExactMatch::Init(const bess::pb::ExactMatchArg &arg) {
...
 
   return CommandSuccess();
 }
 
 void ExactMatch::ProcessBatch(bess::PacketBatch *batch) {
   gate_idx_t default_gate;
-  gate_idx_t out_gates[bess::PacketBatch::kMaxBurst];
-
-  em_hkey_t keys[bess::PacketBatch::kMaxBurst] __ymm_aligned;
+  gate_idx_t out_gates[bess::PacketBatch::kMaxBurst] __ymm_aligned;
```
chris3torek: Oops, forgot to note, I don't think this line needs the `__ymm_aligned`, does it? (Not that it really hurts either)

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,443 @@
...
+    return __AddField(f, mt_attr_name, idx, m);
+  }
+
+  size_t num_fields() const { return num_fields_; }
+
+  // Returns the ith field.
+  const ExactMatchField &get_field(size_t i) const { return fields_[i]; }
+
+  // Returns a helpful error message for the last failed command.
+  const std::string err() const { return err_; }
```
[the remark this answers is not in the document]
melvinw: Ah, yes-- that would be cleaner. Will change

core/utils/exact_match_table.h
melvinw: Ah, a util like that might fit nicely in `utils/bits.h`. I'll open a separate PR for that

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    if ((err = gather_key(fields, &key)).first != 0) {
+      return err;
+    }
+
+    table_.Insert(key, val, ExactMatchKeyHash(total_key_size_),
+                  ExactMatchKeyEq(total_key_size_));
+
+    return MakeError(0, std::string(""));
```
sangjinhan: Just cosmetics, but `MakeError(0, "");` should work fine here. Alternatively, default parameter like this..? `MakeError(int code, const std::string &msg = "")`

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,440 @@
...
+
+  // Turn a rule into a key.
+  // Returns 0 on success, non-zero errno on failure.
+  Error gather_key(const ExactMatchRuleFields &fields, ExactMatchKey *key) {
+    if (fields.size() != num_fields_) {
+      return MakeError(EINVAL, Format("rule should have %zu fields (has %zu)",
+                                      num_fields_, fields.size()));
+    }
+
+    memset(key, 0, sizeof(*key));
```
sangjinhan: minor: `*key = {};`

core/utils/exact_match_table.h
```diff
@@ -0,0 +1,440 @@
...
+
+      memcpy(reinterpret_cast<uint8_t *>(key) + field_pos, f_obj.data(),
+             field_size);
+    }
+
+    return MakeError(0, std::string(""));
+  }
+
+  // Helper for public MakeKey functions
+  void __MakeKeys(ExactMatchKey *keys, const void **bufs, size_t n) const {
```
sangjinhan: Single `_` should be used here, since we are not compiler or low level library. Or `DoMakeKeys` naming convention can be used.

melvinw:
All fixed up and rebased

sangjinhan:
Merged, thanks!
(the Travis error is an unrelated issue)
74561c1bcf64886d70cec9b145cf046c

PR #667, opened 2017-10-06
Avoid undefined behavior of __builtin_clz()

sangjinhan:
The code `uint32_t prefix_length() const { return ((32 - __builtin_clz(mask.raw_value()))); }` may fail if the value of `mask` is 0. In this case the behavior of `__builtin_clz` is undefined. While it does work as intended on newer x86 CPUs with BMI instruction set, the function on older CPUs returns some garbage value. This caused occasional failures on Travis.

Another issue in the same code is that it works only when prefix length is a multiple of 8. Bit reversing should have been done instead of byte reversing. It was fixed by using `ctz` in place of `clz`.


apanda:
LGTM

shinae-woo:
image

PR #669, opened 2017-10-07
Add RandomSplit module by extending RandomDrop

shinae-woo:
This PR addresses the suggestion made in #298, as now I need for my evaluation. 😉


core/modules/random_split.cc
```diff
@@ -0,0 +1,119 @@
...
+
+  for (int i = 0; i < arg.gates_size(); i++) {
+    int elem = arg.gates(i);
+    gates_[i] = elem;
+
+    if (!is_valid_gate(gates_[i])) {
+      return CommandFailure(EINVAL, "Invalid gate %d", gates_[i]);
+    }
+  }
+  ngates_ = arg.gates_size();
```
chris3torek: We should probably verify _all_ the `arg.gates(i)` values before setting any of the `gates_[i]` values, but in any case, it would certainly be better to verify `elem` before assigning it to `gates_[i]` here, so that we don't leave a bogus value in the `gates_` array.

It would also be nice to have both `::Init` and `::CommandSetGates` use the same verification and setting code (though that's less important—anything bogus during `Init` gets wiped away, as if it never happened...).

chris3torek:
Oh, somehow I missed the link to #298. Although I'm still not sure what the motivation is 😺

shinae-woo:
This is not for a use case in reality, but for performance benchmarks (i.e., emulating arbitrary branch splitting). A further extension could be adding some burstiness. 

Will do the cleanup part!

shinae-woo:
Now there should be no more bogus if a command fails. Thank you so much!

Sharing parameter setting code should be desirable, but now it is not straightforward since each function has different types of arguments. This is not critical but keep tickled. As I follow the rules from legacy code, the arguments are 'RandomSplitArg', 'RandomSplitCommandSetArg'. We don't need to necessarily have them differently??

sangjinhan:
One minor code cleanup: the old code used `kRange`, `threshold_`, and `GetRange()`. In the new code you may want to replace them with `drop_rate_` and `GetReal()` for simpler (slightly more efficient) code.

Otherwise looks good to me.

shinae-woo:
@sangjinhan simplification done!

shinae-woo:
I prefer merging this PR after we merged PR for new test framework, as I expect conflict between them.

shinae-woo:
Rebased with #663

bessctl/module_tests/random_split.py
```diff
@@ -30,17 +30,17 @@
 from test_utils import *
 
 
-class BessRandomDropTest(BessModuleTestCase):
+class BessRandomSplitTest(BessModuleTestCase):
 
     def test_dropall(self):
```
sangjinhan: While this is not a part of this PR, but noticed that the names "dropall" and "dropnone" are actually swapped...?

shinae-woo: Fixed.

sangjinhan:
Thanks!
tumblr_m4px4dczqs1r6jd7fo1_1280

PR #670, opened 2017-10-08
Fix Travis build issues

sangjinhan:
1. It seems that Travis VMs recently are more often run on old systems (SandyBridge or older). We use DPDK binary prebuilt in the build container. Since the binary is built with RDRAND support with `-march=native`, all apps linked against the DPDK binary fail if Travis VM runs on an old processor. As a hot fix, now DPDK is built locally in the Travis VM.

1. Also, now the build more frequently fails due to the 10-min timeout. `travis_wait` is used to extend the timeout to 20mins.

1. We switched to a 32-bit Ubuntu container image maintained by Docket (#664). Within the new container `arch` reports `x86_64`, not `i686`, causing build errors. I replaced `arch` with 'gcc -dumpmachine` to work around this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,8 @@ before_script:
   - export CC=$VER_CC
 
 script:
-  - ./container_build.py bess
+  # travis_wait extends the 10-min timeout to 20mins.
+  - travis_wait ./container_build.py bess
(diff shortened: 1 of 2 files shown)


melvinw:
Thanks for doing this! cat

PR #673, opened 2017-10-11
Refactoring Module/Task/ModuleBuilder

shinae-woo:
This PR doesn't change any functionality, but refactor the existing `module.h` and `module.cc` files as they are integrating too much functionality in a file. Once these changes are approved, then I can make change the internal scheduling logic.

Major changes are
- Get global info about `module_graphs` out of `ModuleBuilder` class
- Get `Task/ModuleTask` out of `module.h`

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -1137,7 +1137,7 @@ class BESSControlImpl final : public BESSControl::Service {
       mod_name = request->name();
     } else {
       mod_name = ModuleGraph::GenerateDefaultName(builder.class_name(),
-                                                    builder.name_template());
+                                                  builder.name_template());
     }
 
     pb_error_t* error = response->mutable_error();
@@ -1241,10 +1241,10 @@ class BESSControlImpl final : public BESSControl::Service {
 
     if (is_any_worker_running()) {
       propagate_active_worker();
-      if (m1->num_active_workers() || m2->num_active_workers() ) {
-	WorkerPauser wp; // Only pause when absolutely required
-	ret = m1->ConnectModules(ogate, m2, igate);
-	goto done;
+      if (m1->num_active_workers() || m2->num_active_workers()) {
+        WorkerPauser wp;  // Only pause when absolutely required
+        ret = m1->ConnectModules(ogate, m2, igate);
+        goto done;
       }
     }
     ret = m1->ConnectModules(ogate, m2, igate);
@@ -1286,21 +1286,22 @@ class BESSControlImpl final : public BESSControl::Service {
     return Status::OK;
   }
 
-  Status DumpMempool(ServerContext*,
-                           const DumpMempoolRequest* request,
-                           DumpMempoolResponse* response) override {
+  Status DumpMempool(ServerContext*, const DumpMempoolRequest* request,
+                     DumpMempoolResponse* response) override {
     int socket_filter = request->socket();
-    socket_filter = (socket_filter == -1) ? (RTE_MAX_NUMA_NODES - 1) : socket_filter;
+    socket_filter =
+        (socket_filter == -1) ? (RTE_MAX_NUMA_NODES - 1) : socket_filter;
     int socket = (request->socket() == -1) ? 0 : socket_filter;
     for (; socket <= socket_filter; socket++) {
-      struct rte_mempool *mempool = bess::get_pframe_pool_socket(socket);
-      MempoolDump *dump = response->add_dumps();
+      struct rte_mempool* mempool = bess::get_pframe_pool_socket(socket);
+      MempoolDump* dump = response->add_dumps();
       dump->set_socket(socket);
       dump->set_initialized(mempool != nullptr);
       if (mempool == nullptr) {
         continue;
       }
-      struct rte_ring *ring = reinterpret_cast<struct rte_ring*>(mempool->pool_data);
+      struct rte_ring* ring =
+          reinterpret_cast<struct rte_ring*>(mempool->pool_data);
       dump->set_mp_size(mempool->size);
       dump->set_mp_cache_size(mempool->cache_size);
       dump->set_mp_element_size(mempool->elt_size);
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -50,6 +50,7 @@
 #include "message.h"
 #include "metadata.h"
 #include "module.h"
+#include "module_graph.h"
 #include "opts.h"
 #include "packet.h"
 #include "port.h"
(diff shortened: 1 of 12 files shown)


core/module.h
```diff
@@ -192,14 +183,44 @@ class ModuleBuilder {
...
+  const Commands cmds_;
+  const module_init_func_t init_func_;
+};
+
+// A static class for managing module graphs
+class ModuleGraph {
+ public:
+
+  // Return true if any module from the builder exists
+  static bool ExistModuleClass(const ModuleBuilder *);
```
chris3torek: nit: `HasModuleOfClass` might be a better name (although I'm not sure I really like this name either—the tricky bit is that we're asking whether any module exists using the class _implied by_ the given `ModuleBuilder` instance, which is hard to encode in the predicate name...).

core/bessctl.cc
```diff
@@ -1241,10 +1241,10 @@ class BESSControlImpl final : public BESSControl::Service {
 
     if (is_any_worker_running()) {
       propagate_active_worker();
-      if (m1->num_active_workers() || m2->num_active_workers() ) {
-	WorkerPauser wp; // Only pause when absolutely required
-	ret = m1->ConnectModules(ogate, m2, igate);
-	goto done;
+      if (m1->num_active_workers() || m2->num_active_workers()) {
+        WorkerPauser wp;  // Only pause when absolutely required
```
chris3torek: FYI I think this reformat might be the same as one in one of @melvinw PRs, hopefully they match exactly and Git will be able to handle it easily 😺

barath:
Will look.

core/module.h
shinae-woo: Really, it is better!

core/module.h
```diff
@@ -132,7 +97,8 @@ struct Command {
 
 using Commands = std::vector<struct Command>;
 
-// A class for generating new Modules of a particular type.
+// A class for managing modules of 'a particular type'.
+// It creates new modules and forwarding module-specific commands
```
barath: Minor: perhaps rewrite as `Creates new modules and forwards module-specific commands.`

core/module.h
```diff
@@ -150,34 +116,24 @@ class ModuleBuilder {
...
                                   module_init_func_t init_func);
-
   static bool DeregisterModuleClass(const std::string &class_name);
 
   static std::map<std::string, ModuleBuilder> &all_module_builders_holder(
       bool reset = false);
   static const std::map<std::string, ModuleBuilder> &all_module_builders();
 
-  static const std::map<std::string, Module *> &all_modules();
+  /* returns a pointer to the created module.
```
barath: Perhaps update the comment?

core/module_graph.cc
```diff
@@ -0,0 +1,254 @@
+// Copyright (c) 2014-2016, The Regents of the University of California.
```
barath: I think this copyright line now should be 2014-2017.

core/module_graph.cc
```diff
@@ -0,0 +1,254 @@
...
+
+#include <glog/logging.h>
+
+#include "module.h"
+
+std::map<std::string, Module *> ModuleGraph::all_modules_;
+std::unordered_map<std::string, Node> ModuleGraph::module_graph_;
+std::unordered_set<std::string> ModuleGraph::tasks_;
+
+// Finds the next module that implements a task, and updates it's parents
```
barath: Minor: we can omit comments in the .cc file if it's also in the .h

core/module_graph.cc
```diff
@@ -0,0 +1,254 @@
...
+  if (erase) {
+    all_modules_.erase(m->name());
+  }
+
+  module_graph_.erase(m->name());
+  if (m->is_task_) {
+    tasks_.erase(m->name());
+  }
+
+  delete m;
```
barath: `m` appears to only get deleted here; what happens if one of the other execution paths is taken?  If we need to delete in those cases too, maybe we can put m into a `std::unique_ptr` at the top?

core/module_graph.h
```diff
@@ -0,0 +1,121 @@
...
+
+  // Disconnects two modules (`to` and `from`) together in `module_graph_`.
+  static bool RemoveEdge(const std::string &from, const std::string &to);
+
+ private:
+  // Updates the parents of modules with tasks by traversing `module_graph_` and
+  // ignoring all modules that are not tasks.
+  static bool UpdateTaskGraph();
+
+  // Finds the next module that implements a task, and updates it's parents
```
barath: I'm not sure I understand from this comment what this method does.  (I know the comment was already there, just figured we might clarify it.)

Minor: its

core/module_graph.h
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) 2014-2016, The Regents of the University of California.
```
barath: 2014-2017

core/module_graph.h
```diff
@@ -0,0 +1,121 @@
...
+  // Module that this Node represents.
+  Module *module_;
+
+  // Children of `module_` in the pipeline.
+  std::unordered_set<std::string> children_;
+
+  DISALLOW_COPY_AND_ASSIGN(Node);
+};
+
+class ModuleGraph {
```
barath: Could you add a class comment?

core/task.cc
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) 2014-2016, The Regents of the University of California.
```
barath: 2014-2017

core/task.h
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2014-2016, The Regents of the University of California.
```
barath: 2014-2017

barath:
LGTM.  Just a few minor things.

core/module_graph.cc
shinae-woo: Yes, definitely it would be better if we make the ownership of modules (as well as TCs, tasks) much clearer.

shinae-woo:
Thank you for the review, @barath. Mostly addressed. The ownership parts can be addressed separately so that makes changes gradually. 

@ddiproietto, Thank you for confirming the performance. I often found that compilers are much smarter and more efficient than myself 😢...

core/module_graph.cc
shinae-woo: I am creating a new PR to address the issues handling module creation and deletion correctly. Not only with `DestroyModule()`, but also between `CreateModule()` and `AddModule()` has the same problem.

sangjinhan:
Merged, thanks!

tumblr_m4pxpd98cr1r6jd7fo1_1280

PR #676, opened 2017-10-12
Placement constraint test

shinae-woo:
This PR is for porting existing module placement constraint test (but were not checked with Travis) into the new testing framework.


sangjinhan:
👍 
default-4

PR #681, opened 2017-10-14
experimental_scheduler: fix potential segfault

melvinw:
Currently the experimental scheduler keeps pointers to the tasks
registered by modules in its WakeUpQueue after they have been deleted.
This could lead to nasty crashes. This commit makes two changes:

- Change Scheduler::WakeUpQueue to use utils/extended_priority_queue
  instead of stl::priority_queue in order to support deletion in the
  middle.

- Change Module::DestroyAllTasks() to remove any references to
  registered tasks from its scheduler's wake up queue.

### DIFF
--- a/core/module.cc
+++ b/core/module.cc
@@ -355,6 +355,13 @@ task_id_t Module::RegisterTask(void *arg) {
 void Module::DestroyAllTasks() {
   for (auto task : tasks_) {
     auto c = task->GetTC();
+
+    int wid = c->WorkerId();
+    if (wid >= 0) {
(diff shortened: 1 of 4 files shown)


shinae-woo:
Thanks! I was doing some refactoring on TC and ModuleTask, to ensure that objects are freed where they were allocated. That refactoring and this PR are very likely to conflict though. If this issue is not urgent, let me include this fix (or its equivalent) in my later PR. 😸

melvinw:
Ah, nice-- please take it! It's not super urgent :)

barath:
LGTM

melvinw:
Will be fixed in #673

PR #682, opened 2017-10-16
Unify the responsibility of module creation/deletion to the ModuleGraph

shinae-woo:
* No other class (e.g., bessctl, bessd) should take the ownership

### DIFF
--- a/core/module_graph.cc
+++ b/core/module_graph.cc
@@ -128,22 +128,58 @@ bool ModuleGraph::RemoveEdge(const std::string &from, const std::string &to) {
   return UpdateTaskGraph();
 }
 
-bool ModuleGraph::AddModule(Module *m) {
+// Creates a module to the graph.
+Module *ModuleGraph::CreateModule(const ModuleBuilder &builder,
+                                  const std::string &module_name,
+                                  const google::protobuf::Any &arg,
+                                  pb_error_t *perr) {
+  Module *m =
+      builder.CreateModule(module_name, &bess::metadata::default_pipeline);
+
+  CommandResponse ret = m->InitWithGenericArg(arg);
+  {
+    google::protobuf::Any empty;
+
+    if (ret.data().SerializeAsString() != empty.SerializeAsString()) {
+      LOG(WARNING) << module_name << "::" << builder.class_name()
+                   << " Init() returned non-empty response: "
+                   << ret.data().DebugString();
+    }
+  }
+
+  if (ret.error().code() != 0) {
+    *perr = ret.error();
+    delete m;
+    return nullptr;
+  }
+
   if (m->is_task_) {
     if (!tasks_.insert(m->name()).second) {
-      return false;
+      *perr = pb_errno(ENOMEM);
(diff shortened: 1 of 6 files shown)


core/module_graph.cc
```diff
@@ -128,22 +128,58 @@ bool ModuleGraph::RemoveEdge(const std::string &from, const std::string &to) {
   return UpdateTaskGraph();
 }
 
-bool ModuleGraph::AddModule(Module *m) {
+// Creates a module to the graph.
+Module *ModuleGraph::CreateModule(const ModuleBuilder &builder,
+                                  const std::string &module_name,
+                                  const google::protobuf::Any &arg,
+                                  pb_error_t *perr) {
+  Module *m =
```
barath: Maybe this could be a std::unique_ptr?

core/module_test.cc
```diff
@@ -239,12 +235,11 @@ TEST_F(ModuleTester, RunCommand) {
...
 
 TEST_F(ModuleTester, ConnectModules) {
+  pb_error_t perr;
   Module *m1, *m2;
 
-  EXPECT_EQ(0, create_acme("m1", &m1));
-  ASSERT_NE(nullptr, m1);
-  EXPECT_EQ(0, create_acme("m2", &m2));
-  ASSERT_NE(nullptr, m2);
+  EXPECT_NE(nullptr, m1 = create_acme("m1", &perr));
```
sangjinhan: `ASSERT_NE` seems necessary here.

core/module_test.cc
```diff
@@ -295,14 +292,13 @@ TEST_F(ModuleTester, GenerateTCGraph) {
...
    */
-  Module *t1, *t2, *t3, *t4, *m1, *m2, *m3;
-  EXPECT_EQ(0, create_acme("m1", &m1));
-  EXPECT_EQ(0, create_acme("m2", &m2));
-  EXPECT_EQ(0, create_acme("m3", &m3));
-  EXPECT_EQ(0, create_acme_with_task("t1", &t1));
-  EXPECT_EQ(0, create_acme_with_task("t2", &t2));
-  EXPECT_EQ(0, create_acme_with_task("t3", &t3));
-  EXPECT_EQ(0, create_acme_with_task("t4", &t4));
+  EXPECT_NE(nullptr, m1 = create_acme("m1", &perr));
```
sangjinhan: Existing code had `EXPECT_EQ` here, `ASSERT_*` should be used for null pointer check.

core/module_graph.cc
shinae-woo: Unfortunately, it is somewhat awkward to use `unique_ptr` here since the pointer is stored in separate data structures in the middle of the function. :(

chris3torek: You could `.release` the `unique_ptr` but I agree that it gets awkward.

barath: No worries.

sangjinhan:
Thanks!
tumblr_m4opb2ldgz1qzfanlo1_1280

PR #683, opened 2017-10-16
Fix starvation of the RX path in UnixSocketPort

sangjinhan:
Due to the cost of system calls, UnixSocketPort had a throttling mechanism for RX busy-wait polling. When the port is idle, `recv()` is invoked every 256 ticks (in other words, scheduling rounds). This behavior becomes problematic when it is used with the experimental scheduler. Since the scheduler slows down idle tasks, when combined with UnixSocketPort's internal mechanism, `recv()` is hardly invoked (every 256 * 2^32 cycles, which translates to a few minutes). This PR makes two changes:

* The maximum exponential backoff of the experimental scheculer is now 2^20 cycles,  reduced from 2^32 cycles.
* UnixSocketPort throttles based on wall-clock time (by default 50us, but configurable), not on tick.

Note: while this throttling feature looks redundant with the role of RateLimit traffic class, it is slightly different. The RX throttling kicks in only when the port is idle. It is rather similar to interrupt coalescing in kernel device drivers.

### DIFF
--- a/core/drivers/unix_socket.cc
+++ b/core/drivers/unix_socket.cc
@@ -28,7 +28,6 @@
 // ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
-
 #include <glog/logging.h>
 #include <poll.h>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


melvinw:
Thanks for doing this! cat

PR #685, opened 2017-10-16
Refactoring task, scheduler

shinae-woo:
This PR does not change functionality but simplify the Task and Scheduler. 

The main changes are:
1. Remove ModuleTask
2. De-templatize scheduler and leaf traffic class
3. Merge @melvinw's PR #681 for fix potential segfault.


chris3torek:
I scanned through this and the changes look fine to me, assuming there's no reason to keep the complexity of type-generic scheduler.

I did notice that `core/utils/extended_priority_queue.h` misses a trick: its `delete_single_element` does the required linear search and then just makes a whole new heap, rather than re-sifting the heap a la its `decrease_key_top`.  (If this is rare, though, that's probably OK, and of course O(2N) is still O(N)....)

sangjinhan:
Merged, thanks!

tumblr_m4phw5a9q71qenqklo1_1280

PR #686, opened 2017-10-17
fix race in unix_socket.cc

chris3torek:
As unlikely as it seems, the Python tests for bessd are triggering
a race between the unix-socket accept() thread and the vlan test
code running in the Python interpreter.

There is a long comment in bessctl/test_utils.py explaining the race.
On the BESS side, we add a new boolean, `confirm_connect`, to the unix-
socket port, and send a magic number -- the byte-string "yes\0" --
across the socket to indicate that the accept thread has set it all up
for use.

### DIFF
--- a/protobuf/ports/port_msg.proto
+++ b/protobuf/ports/port_msg.proto
@@ -53,6 +53,11 @@ message UnixSocketPortArg {
   /// Use a negative number for unthrottled polling. If unspecified or 0,
   /// it is set to 50,000 (50 microseconds, or 20k polls per second)
   int64 min_rx_interval_ns = 2;
+
+  /// If set, the port driver will send a confirmation once
+  /// the port is connected.  This lets pybess avoid a race during
+  /// testing.  See bessctl/test_utils.py for details.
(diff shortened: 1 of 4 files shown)


protobuf/ports/port_msg.proto
```diff
@@ -53,6 +53,11 @@ message UnixSocketPortArg {
   /// Use a negative number for unthrottled polling. If unspecified or 0,
   /// it is set to 50,000 (50 microseconds, or 20k polls per second)
   int64 min_rx_interval_ns = 2;
+
+  /// If set, the port driver will send a confirmation once
+  /// the port is connected.  This lets pybess avoid a race during
+  /// testing.  See bessctl/test_utils.py for details.
+  bool confirm_connect = 3;
```
chris3torek: I'll make two notes here on my own commit: (1) perhaps we should just always confirm? (2) if not, would it be better to make the confirm-packet-contents a `bytes` here rather than a magic string? (I was originally going to use a zero-size packet...)

sangjinhan:
i15410743199

PR #687, opened 2017-10-17
clean up unittests a bit

chris3torek:
These three commits should each stand alone - I'm just bundling them together because they happened in sort of reverse order, as part of debugging the vlan race thing.  The first is important, the second doesn't affect any current test, and the third commit has no actual fix in it, it's just some code I had trouble with while debugging, so I rearranged it.

Note that these changes also need the vlan-racy-test fix to be installed for them to work on my laptop. Anyone not hitting the race should be able to run this version without that, though.

### DIFF
--- a/bessctl/module_tests/vlan.py
+++ b/bessctl/module_tests/vlan.py
@@ -34,12 +34,8 @@ from test_utils import *
 class BessVlanTest(BessModuleTestCase):
 
     def test_run_vlan(self):
-        if sys.version_info[0] != 2:
-            print(
-                "python3-compatible scapy does not have Dot1AD. Skipping vlan test...")
-        else:
-            self.run_for(VLANSplit(), [0], 3)
-            self.assertBessAlive()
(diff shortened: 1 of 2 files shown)


bessctl/module_tests/vlan.py
```diff
@@ -34,12 +34,8 @@
...
 
     def test_run_vlan(self):
-        if sys.version_info[0] != 2:
-            print(
-                "python3-compatible scapy does not have Dot1AD. Skipping vlan test...")
-        else:
-            self.run_for(VLANSplit(), [0], 3)
-            self.assertBessAlive()
+        self.run_for(VLANSplit(), [0], 3)
+        self.assertBessAlive()
```
shinae-woo: Shall we skip this as well?

shinae-woo: Aha, I see.

sangjinhan:
Merged, thanks!

PR #688, opened 2017-10-18
misc sanitizer fixes

melvinw:
While building with `clang++-3.9` and `SANITIZE=1`on my machine I found some new complaints from the leak and undefined behavior and leak sanitizers.

undefined behavior:

```
utils/ip.cc:82:39: runtime error: signed integer overflow: -2147483648 - 1 cannot be represented in type 'int'
    #0 0x185bfd7 in bess::utils::Ipv4Prefix::Ipv4Prefix(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) utils/ip.cc:82:39
    #1 0xb1d68d in (anonymous namespace)::IPTest_PrefixCalc_Test::TestBody() utils/ip_test.cc:97:16
    #2 0x11a70bf in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /usr/src/gtest/src/gtest.cc:2078:10
    #3 0x11a70bf in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /usr/src/gtest/src/gtest.cc:2114
    #4 0x116584c in testing::Test::Run() /usr/src/gtest/src/gtest.cc:2150:5
    #5 0x1167a36 in testing::TestInfo::Run() /usr/src/gtest/src/gtest.cc:2326:11
    #6 0x1169c0d in testing::TestCase::Run() /usr/src/gtest/src/gtest.cc:2444:28
    #7 0x1181bd5 in testing::internal::UnitTestImpl::RunAllTests() /usr/src/gtest/src/gtest.cc:4315:43
    #8 0x11adba8 in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /usr/src/gtest/src/gtest.cc:2078:10
    #9 0x11adba8 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /usr/src/gtest/src/gtest.cc:2114
    #10 0x1180229 in testing::UnitTest::Run() /usr/src/gtest/src/gtest.cc:3926:10
    #11 0xa683f0 in RUN_ALL_TESTS() /usr/include/gtest/gtest.h:2288:46
    #12 0xa683f0 in main gtest_main.cc:40
    #13 0x7fc30870d82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #14 0x897aa8 in _start (all_test+0x897aa8)
```

the leak:

```
Direct leak of 1 byte(s) in 1 object(s) allocated from:
    #0 0x7cd5e0 in __strdup (/home/ubuntu/bess/core/all_test+0x7cd5e0)
    #1 0x17d99f8 in pcap_create_common (/home/ubuntu/bess/core/all_test+0x17d99f8)
    #2 0x7ffbffffffff  (<unknown module>)
    #3 0xf618af in void std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_construct<char const*>(char const*, char const*, std::forward_iterator_tag) /usr/bin/../lib/gcc/x86_64-linux-gnu/7.2.0/../../../../include/c++/7.2.0/bits/basic_string.tcc:232:2
```

### DIFF
--- a/core/lsan.suppress
+++ b/core/lsan.suppress
@@ -5,7 +5,6 @@
 # there are some internal memory leaks in libpcap
 leak:pcap_alloc_pcap_t
 leak:pcap_check_header
-leak:pcap_open_offline_common
 
 # at startup in gtest
 leak:__si_class_type_info::__do_dyncast
--- a/core/lsan.suppress
+++ b/core/lsan.suppress
@@ -2,9 +2,12 @@
 # clang's LeakSanitizer. For more details, see:
 # https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer
 
-# there are some internal memory leaks in libpcap
+# there are some internal memory leaks in older versions of libpcap (see:
+# https://github.com/the-tcpdump-group/libpcap/commit/cd46a74cf492). For now
+# suppress leaks from these functions
 leak:pcap_alloc_pcap_t
 leak:pcap_check_header
+leak:pcap_create_common
 
 # at startup in gtest
 leak:__si_class_type_info::__do_dyncast
--- a/core/utils/ip.cc
+++ b/core/utils/ip.cc
@@ -79,7 +79,7 @@ Ipv4Prefix::Ipv4Prefix(const std::string &prefix) {
   } else if (len >= 32) {
     mask = be32_t(0xffffffff);
   } else {
-    mask = be32_t(~((1 << (32 - len)) - 1));
+    mask = be32_t(~((1u << (32 - len)) - 1));
   }
 }
 
--- a/core/utils/tcp_flow_reconstruct_test.cc
+++ b/core/utils/tcp_flow_reconstruct_test.cc
@@ -50,13 +50,13 @@ class TcpFlowReconstructTest : public ::testing::TestWithParam<const char *> {
     std::string tracefile_prefix = "testdata/test-pktcaptures/tcpflow-http-3";
 
     char errbuf[PCAP_ERRBUF_SIZE];
-    pcap_t *handle =
+    handle_ =
         pcap_open_offline((tracefile_prefix + ".pcap").c_str(), errbuf);
-    ASSERT_TRUE(handle != nullptr);
+    ASSERT_TRUE(handle_ != nullptr);
 
     const u_char *pcap_pkt;
     struct pcap_pkthdr pcap_hdr;
-    while ((pcap_pkt = pcap_next(handle, &pcap_hdr)) != nullptr) {
+    while ((pcap_pkt = pcap_next(handle_, &pcap_hdr)) != nullptr) {
       ASSERT_EQ(pcap_hdr.caplen, pcap_hdr.len)
           << "Didn't capture the full packet.";
       Packet *p = new Packet();
@@ -80,6 +80,9 @@ class TcpFlowReconstructTest : public ::testing::TestWithParam<const char *> {
     for (Packet *p : pkts_) {
       delete p;
     }
+    if (handle_) {
+      pcap_close(handle_);
+    }
   }
 
   // The packets of the pcap trace file.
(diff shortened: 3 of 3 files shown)


core/utils/ip.cc
```diff
@@ -79,7 +79,7 @@ Ipv4Prefix::Ipv4Prefix(const std::string &prefix) {
   } else if (len >= 32) {
     mask = be32_t(0xffffffff);
   } else {
-    mask = be32_t(~((1 << (32 - len)) - 1));
+    mask = be32_t(~((1u << (32 - len)) - 1));
```
chris3torek: This is one of those things that should go into that field-of-bits utility header file we were talking about. (meanwhile, fine for now)

core/lsan.suppress
```diff
@@ -6,6 +6,7 @@
 leak:pcap_alloc_pcap_t
 leak:pcap_check_header
 leak:pcap_open_offline_common
+leak:pcap_create_common
```
ddiproietto: I think the leak is in our code: if I call `pcap_close(handle)` in `TcpFlowReconstructTest::SetUp()`, the LeakSanitizer is happy.

If you confirm that, maybe we can remove all the pcap_* suppressions

melvinw: Ah, neat. Will amend after reproducing

sangjinhan: I think what @ddiproietto meant was calling `pcap_close(handle)` at the end of `SetUp()`, not `TearDown()`. Since `handle` is only used in the setup method it doesn't need to be a class member...

sangjinhan: from offline discussion with @melvinw: it is for exception safety.

btw, why are we removing the three existing pcap functions?

melvinw: While debugging this, I noticed that they weren't being tripped anymore (may or may not be a result of the added `pcap_close()`) Happy to put them back if needed

sangjinhan: Ah, I see! @ddiproietto said that you may be able to remove all `pcap_*` functions. Do you still see `pcap_create_common` in the leak report?

melvinw: Yup, we (mostly @ddiproietto) are debugging the last source of the `pcap_create_common()` leak. There's a test case in `pcap_handle_test` that causes the leak under __very__ specific reproduction conditions.

ddiproietto: The `leak:pcap_open_offline_common` was necessary because of the bug in our code, Now that we fix the bug, I think we can remove it.

I guess the rest of them are necessary on older versions of libpcap (i.e. the one packaged in ubuntu16.04) that don't include commit the-tcpdump-group/libpcap@cd46a74cf492 ("Use pcap_close() if pcap_create_handle() fails.")

Sorry, it was working on my system (just because I had another version of tcpdump), and I hoped we could remove the suppressions

How about keeping the suppressions (except open_offline_common) and adding a link to the commit that fixes the problem in libpcap, so that I can remember it next time I look at the suppressions? 😄

Sorry, I'm begin way too picky on this!

melvinw: Sounds good to me. Will put them back

melvinw: Alrighty, added some old ones back and updated the comment

melvinw:
Weird... I can't reproduce this new leak locally.

ddiproietto:
gatto

PR #691, opened 2017-10-19
Highlight .bess files as Python

pchaigno:
Define a new [Linguist override rule](https://github.com/github/linguist#overrides) to recognize all `.bess` files as
Python files.

This will, as far as I know, only impact github.com.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.bess  linguist-language=Python


sangjinhan:
Great, thanks!

(for future reference for others: vi users can add `autocmd BufNewFile,BufRead *.bess set filetype=python` for syntax coloring)

sangjinhan:
tumblr_m4ol1bdgme1qd477zo1_500

PR #692, opened 2017-10-19
utils/bits: add some mask generation utilities

melvinw:


melvinw:
I only replaced two occurrences of this pattern. If there are others I missed I'm happy to patch them and append to this PR.

chris3torek:
There's one in `core/modules/split.cc`:

      mask_ = (size_ == 8) ? 0xffffffffffffffffull : (1ull << (size_ * 8)) - 1;

(line 47).

(There's one more in core/modules.h for socket mask but it doesn't really fit the pattern and probably shouldn't be changed.)

core/utils/bits.h
```diff
@@ -163,6 +163,24 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
   }
 }
 
+// TODO(melvinw): it would be nice if MaskTopN() supported blobs.
+// Returns a mask for the first `n` bits of an integral type T, i.e. starting
+// from the most significant bit toward the least significant.
+// For example: MaskTopN<uint32_t>(9) will return 0xF8000000
+template <typename T, typename = std::enable_if<std::is_integral<T>::value>>
+static inline T MaskTopN(size_t n) {
+  return (n == sizeof(T) * 8) ? ~T{0} : (T{1} << n) - 1;
```
chris3torek: Question: should this use >= here, in which case, we don't need the other test in `ip.cc`? It's not clear to me what it means to ask for the top 47 bits in a 32 bit mask...

core/utils/bits.h
```diff
@@ -163,6 +163,24 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
...
+}
+
+// TODO(melvinw): it would be nice if MaskBottomN() supported blobs.
+// Returns a mask for the last `n` bits of an integral type T, i.e. starting
+// from the least significant bit toward the most signifant.
+// For example: MaskBottomN<uint32_t>(9) will return 0x0000001F
+template <typename T, typename = std::enable_if<std::is_integral<T>::value>>
+static inline T MaskBottomN(size_t n) {
+  return ~MaskTopN<T>(n);
```
sangjinhan: should be`32 - n` or so...?

core/utils/bits.h
```diff
@@ -163,6 +163,24 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
   }
 }
 
+// TODO(melvinw): it would be nice if MaskTopN() supported blobs.
+// Returns a mask for the first `n` bits of an integral type T, i.e. starting
+// from the most significant bit toward the least significant.
+// For example: MaskTopN<uint32_t>(9) will return 0xF8000000
+template <typename T, typename = std::enable_if<std::is_integral<T>::value>>
+static inline T MaskTopN(size_t n) {
```
sangjinhan: The name Mask here might be ambiguous here; since its definition could be different depending on the contexts (masking on? masking off?). How about more explicit names, such as `SetBitsHi()/SetBitsLo()` or so?

core/utils/bits.h
melvinw: Good point! Fixed

core/utils/bits.h
melvinw: Ah, I like `SetBits`-- will change

core/utils/bits.h
melvinw: You're right. Fixed

core/utils/bits.h
```diff
@@ -163,6 +163,26 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
   }
 }
 
+// TODO(melvinw): it would be nice if MaskTopN() supported blobs.
+// Returns a mask for the first `n` bits of an integral type `T`, i.e. starting
+// from the most significant bit toward the least significant. Returns a `T`
+// with all bits set if `n` is greater than the number of bits in `T`.
+// For example: MaskTopN<uint32_t>(9) will return 0xF8000000
```
chris3torek: Occurs to me on rescan, 9 is `0xFF800000`. `0xF8000000` is what you get for 5...

melvinw: Doh! Good catch

melvinw: Fixed and rebase

core/utils/bits.h
```diff
@@ -163,6 +163,26 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
...
+template <typename T, typename = std::enable_if<std::is_integral<T>::value>>
+static inline T SetBitsHigh(size_t n) {
+  return (n >= sizeof(T) * 8) ? ~T{0} : (T{1} << n) - 1;
+}
+
+// TODO(melvinw): it would be nice if MaskBottomN() supported blobs.
+// Returns a mask for the last `n` bits of an integral type `T`, i.e. starting
+// from the least significant bit toward the most signifant. Returns a `T` with
+// all bits set if `n` is greater than the number of bits in `T`.
+// For example: MaskBottomN<uint32_t>(9) will return 0x000001FF
```
chris3torek: This comment is wrong, because `size_t` is unsigned: SetBitsLow<uint32_t>(31) works great (we get `~SetBitsHigh<uint32_t>(1)`), 32 works as well (32-32 = 0 = set no high bits = set all bits), but 33 means we do 32 - 33 = -1, convert to `size_t` and get some big number, big number > 32, `SetBitsHigh` returns 0xFFFFFFFF, and we get 0. This would need a similar (`n >= sizeof(T) * 8)` test.

(also, both comments need the new name 😺 )

melvinw: I was testing you :) Good catch!

core/utils/bits.h
```diff
@@ -163,6 +165,38 @@ static inline void MaskBytes(uint8_t *buf, uint8_t const *mask,
   }
 }
 
+// Dangerous Helper! Use SetBitsHigh<T>() and SetBitsLow<T>() instead.
+template <typename T, typename = std::enable_if<std::is_integral<T>::value>>
+static inline T _SetBitsHigh(size_t n) {
+  return (T{1} << n) - 1;
+}
+
+// TODO(melvinw): it would be nice if MaskTopN() supported blobs.
```
chris3torek: nit: this TODO line still has the old name

melvinw:
Fixed, squashed and rebased

sangjinhan:
tumblr_m4okzvllic1qd477zo1_1280

PR #693, opened 2017-10-20
Module benchmark

shinae-woo:
This PR 1) ports module benchmark code from .cc (using google benchmark) to .bess 2) adds some benchmarks to see module split/merge inside a single task. Some reasons behind this refactoring are
 - Using .bess shows more realistic performance (e.g., using mempool)
 - Packets on the stack should not be used in other modules such as non-recursive inter-module scheduler.
 - Add more comprehensive benchmark (split, merge, asymmetric pipeline, etc)


bessctl/conf/perftest/chain.bess
```diff
@@ -0,0 +1,71 @@
...
+    last -> Sink()
+
+def measure(duration):
+    tcs = bess.list_tcs().classes_status
+    root_tc = None
+    for tc in tcs:
+        if not getattr(tc, 'parent'):
+            root_tc = getattr(tc, 'class')
+    if not root_tc:
+        raise Exception('Fail to find root tc')
```
chris3torek: Should we turn this little chunk of code into a method available in unit tests and an exported function in old style tests?  There's one copy in `bessctl/test_utils.py` and now this here and below...

Or, perhaps `measure` should be the exported function.

bessctl/conf/perftest/complex_merge.bess
```diff
@@ -0,0 +1,103 @@
...
+psize = int($BESS_PKT_SIZE!'60')
+print('Using packet size %d (envvar "BESS_PKT_SIZE")' % psize)
+
+def create_pipeline(child_cnt, tree_depth):
+    splits = {}
+    merges = {}
+    src = Source(pkt_size=psize)
+    splits[0] = RandomSplit(gates=range(child_cnt))
+
+    src-> splits[0]
```
chris3torek: trivial formatting: space before `->`

bessctl/conf/perftest/complex_merge.bess
```diff
@@ -0,0 +1,103 @@
...
+    # Start split
+    pidx = 0;
+    cidx = 1;
+    for d in range(tree_depth-1):
+        for p in range(pow(child_cnt, d)): # number of parents at depth d
+            for c in range(child_cnt): # for each parent connect child
+                splits[cidx] = RandomSplit(gates=range(child_cnt))
+                splits[pidx]:c -> splits[cidx]
+                cidx = cidx + 1
+            pidx = pidx + 1
```
chris3torek: trivial: these should probably all use `+= 1`.

I thought about suggesting `for pidx, p in enumerate(range(child_cnt ** d))` but in the end I think it might be clearer this way. It's possible to use `enumerate` on the child index numbers as well, with `start=`, but that seems to get quite obscure.

bessctl/conf/perftest/complex_merge.bess
```diff
@@ -0,0 +1,103 @@
...
+    for p in range(pow(child_cnt, tree_depth-1)): # number of parents at depth d
+        for c in range(child_cnt): # for each parent connect child
+            merges[cidx] = Bypass()
+            splits[pidx]:c -> merges[cidx]
+            cidx = cidx + 1
+        pidx = pidx + 1
+
+    # Start merging (reversely construct)
+    cidx, pidx = pidx - 1, cidx - 1
+    all = pow(child_cnt, tree_depth+1) -2
```
chris3torek: trivial: `all` is not used here, we can delete it (which is good for several reasons 😺 )

bessctl/conf/perftest/complex_merge.bess
```diff
@@ -0,0 +1,103 @@
...
+    while cidx >= 0:
+        merges[cidx] = Bypass()
+        for c in range(child_cnt):
+            merges[pidx] -> merges[cidx]
+            pidx = pidx - 1
```
chris3torek: (as above, can use `-= 1`)

bessctl/conf/perftest/complex_split.bess
```diff
@@ -0,0 +1,86 @@
...
+
+psize = int($BESS_PKT_SIZE!'60')
+print('Using packet size %d (envvar "BESS_PKT_SIZE")' % psize)
+
+def create_pipeline(child_cnt, tree_depth):
+    splits = []
+    src = Source(pkt_size=psize)
+    splits.insert(0, RandomSplit(gates=range(child_cnt)))
+
+    src-> splits[0]
```
chris3torek: trivial formatting: space before `->`

bessctl/conf/perftest/complex_split.bess
```diff
@@ -0,0 +1,86 @@
...
+
+    pidx = 0;
+    cidx = 1;
+    for d in range(tree_depth-1):
+        for p in range(pow(child_cnt, d)): # number of parents at depth d
+            for c in range(child_cnt): # for each parent connect child
+                splits.insert(cidx, RandomSplit(gates=range(child_cnt)))
+                splits[pidx]:c -> splits[cidx]
+                cidx = cidx + 1
+            pidx = pidx + 1
```
chris3torek: (same `+=` suggestion, here and below)

bessctl/conf/perftest/merge_asym.bess
```diff
@@ -0,0 +1,68 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+import time
+
+psize = int($BESS_PKT_SIZE!'60')
+print('Using packet size %d (envvar "BESS_PKT_SIZE")' % psize)
+
+def create_pipeline():
+    src::Source(pkt_size=psize) -> rr::RoundRobin(gates=[0, 1, 2, 3], mode='packet')
```
sangjinhan: Just checking: did you intend `RoundRobin` here? (all other scripts seem to use RandomSplit)

sangjinhan:
Not directly related to this PR, but my two cents: the current setup may not signify the overhead of the scheduler (or framework in general) enough since the Bypass module is oblivious to the packet/batch size. Maybe it's overly simplistic.

It would be interesting to see the impact of batch merging, thus amortized per-batch cost as well. For example, modules declared like this `Bypass(cycles_per_batch=100, cycles_per_packet=10, cycles_per_byte=0.1)` burns cycles accordingly.

bessctl/conf/perftest/complex_merge.bess
shinae-woo: Thank you for such detailed review!!

bessctl/conf/perftest/chain.bess
shinae-woo: Right, those are quite redundant.

shinae-woo:
@sangjinhan I agree! I am pretty excited to see the actual benefits, too.

sangjinhan:
Merged, thanks.
3ee83d1b4e08841d47cd492e5e9d4f50

PR #696, opened 2017-10-21
worker_split: allow custom worker->gate mappings

melvinw:
It would be nice if WorkerSplit supported mappings other than f(x) = x.
This would be particularly useful in cases when BESS is configured with
a large number of workers and a module graph with a WorkerSplit module
that has only a few downstream modules. With the current implementation
this would require to a lot of edges from the WorkerSplit to the same
place.  For example, you might want to steer packets from workers on the
same CPU socket towrad the same destination.

To make these sorts of mappings a bit more concise, this commit changes
WorkerSplit to operate in one of two modes:

1) Packets from worker `x` are mapped to output gate `x`. This is the
   default mode.

2) When the `worker_gates` field is set, packets from a worker `x` are
   mapped to `worker_gates[x]`. In this mode, packet batches from
   workers not mapped to an output gate will be dropped.

The default mode(1) is about 2% slower (141Mpps vs 138Mpps) on my 2.1
GHz Xeon E5-2620 v4 after this change.

### DIFF
--- a/bessctl/module_tests/worker_split.py
+++ b/bessctl/module_tests/worker_split.py
@@ -27,12 +27,13 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
+import multiprocessing as mp
 from test_utils import *
 
 
 class BessWorkerSplitTest(BessModuleTestCase):
 
-    def test_worker_split(self):
+    def test_worker_split_default(self):
         NUM_WORKERS = 2
 
         for wid in range(NUM_WORKERS):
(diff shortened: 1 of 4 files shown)


bessctl/module_tests/worker_split.py
```diff
@@ -62,6 +62,52 @@ def test_worker_split(self):
 
             bess.reset_all()
 
+    def test_worker_split_mask_mode(self):
+        NUM_WORKERS = 6
+
+        masks = dict()
+        # map even workers to gate 0
+        masks[0] = reduce(lambda x,y: x|y, [0x1 << 2*i for i in range(NUM_WORKERS/2)])
```
chris3torek: Need `NUM_WORKERS // 2` to force integer divide in python3:
```
>>> range(6/2)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'float' object cannot be interpreted as an integer
```

bessctl/module_tests/worker_split.py
```diff
@@ -62,6 +62,52 @@ def test_worker_split(self):
 
             bess.reset_all()
 
+    def test_worker_split_mask_mode(self):
+        NUM_WORKERS = 6
```
chris3torek: Maybe note that this must be even? (seems kind of obvious though)

core/modules/worker_split.cc
```diff
@@ -30,16 +30,58 @@
...
+  }
+
+  for (auto it : arg.worker_masks()) {
+    gate_idx_t ogate = it.first;
+    int wid = 0;
+    for (uint64_t mask = it.second; mask; mask >>= 1, wid++) {
+      if (mask & 0x1ull) {
+        if (gates_[wid] >= 0) {
+          return CommandFailure(
+              EINVAL, "worker %d is already mapped to gate %" PRIu16 "", wid,
```
chris3torek: (What's the empty `""` for here?)

bessctl/module_tests/worker_split.py
```diff
@@ -63,7 +63,7 @@ def test_worker_split_default_mode(self):
             bess.reset_all()
 
     def test_worker_split_mask_mode(self):
-        NUM_WORKERS = 6
+        NUM_WORKERS = 2
```
chris3torek: (Hopefully my individual comment on earlier commit doesn't vanish...)

I presume this is needed for Travis testing because of just 2 cores. Maybe we could limit it to ncpu rounded to even? (ncpu available by running `nproc` or importing `multiprocessing` and calling `multiprocessing.cpu_count()`)

bessctl/module_tests/worker_split.py
```diff
@@ -62,6 +62,52 @@ def test_worker_split(self):
...
             bess.reset_all()
 
+    def test_worker_split_mask_mode(self):
+        NUM_WORKERS = 6
+
+        masks = dict()
+        # map even workers to gate 0
+        masks[0] = reduce(lambda x,y: x|y, [0x1 << 2*i for i in range(NUM_WORKERS/2)])
+        # map odd workers to gate 1
+        masks[1] = reduce(lambda x,y: x|y, [0x1 << (2*i + 1) for i in range(NUM_WORKERS/2)])
```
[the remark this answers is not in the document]
melvinw: Good point. Updated

bessctl/module_tests/worker_split.py
melvinw: Great catch! Fixed

bessctl/module_tests/worker_split.py
melvinw: No enforced with `multiprocessing.cpu_count() & 0xFE`

core/modules/worker_split.cc
melvinw: Serves no purpose at all :) Removed

sangjinhan:
One high-level comment: since we don't allow one worker to be mapped to multiple gates, it would be more straightforward to use a direct map (e.g., `gates={4:99, 7:99}` for your example) than using bitmasks?

melvinw:
I chose bitmasks because they seem more compact. Your way is much easier for humans to read and less error prone, though. Will adopt

sangjinhan:
tumblr_m4nb9ocfi31qa19lvo1_500

PR #697, opened 2017-10-21
minor resume hook fixes

melvinw:
I realized I forgot to push a bunch of requested changes before #661 was merged.

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -506,7 +506,7 @@ class BESSControlImpl final : public BESSControl::Service {
       attach_orphans();
     }
 
-    run_global_resume_hooks();
+    bess::run_global_resume_hooks();
 
     LOG(INFO) << "*** Resuming ***";
     resume_all_workers();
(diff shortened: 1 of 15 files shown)


core/bessctl.cc
```diff
@@ -524,7 +524,14 @@ class BESSControlImpl final : public BESSControl::Service {
   Status ResetWorkers(ServerContext*, const EmptyRequest*,
                       EmptyResponse*) override {
     WorkerPauser wp;
-    bess::global_resume_hooks.clear();
+    auto& hooks = bess::global_resume_hooks;
```
sangjinhan: It seems that an implicit assumption here is that the lifetime of a registered global hook is ended when `ResetWorkers` is invoked. While this is a reasonable choice (unless we want to add another API), one concern is that now `ResetWorkers` and destroying all modules individually have different behaviors. You may want to move this code block to the end of `destroy_worker`, after checking if `num_workers` has become zero.

melvinw: Good point. The process should be consistent across all forms of worker deletion. Moved.

sangjinhan:
tumblr_m4okzorvsy1qd477zo1_1280

PR #700, opened 2017-10-25
add state save/restore code for wildcard match module

chris3torek:
This series of three commits:

1. adds a Python test for wildcard match
2. adds the same kind of state save/load for wildcard match as for exact match
3. removes the old (wildcard-match specific) state-grabber

Commit 3 is entirely optional, and if it is to be done, we can simplify the new code by removing the leftover template-ized internal functions.  I figured I'd put it up in this form first, for reviewing.

### DIFF
--- a/bessctl/module_tests/wildcard_match.py
+++ b/bessctl/module_tests/wildcard_match.py
@@ -1,0 +1,181 @@
+# Copyright (c) 2017, Nefeli Networks, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# * Neither the names of the copyright holders nor the names of their
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import socket
+import sys
+from test_utils import *
+from pybess import protobuf_to_dict as pb_conv
+
+
+def vstring(*values):
+    return [{'value_bin': bytes(bytearray(i))} for i in values]
+
+
+class BessWildcardMatchTest(BessModuleTestCase):
+
+    def test_run_wildcardmatch(self):
+        # basically the same mask/match values as for ExactMatch
+        # (not the most thorough test)
+        wm = WildcardMatch(fields=[{'offset': 23, 'num_bytes': 1},
+                                   {'offset': 2, 'num_bytes': 2},
+                                   {'offset': 29, 'num_bytes': 1}])
+        mask_list = vstring([0xff], [0xff, 0xff], [0xff])
+        wm.add(gate=0, priority=0, masks=mask_list,
+               values=vstring([0xff], [0x23, 0xba], [0x34]))
+        wm.add(gate=1, priority=0, masks=mask_list,
+               values=vstring([0xff], [0x34, 0xaa], [0x12]))
+        wm.add(gate=2, priority=0, masks=mask_list,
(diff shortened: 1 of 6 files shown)


core/modules/wildcard_match.cc
```diff
@@ -374,20 +383,21 @@ CommandResponse WildcardMatch::CommandDelete(
...
 }
 
-CommandResponse WildcardMatch::CommandGetRules(const bess::pb::EmptyArg &) {
-  bess::pb::WildcardMatchCommandGetRulesResponse resp;
-  resp.set_default_gate(default_gate_);
-
+// This is left over from when there was a separate
+// WildcardMatchCommandGetRules; we could just inline this now.
+template <typename T>
+void WildcardMatch::GetFields(T &resp) {
```
sangjinhan: Not that I like this convention, but for consistency: the Google style guide states that in the argument list mutable objects should be given as a pointer, and non-const references are not allowed [(link)](http://google.github.io/styleguide/cppguide.html#Reference_Arguments).

core/modules/wildcard_match.cc
```diff
@@ -397,30 +407,33 @@ CommandResponse WildcardMatch::CommandGetRules(const bess::pb::EmptyArg &) {
     }
     f->set_num_bytes(field.size);
   }
+}
+
+// Sets up default gate, adds a rule for each (gate, priority) entry,
+// and then calls put_field on each (field_mask, field_match, size).
+template <typename rule_t, typename T, typename Func>
+void WildcardMatch::GetRules(T &resp, Func put_field) {
```
sangjinhan: can this be a const function?

core/modules/wildcard_match.cc
```diff
@@ -397,30 +407,33 @@ CommandResponse WildcardMatch::CommandGetRules(const bess::pb::EmptyArg &) {
...
       for (auto &field : fields_) {
-        uint64_t data = 0;
-        bess::utils::bin_to_uint64(&data, entry_data + field.pos, field.size,
-                                   true);
-        rule->add_values(data);
-        uint64_t mask_data = 0;
-        bess::utils::bin_to_uint64(&mask_data, entry_mask + field.pos,
-                                   field.size, true);
-        rule->add_masks(mask_data);
+        put_field(rule, entry_data + field.pos, entry_mask + field.pos,
```
sangjinhan: Given that there is only one caller of `GetRules` yet, can we inline the function `put_field` here for now? Template parameter `Func` doesn't seem to do much here.

core/modules/wildcard_match.h
```diff
@@ -166,10 +168,21 @@ class WildcardMatch final : public Module {
...
   CommandResponse ExtractKeyMask(const T &arg, wm_hkey_t *key, wm_hkey_t *mask);
 
+  template <typename T>
+  CommandResponse Add(const T &arg);
+
+  template <typename T>
+  void GetFields(T &arg);
+
+  template <typename rule_t, typename T, typename Func>
+  void GetRules(T &, Func put_rule);
```
sangjinhan: In the .cc file the second argument is named `put_field`.

core/modules/wildcard_match.cc
```diff
@@ -312,8 +316,8 @@ int WildcardMatch::DelEntry(int idx, wm_hkey_t *key) {
   return 0;
 }
 
-CommandResponse WildcardMatch::CommandAdd(
-    const bess::pb::WildcardMatchCommandAddArg &arg) {
+template <typename T>
+CommandResponse WildcardMatch::Add(const T &arg) {
```
sangjinhan: Why is it templatized? `T` seems to be the same type, for all the callers of this function.

chris3torek: This is because of the three part nature of the original commit series...

> we can simplify the new code by removing the leftover template-ized internal functions

core/modules/wildcard_match.cc
chris3torek: Ah, right, I'm still slowly absorbing the Google style guide.  Will fix.

core/modules/wildcard_match.h
chris3torek: Left over from evolution.  If we're really taking out the old get-rules command, can de-templatize and just inline the whole thing anyway.

core/modules/wildcard_match.cc
chris3torek: I think so (or just all moved inline when we ditch the extra templates...).

bessctl/module_tests/wildcard_match.py
```diff
@@ -0,0 +1,190 @@
...
+        return [{'value_bin': bytes(i)} for i in values]
+else:
+    def vstring(*values):
+        ret = []
+        for i in values:
+            s = []
+            for j in i:
+                s.append(chr(j))
+            ret.append({'value_bin': bytes(''.join(s))})
+        return ret
```
chris3torek: This is clumsy btw, I've replaced it with:

    def vstring(*values):
        return [{'value_bin': ''.join(chr(j) for j in i)} for i in values]

(it's still annoying to have to have py2k vs py3k variants).

chris3torek:
I rebased, squashed, de-template-ized, and re-arranged everything a bit to make the diffs as nice as possible.

bessctl/module_tests/wildcard_match.py
```diff
@@ -0,0 +1,184 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+
+import socket
+import sys
+from test_utils import *
+from pybess import protobuf_to_dict as pb_conv
+
+if sys.version_info[0] >= 3:
+    def vstring(*values):
+        return [{'value_bin': bytes(i)} for i in values]
```
ddiproietto: I'm if it's better, but if we use `bytes(bytearray(i))` instead of `bytes(i)`, this should also work on python2.

Source:

https://stackoverflow.com/a/29182294

chris3torek: Ah, that _is_ nicer (no more sys.version testing) - I'll re-revamp.

core/utils/endian.h
```diff
@@ -173,10 +173,6 @@ static_assert(sizeof(be64_t) == 8, "be64_t is not 8 bytes");
 // Copy data from val to *ptr. Set "big_endian" to store in big endian
 bool uint64_to_bin(void *ptr, uint64_t val, size_t size, bool big_endian);
 
-// Copy data from *ptr to *pval. Set "big_endian" if *ptr has big endian data
-bool bin_to_uint64(uint64_t *pval, const void *ptr, size_t size,
-                   bool big_endian);
-
```
shinae-woo: Just of curiosity, is there any reason to delete this util function? Guessing the functions are used by nowhere now.

chris3torek: That was the reason (this was the last user). I wasn't sure if deleting it was appropriate, but nobody objected to the previous three-commit version deleting it. 😺 I can leave it in, or we can always resurrect it from the Git history.

chris3torek:
(re-rebased/squashed with the comment typo fixed and @ddiproietto's improved python bit)

ddiproietto:
gatto

PR #701, opened 2017-10-25
port_out: properly initialize worker-queue mapping

melvinw:


core/modules/port_out.cc
```diff
@@ -135,7 +139,7 @@ int PortOut::OnEvent(bess::Event e) {
   }
 
   // Assign remaining queues to any newly attached workers.
-  for (size_t i = 0; i < actives.size(); i++) {
+  for (size_t i = 0; i < Worker::kMaxWorkers; i++) {
```
chris3torek: I had to check that `actives` actually has `kMaxWorkers` accessible elements (it does, it's explicitly allocated at that size)...

PR #705, opened 2017-10-27
some test framework fixes/improvements

melvinw:


bessctl/conf/testing/run_module_tests.bess
```diff
@@ -42,6 +42,10 @@ from pybess import protobuf_to_dict as pb_conv
 # Env variable used to run a single test
 test_name = $TEST_NAME!'*'
 
+default_path = os.path.join(os.path.dirname(os.path.realpath(sys.argv[0])),
+                            'conf/testing/module_tests')
```
chris3torek: Shouldn't this just be `'module_tests`?  I forget precisely how this worked before all the changes but the current code (below) just adds `module_tests`...

bessctl/conf/testing/run_module_tests.bess
```diff
@@ -158,14 +162,11 @@ def monitor_task(module, wid):
 def run_tests():
     DID_TESTS_FAIL = False
 
-    path = os.path.dirname(os.path.realpath(sys.argv[0]))
-
     # Note: os.path.join will discard the leading parts if test
     # begins with '/', so callers should provide an absolute path
     # if and only if they want to override the internal BESS
     # path prefixing.
```
chris3torek: This comment has become out of date.  It referred to this behavior:

    >>> os.path.join('/abs/path/to/dir', 'b')
    '/abs/path/to/dir/b'
    >>> os.path.join('/abs/path/to/dir', '/different/abs/b')
    '/different/abs/b'

 which reminds me of how I think I had changed it, I think I just let an abs path in `$TEST_PATH` override the whole thing, and you had to include the `.../module_tests` part.

Anyway, not really critical how it's done, but I think the first part is a bug.

bessctl/conf/testing/run_module_tests.bess
melvinw: Ah, the tests moved. I just copied this from the original impl. Will update

melvinw: Fixed

shinae-woo:
Thank you 👍 
image

PR #706, opened 2017-10-31
Make task being pointer than reference for LeafTC

shinae-woo:
- While doing refactoring for intra-task scheduling, the task can have internally malloc()ed data fields, so shallow copy should not work.
 - Using pointer would more make sense when we will extend the Task functionality (e.g., Timer).


sangjinhan:
tumblr_kony0f37tp1qzex9io1_1280

Issue #703, opened 2017-10-27
Creating a pipeline with lots of Queue()s is slow


nemethf:
Consider the following bess script (which is a simplified version of a real use-case.)  
```python
case = int($case!'1')

split = Split(attribute='tun_id', size=4)
ip_lookup = IPLookup()
buf = Buffer()
Source() -> split

if case == 1:
  buf -> ip_lookup

for i in range(1, 2000):
  split:i -> Queue() -> buf

for i in range(1, 250):
  ip_lookup:i -> Bypass() -> Bypass() -> Sink()

if case == 2:
  buf -> ip_lookup
```
The configuration time largely depends on how how we connect the modules: 
```
~/bess/bessctl$ time ./bessctl daemon start -- run add_edge case=1 -- daemon stop

real    3m17.908s
user    0m4.952s
sys     0m0.516s
~/bess/bessctl$ time ./bessctl daemon start -- run add_edge case=2 -- daemon stop

real    0m19.026s
user    0m4.424s
sys     0m0.532s
```

I think the slowness comes from the fact that ModuleGraph::AddEdge updates the whole task graph, but it would be enough to update only those parts of the graph that might be affected by the new edge.   With this optimization the running times decrease in both cases :
```
~/bess/bessctl$ time ./bessctl daemon start -- run add_edge case=1 -- daemon stop

real    0m9.865s
user    0m2.924s
sys     0m0.268s
~/bess/bessctl$ time ./bessctl daemon start -- run add_edge case=2 -- daemon stop

real    0m9.801s
user    0m2.900s
sys     0m0.316s
```

If my analysis is correct I can send the following patch as a pull request.  However, I think further optimizations are possible in this area.  Thanks.
 
```
diff --git a/core/module_graph.cc b/core/module_graph.cc
index d760cbf..9ce18bf 100644
--- a/core/module_graph.cc
+++ b/core/module_graph.cc
@@ -88,6 +88,21 @@ bool ModuleGraph::UpdateTaskGraph() {
   return true;
 }
 
+bool ModuleGraph::UpdateTaskSubgraph(const std::string &node_name) {
+  auto node_it = all_modules_.find(node_name);
+  if (node_it == all_modules_.end()) {
+    return false;
+  }
+  Module *node = node_it->second;
+  for (Module *it : node->parent_tasks_) {
+    std::unordered_set<std::string> visited;
+    if (!FindNextTask(it->name(), it->name(), &visited)) {
+      return false;
+    }
+  }
[...]

PR #707, opened 2017-10-31
Task graph

shinae-woo:
This PR partly address the issue on #703, issued by @nemethf.

Now the time to start becomes about 6s from 3min 20s.
$ `time bessctl daemon start -- run add_edge`
real        0m6.298s
user        0m1.436s
sys        0m0.388s

This includes many changes in source code including
1. Remove module_graph_ data structure: We don't need to manage multiple module graphs as makes maintenance harder, and higher risks for potential bugs. 
2. Follow up from 1, all the module graphs changes should be maintained at one place - module graphs. No direct module connection/disconnection is not allowed any longer. (and now those are private functions of modules)
3. Makes the generating a task graph as a default resume hook. It does not update the task graph per-connection updates but updates at once right before resume.
4. The resume hooks for updating task graph only regenerate task graphs if there are any changes on module connections. So just resume for updating module parameters or measure some statistics will not trigger task graphs update.

More optimizations would be required. I leave these as a future PR or issues to reduce the overhead of single big PR. (current PR is huge enough)
- A single module connection changes will trigger the whole task graph regeneration. As @nemethf suggested, subgraph update only regarding changes would improve the performance.
- Updating workers per task or checking constraints can also be put as a resume hook. At least, checking constraints and updating workers per task should be managed to the same control flow wherever it is. Now one is contolled by `pybess` and while the other is controlled by `bessd`.
- It would be better if we limit the role of `bessctl.cc` not to directly changes the internal of bessd (such as TC), and migrate logics into the separate classes so that a single file takes the responsibility of regarding all related changes. (TC tree is now managed in `bessctl.cc, module.cc, worker.cc` now)

### DIFF
--- a/core/gate.h
+++ b/core/gate.h
@@ -165,29 +165,7 @@ class Gate {
   DISALLOW_COPY_AND_ASSIGN(Gate);
 };
 
-class IGate;
-
-// A class for output gate. It connects to an input gate of the next module.
-class OGate : public Gate {
- public:
-  OGate(Module *m, gate_idx_t idx, Module *next)
-      : Gate(m, idx), next_(next), igate_(), igate_idx_() {}
-
-  void set_igate(IGate *ig) { igate_ = ig; }
-
-  IGate *igate() const { return igate_; }
-  Module *next() const { return next_; }
-
-  void set_igate_idx(gate_idx_t idx) { igate_idx_ = idx; }
-  gate_idx_t igate_idx() const { return igate_idx_; }
-
- private:
-  Module *next_;          // next module connected with
-  IGate *igate_;          // next igate connected with
-  gate_idx_t igate_idx_;  // cache for igate->gate_idx
-
-  DISALLOW_COPY_AND_ASSIGN(OGate);
-};
+class OGate;
 
 // A class for input gate
 class IGate : public Gate {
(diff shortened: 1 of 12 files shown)


barath:
LGTM; Panda should also take a look.

core/gate.h
```diff
@@ -165,20 +165,39 @@ class Gate {
   DISALLOW_COPY_AND_ASSIGN(Gate);
 };
 
-class IGate;
+class OGate;
+
+// A class for input gate
+class IGate : public Gate {
```
apanda: I assume you had to move this up so that `OGate::SetIgate` could use `gate_idx()` in line 195? I wonder if it might be worthwhile moving the implementations for IGate and OGate methods to the cc file, since neither is templated or shows any other reason why doing so would be disallowed? This might simplify future changes, but it is entirely optional since I am not sure of the benefits.

shinae-woo: I agree to the point.  I changed the order of two classes definition because I'm lazy and want to leave a minimal footprint. I will do, it's time to burn my hot chocolate. 🍫

apanda:
Thank you! Will merge once Travis is done.

apanda:
Thanks
image

PR #709, opened 2017-11-01
Burn cycles for performance tests

shinae-woo:
This PR is following up for #693, module benchmarks.

Now, `Bypass` burns CPU cycles for per-batch, per-packet, and per-byte. In this way, we can see how much per-batch cost can be amortized by merging back batches after splitting.

### DIFF
--- a/core/modules/bypass.cc
+++ b/core/modules/bypass.cc
@@ -1,4 +1,4 @@
-// Copyright (c) 2014-2016, The Regents of the University of California.
+// Copyright (c) 2014-2017, The Regents of the University of California.
 // Copyright (c) 2016-2017, Nefeli Networks, Inc.
 // All rights reserved.
 //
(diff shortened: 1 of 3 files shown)


protobuf/module_msg.proto
```diff
@@ -450,12 +450,17 @@ message BufferArg {
...
+ * It burns cpu cycles per_batch, per_packet, and per-bytes.
+ * Bypass is useful primarily for testing and performance evaluation.
  *
  * __Input Gates__: 1
  * __Output Gates__: 1
  */
 message BypassArg {
+  uint32 cycles_per_batch = 1;
+  uint32 cycles_per_packet = 2;
+  double cycles_per_byte = 3;
```
melvinw: Why `double` here instead of `uint32` like the other members?

core/modules/bypass.cc
```diff
@@ -30,7 +30,37 @@
...
+  return CommandSuccess();
+}
+
 void Bypass::ProcessBatch(bess::PacketBatch *batch) {
+  uint32_t cur_tsc = ctx.current_tsc();
+  uint32_t cycles = cycles_per_batch_ + cycles_per_packet_ * batch->cnt();
+
+  if (cycles_per_byte_) {
+    uint64_t total_bytes = 0;
+    for (int i = 0; i < batch->cnt(); i++) {
```
melvinw: Shouldn't we subtract out the cycles spent doing this calculation? The extra call to `rdtsc()` might not make sense for low values of `cycles_per_*`, but for larger values it think it would come out in the wash.

protobuf/module_msg.proto
shinae-woo: Just following the suggestions from the issue discussion. uint32 also fine.

core/modules/bypass.cc
shinae-woo: http://www.agner.org/optimize/instruction_tables.pdf

Although RDTSC is one of the expensive ones, in my opinion, it is not comparable with packet processing. Default Bypass will not call it as well. The overall cycling measurement is anyhow including all cycles computing those.

melvinw: Ah, you're right! I ignored that `cur_tsc` gets set before this -.-

sangjinhan:
cat?!

melvinw:
AH!

melvinw:
Thanks! :)

cat

melvinw:
Have a second cat since I made you wait!

cat3

PR #710, opened 2017-11-02
nat: add get/set arg/state

chris3torek:
Here the set of IP addresses being NAT-ed is all arguments,
rather than all state.  So we don't (yet) define a NATConfig.
This may change if necessary.

---

This PR is mainly for @apanda to look at, as it doesn't have the "move args to config" that @ddiproietto and I talked about.  But we can use it in this form if we decide that's not important after all.

### DIFF
--- a/bessctl/module_tests/nat.py
+++ b/bessctl/module_tests/nat.py
@@ -102,6 +102,18 @@ class BessNatTest(BessModuleTestCase):
         nat = NAT(ext_addrs=['192.168.1.1'])
         self._test_l4(nat, scapy.ICMP(), '192.168.1.1')
 
+    def test_nat_selfconfig(self):
+        # Send initial conf unsorted, see that it comes back sorted
+        # (note that this is a bit different from other modules
+        # where argument order often matters).
+        iconf = {'ext_addrs': ['0.1.2.3', '192.168.1.1', '3.2.1.0']}
+        nat = NAT(**iconf)
+        iconf['ext_addrs'].sort(key=lambda s: socket.inet_aton(s))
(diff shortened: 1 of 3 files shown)


bessctl/module_tests/nat.py
```diff
@@ -102,6 +102,18 @@ def test_nat_icmp(self):
...
+        # Send initial conf unsorted, see that it comes back sorted
+        # (note that this is a bit different from other modules
+        # where argument order often matters).
+        iconf = {'ext_addrs': ['0.1.2.3', '192.168.1.1', '3.2.1.0']}
+        nat = NAT(**iconf)
+        iconf['ext_addrs'].sort(key=lambda s: socket.inet_aton(s))
+        arg = pb_conv.protobuf_to_dict(nat.get_initial_arg())
+        expect_config = {}
+        cur_config = pb_conv.protobuf_to_dict(nat.get_runtime_config())
+        assert arg == iconf and cur_config == expect_config
```
sangjinhan: Using `self.assertEquals` might be helpful here to pinpoint problems more easily.

chris3torek: That would be an improvement, but I have a plan to put in a generic arg+config tester, where the test builds a module and returns its expected config, then calls `self.assertConfiguration(iconf, expected)`.  The test framework can show what's wrong, if something is wrong, and then the test framework can create a new module and call its `set_runtime_config` and verify that the new module matches the old module as well (this will test all the `set_runtime_config`s that, at the moment, are not being unit tested at all...).

sangjinhan:
qgib8teriqphq7l484g6q2e9o1_500

PR #711, opened 2017-11-02
Make NAT configuration more flexible

apanda:
This PR allows users to decide what ports the NAT should use at any given time. This not ready until merging with PR #710

### DIFF
--- a/bessctl/conf/perftest/nat.bess
+++ b/bessctl/conf/perftest/nat.bess
@@ -41,11 +41,15 @@ ip = scapy.IP(src='10.0.0.1', dst='192.168.1.1')
 udp = scapy.UDP(sport=10001, dport=10002)
 payload = 'helloworld'
 pkt_bytes = bytes(eth/ip/udp/payload)
+nat_config = [{'ext_addr': '1.1.1.1', \
+    'port_ranges': [{'begin': 0, 'end': 65535, 'usable': True}]}, \
+    {'ext_addr': '1.1.1.2', \
+    'port_ranges': [{'begin': 0, 'end': 65535, 'usable': True}]}]
 
 Source() \
     -> Rewrite(templates=[pkt_bytes]) \
     -> RandomUpdate(fields=[{'offset': 26, 'size': 4, 'min': 0x0a000001, 'max': 0x0a000001 + num_flows - 1}]) \
-    -> nat::NAT(ext_addrs=['1.1.1.1', '1.1.1.2'])
+    -> nat::NAT(ext_addrs=nat_config)
 
 if bidirectional:
     nat -> MACSwap() -> IPSwap() -> 1:nat:1 -> Sink()
--- a/bessctl/conf/perftest/nat.bess
+++ b/bessctl/conf/perftest/nat.bess
@@ -42,9 +42,9 @@ udp = scapy.UDP(sport=10001, dport=10002)
 payload = 'helloworld'
 pkt_bytes = bytes(eth/ip/udp/payload)
 nat_config = [{'ext_addr': '1.1.1.1', \
-    'port_ranges': [{'begin': 0, 'end': 65535, 'usable': True}]}, \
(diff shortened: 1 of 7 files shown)


core/modules/nat.cc
```diff
@@ -122,74 +136,88 @@ NAT::HashTable::Entry *NAT::CreateNewEntry(const Endpoint &src_internal,
...
-      // ignore port number 0
-      return nullptr;
-    } else if (src_internal.port & ~be16_t(1023)) {
-      min = 1024;
-      range = 65535 - min + 1;
-    } else {
-      // Privileged ports are mapped to privileged ports (rfc4787 REQ-5-a)
-      min = 1;
-      range = 1023;
+  for (auto port_range : port_ranges_[ext_addr_index]) {
```
sangjinhan: const auto &

core/modules/nat.h
```diff
@@ -118,6 +118,18 @@ struct NatEntry {
   uint64_t last_refresh;  // in nanoseconds (ctx.current_ns)
 };
 
+// Port ranges are used to scale out the NAT.
+struct PortRange {
+  // Start of port range.
+  uint16_t begin;
+  // End of port range.
+  uint16_t end;
```
sangjinhan: Clarification: is this "end" inclusive?

core/modules/nat.cc
```diff
@@ -54,14 +54,28 @@ using bess::utils::UpdateChecksumWithIncrement;
...
+
     bool ret = bess::utils::ParseIpv4Address(ext_addr, &addr);
     if (!ret) {
       return CommandFailure(EINVAL, "invalid IP address %s", ext_addr.c_str());
     }
 
     ext_addrs_.push_back(addr);
+    // Add a port range list
+    std::list<PortRange> port_list;
+    for (const auto &range : address_range.port_ranges()) {
```
sangjinhan: it would be nice to do simple sanity check here, (e.g., `start <= end`, 16-bit port numbers, etc.)

bessctl/conf/perftest/nat.bess
```diff
@@ -41,11 +41,15 @@ ip = scapy.IP(src='10.0.0.1', dst='192.168.1.1')
 udp = scapy.UDP(sport=10001, dport=10002)
 payload = 'helloworld'
 pkt_bytes = bytes(eth/ip/udp/payload)
+nat_config = [{'ext_addr': '1.1.1.1', \
+    'port_ranges': [{'begin': 0, 'end': 65535, 'usable': True}]}, \
```
sangjinhan: Perhaps it's my personal fetish, but can we replace `usable` with `suspended` (or something like that)? It will reduce verbosity by skipping the argument, since the Protobuf default value `False` can be used here.

core/modules/nat.h
```diff
@@ -118,6 +118,18 @@ struct NatEntry {
...
+  // Is range actively in use, i.e., have we given out ports in this range.
+  bool in_use;
```
sangjinhan: If this field is currently not being updated/used, can we remove it for now?

bessctl/conf/perftest/nat.bess
sangjinhan: if `port_ranges` is not specified (thus an empty list), maybe we can assume all ports can be used, rather than none? I see this alternative is less "pure" but makes it much less verbose too.

core/modules/nat.h
```diff
@@ -155,6 +167,10 @@ class NAT final : public Module {
 
   std::vector<be32_t> ext_addrs_;
 
+  // Port ranges available for each address. The first index is the same as the
+  // ext_addrs_ range.
+  std::vector<std::list<PortRange>> port_ranges_;
```
sangjinhan: How about vector in place of list, as per Effective STL (vectors should be used by default for read-mostly sequences)?

sangjinhan:
Merged, thanks!
jjkybd3nscvuclrpjipekbxa_500

PR #715, opened 2017-11-08
URL filter updates

chris3torek:
Two part commit:

 * first one adds a cheesy trie traversal into vector so that we can do state-dump
 * second one adds state save/load to url_filter, adds tests, fixes bugs so that tests pass

### DIFF
--- a/bessctl/module_tests/url_filter.py
+++ b/bessctl/module_tests/url_filter.py
@@ -1,0 +1,132 @@
+# Copyright (c) 2017, Nefeli Networks, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# * Neither the names of the copyright holders nor the names of their
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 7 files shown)


chris3torek:
The address and leak sanitizer is suggesting that the http parser is overrunning buffers.  I haven't dug into that code yet...

chris3torek:
ASAN code seems to be complaining about the 16-byte read of variable `ranges2`:

    static const char ALIGNED(16) ranges2[] = "\000\040\177\177";
    ... buf = findchar_fast(buf, buf_end, ranges2, sizeof(ranges2) - 1, &found2); ...
    [findchar_fast]
    __m128i ranges16 = _mm_loadu_si128((const __m128i *)((const void *)ranges));
    ...
      __m128i b16 = _mm_loadu_si128((const __m128i *)((const void *)buf));
      int r = _mm_cmpestri(
          ranges16, ranges_size, b16, 16,
          _SIDD_LEAST_SIGNIFICANT | _SIDD_CMP_RANGES | _SIDD_UBYTE_OPS);

The variable is indeed only 5 bytes long, but that's because we want to use `pcmpestri` instructions to look at 2 pairs of byte sequences (`_SIDD_LEAST_SIGNIFICANT | _SIDD_CMP_RANGES | _SIDD_UBYTE_OPS`).  In other words this seems to be testing for characters in the 0-040 and 0177-0177 ranges.

[The documentation](https://msdn.microsoft.com/en-us/library/bb531465(v=vs.120).aspx) implies that this code has the arguments backwards, that we'll test all characters in `a` vs the ranges specified in `b`, which would make the text loaded from `buf` the ranges; we want `a` to be the ranges.)  The code itself is apparently from [picohttpparser](https://github.com/h2o/picohttpparser/blob/23b8a0308a540620729ce3f4c5d61db547831cad/picohttpparser.c#L92) and is probably correct.  There's a similar issue with the next call to `findchar_fast` for `ranges1`: to make the leak sanitizer happy, we need to have some pad bytes after the useful bytes.

As long as the variables are aligned and only the leading characters are actually used, the code should be OK.  The variable `ranges2` _is_ aligned, but `ranges1` is _not_ aligned.  I'll align `ranges2` and add pad bytes and change the arguments to `findchar_fast` to use constants, to keep the sanitizer happy...

chris3torek:
Aha, now it's a complaint about `GenerateResetPacket`.  Both that and `Generate403Packet` call `Copy` with sloppy set to `true`, which isn't necessary, we can use `sizeof` instead of `strlen` to get compile time constant sizes for the packets.

sangjinhan:
e5b944216f80962d057d242823dd1c77

PR #719, opened 2017-11-10
Fix bugs on bypass

shinae-woo:
- Fix types (uint32_t -> uint64_t)
- Use rdtsc() than current_tsc() --> in case of multiple bypass modules chained, they'd use quite stale values to calculate tsc.


melvinw:
Thanks!

cat

PR #720, opened 2017-11-10
Add drop counter to Queue module

sangjinhan:
... for accountability.

It adds overheads to the data path, but the impact should be minimal since it is per-batch cost.


melvinw:
Thanks!

cat

PR #724, opened 2017-11-14
utils: add utility functions to BigEndian and Ethernet::Address

sangjinhan:


core/utils/endian.h
```diff
@@ -152,6 +153,15 @@ class[[gnu::packed]] BigEndian final : public EndianBase<T> {
     return os;
   }
 
+  const std::vector<uint8_t> ToByteVector() const {
+    union {
+      T data;
+      uint8_t bytes[0];
```
chris3torek: Why `0` instead of `sizeof(T)`? (Should be OK either way, just seems a bit odd.)

sangjinhan: Oh thanks. I was fiddling with non-template version of this but forgot go rollback. Fixed.

melvinw:
Thanks!

cat

PR #728, opened 2017-11-16
bessctl: add measurement utilities

melvinw:
While developing https://github.com/nefeli/trafficgen I realized that a utility like this might be generally useful. As a usage example consider the simple packet generator below.

```
from measurement_utils import *

import scapy.all as scapy
import time

dst_mac = $DST_MAC!'02:11:22:33:44:55'

def make_pkt(src_mac='02:00:11:22:33:44', dst_mac='02:11:22:33:44:55',
              src_ip='10.0.0.1', dst_ip='11.0.0.1', size=64):
    eth = scapy.Ether(src=src_mac, dst=dst_mac)
    ip = scapy.IP(src=src_ip, dst=dst_ip)
    udp = scapy.UDP(sport=10001, dport=10002, chksum=0)
    payload = ('hello' + '0123456789' * 200)[:size - len(eth / ip / udp)]
    return bytes(eth / ip / udp / payload)

bess.add_worker(wid=1, core=1)
port_cfg = PortConfig('port0', port_id=0)
p = MeasureablePort(tx_port=port_cfg, rx_port=port_cfg, tx_ts_offset=42, rx_ts_offset=42)

templates = [make_pkt(src_mac=p.tx_port.mac_addr, dst_mac=dst_mac)]
src::Source() -> rewrite::Rewrite(templates=templates)
p.connect_tx(rewrite, 0)

sink::Sink()
p.connect_rx(sink, 0)

bess.resume_all()
for stats in p.rate_stats(rtt_percentiles=[50, 99]):
    print(stats)
    print('--')
    time.sleep(2)
```

```
$ bessctl run file conf/simple_pktgen.py DST_MAC='"3c:fd:fe:a5:40:58"'
[in / out] pkts: 0.000M / 0.000M, drops: 0.000M / 0.000M, bytes: 0 / 0
rtt_50: 0.000 us, rtt_99: 0.000 us
jit_50: 0.000 us, jit_99: 0.000 us
--
[in / out] pkts: 12.935M / 20.384M, drops: 0.001M / 3.218M, bytes: 827871136 / 1138372254
rtt_50: 8.900 us, rtt_99: 11.250 us
jit_50: 0.000 us, jit_99: 0.900 us
--
[in / out] pkts: 19.013M / 19.019M, drops: 0.001M / 0.001M, bytes: 1216921536 / 1217245004
rtt_50: 17.700 us, rtt_99: 21.750 us
jit_50: 0.000 us, jit_99: 1.800 us
--
[in / out] pkts: 18.971M / 18.979M, drops: 0.001M / 0.001M, bytes: 1214248420 / 1214668440
rtt_50: 17.700 us, rtt_99: 21.100 us
jit_50: 0.000 us, jit_99: 1.800 us
```

### DIFF
--- a/bessctl/measurement_utils.py
+++ b/bessctl/measurement_utils.py
@@ -88,10 +88,8 @@ class PortStats(object):
         Returns the dictionary of all RTT value if `percentile` is none.
         Otherwise returns the `percentile`th percentile RTT.
         """
-        if self.rtt is None:
-            return None
-        if percentile is None:
-            return self.rtt
+        if self.rtt is None or percentile is None:
+            self.rtt
         return self.rtt[percentile]
 
     def jitter(self, percentile=None):
@@ -99,27 +97,22 @@ class PortStats(object):
         Returns the dictionary of all jitter value if `percentile` is none.
         Otherwise returns the `percentile`th percentile jitter.
         """
-        if jitter is None:
-            return None
-        if percentile is None:
-            return self.jitter
+        if self.jitter is None or percentile is None:
+            self.jitter
         return self.jitter[percentile]
 
     def __str__(self):
+        def format_data(prefix, data, dst):
+            keys = sorted(data.keys())
+            for k in keys:
+                dst.append(data[k] / 1e3)
+            return ', '.join(['{}{}: {{:.3f}} us'.format(prefix, k) for k in keys])
+
         fmt = '[in / out] pkts: {:.3f}M / {:.3f}M, drops: {:.3f}M / {:.3f}M, bytes: {} / {}'
         rtt_data = []
         if self.rtt is not None:
-            fmt += '\n'
-            fmt += ', '.join(['rtt_{}: {{:.3f}} us'.format(k)
-                             for k in sorted(self.rtt.keys())])
-            for k in sorted(self.rtt.keys()):
-                rtt_data.append(self.rtt[k] / 1e3)
-
-            fmt += '\n'
-            fmt += ', '.join(['jit_{}: {{:.3f}} us'.format(k)
-                             for k in sorted(self.jitter.keys())])
-            for k in sorted(self.jitter.keys()):
-                rtt_data.append(self.jitter[k] / 1e3)
+            fmt += '\n' + format_data('rtt', self.rtt, rtt_data)
+            fmt += '\n' + format_data('jitter', self.jitter, rtt_data)
 
         return fmt.format(
             self.pkts_in() / 1e6, self.pkts_out() / 1e6,
@@ -179,11 +172,8 @@ class PortStatsGenerator(object):
     @staticmethod
     def aggregate_rtt(old, new):
         if old is None:
-            return copy.copy(new)
-        agg = dict()
-        for k in old:
-            agg[k] = (old[k] + new[k]) / 2
-        return agg
+            return new.copy()
+        return {k: (old[k] + new[k]) / 2 for k in old}
 
     def next(self):
         throughput = self.bess.get_port_stats(self.tx_port)
@@ -247,8 +237,8 @@ class PortConfig(object):
         set, no PortInc will be created. If `bess` is None, will attempt to
         create all ports and modules on bessd running on localhost.
         """
-        if sum(map(int, [pci is not None, port_id is not None, vdev is not None])) != 1:
-            raise Exception(
+        if [pci, port_id, vdev].count(None) != 2:
+            raise TypeError(
                 'exactly one of `pci`, `port_id` or `vdev` must be specified')
 
         self.name = name
@@ -276,7 +266,7 @@ class Port(object):
 
         if conf.pci is not None:
             self.pmd = bess.create_port('PMDPort', conf.name,
-                                        {'pci': pci, 'num_inc_q':
+                                        {'pci': conf.pci, 'num_inc_q':
                                             conf.num_queues, 'num_out_q': conf.num_queues})
         elif conf.port_id is not None:
             self.pmd = bess.create_port('PMDPort', conf.name,
--- a/bessctl/measurement_utils.py
+++ b/bessctl/measurement_utils.py
@@ -1,0 +1,393 @@
+# Copyright (c) 2017, Nefeli Networks, Inc.
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# * Redistributions of source code must retain the above copyright notice, this
+# list of conditions and the following disclaimer.
+#
+# * Redistributions in binary form must reproduce the above copyright notice,
+# this list of conditions and the following disclaimer in the documentation
+# and/or other materials provided with the distribution.
+#
+# * Neither the names of the copyright holders nor the names of their
+# contributors may be used to endorse or promote products derived from this
+# software without specific prior written permission.
+#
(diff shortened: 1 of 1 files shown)


bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def __init__(self, name, num_queues=1, pci=None, port_id=None, vdev=None,
+                 no_tx=False, no_rx=False):
+        """
+        Creates a configuraiton for a Port named `name` and configures it with `num_queues` tx and
+        rx queues. If `no_tx` is set, no PortOut will be created. If `no_rx` is
+        set, no PortInc will be created. If `bess` is None, will attempt to
+        create all ports and modules on bessd running on localhost.
+        """
+        if sum(map(int, [pci is not None, port_id is not None, vdev is not None])) != 1:
```
chris3torek: Simpler:

    if [pci, port_id, vdev].count(None) != 2:

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+            raise Exception(
```
chris3torek: Probably should be `TypeError` (that seems to be the most typical variant on "bad arguments").

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+import copy
+import os
+import sys
+import time
+
+try:
+    this_dir = os.path.dirname(os.path.realpath(__file__))
+    bessctl = os.path.join(this_dir, 'bessctl')
+    sys.path.insert(1, os.path.join(this_dir, '../../../'))
```
chris3torek: Nothing you can do about this (I think), but we need a Better Way! 😛

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+    print('Cannot import the API module (pybess)', file=sys.stderr)
+    raise
+
+
+def get_local_bess_handle():
+    bess = BESS()
+    try:
+        bess.connect()
+    except BESS.APIError:
+        raise Exception('BESS is not running')
```
chris3torek: We probably should have a bess exception class, too. Again not something to fix here though.

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def rtt(self, percentile=None):
+        """
+        Returns the dictionary of all RTT value if `percentile` is none.
+        Otherwise returns the `percentile`th percentile RTT.
+        """
+        if self.rtt is None:
+            return None
+        if percentile is None:
+            return self.rtt
```
chris3torek: Nothing wrong with these as written, but it occurs to me that you could use:

    if self.rtt is None or percentile is None:
        return self.rtt

to save a couple of lines.  Same goes for `jitter` below.

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+        return self.jitter[percentile]
+
+    def __str__(self):
+        fmt = '[in / out] pkts: {:.3f}M / {:.3f}M, drops: {:.3f}M / {:.3f}M, bytes: {} / {}'
+        rtt_data = []
+        if self.rtt is not None:
+            fmt += '\n'
+            fmt += ', '.join(['rtt_{}: {{:.3f}} us'.format(k)
+                             for k in sorted(self.rtt.keys())])
+            for k in sorted(self.rtt.keys()):
```
chris3torek: I hate to see `sorted(arg)` twice like this, and I guess we are guaranteed that if self.rtt is not None, self.jitter is also not None, but it might be nice to have a static method that takes a string-and-dict pair and computes the format and data from the sorted keys, or returns '' if they are None.  Then you could do:

    rtt_data = []
    fmt += self.format_data('rtt', self.rtt, rtt_data)
    fmt += self.format_data('jit', self.jitter, rtt_data)
    return fmt.format(fmt, ...)
and have the `@staticmethod` do a single sorted() and data-append.

Kind of sixes though.

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        return self.next()
+
+    @staticmethod
+    def throughput_delta(old, new, secs):
+        delta = copy.copy(old)
```
chris3torek: Consider using `old.copy()` here, and ... (continued next comment, can't comment on a big line range 😺 )

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+            delta.inc.bytes = (new.inc.bytes - old.inc.bytes) // secs
+            delta.out.packets = (new.out.packets - old.out.packets) // secs
+            delta.out.dropped = (new.out.dropped - old.out.dropped) // secs
+            delta.out.bytes = (new.out.bytes - old.out.bytes) // secs
+        return delta
+
+    @staticmethod
+    def aggregate_rtt(old, new):
+        if old is None:
+            return copy.copy(new)
```
chris3torek: ... `new.copy()` here, and then you won't need to `import copy`.  Not really a big deal either way (`copy.copy` deals correctly with things that aren't dictionaries, but dictionaries do have `.copy` methods).

bessctl/measurement_utils.py
```diff
@@ -0,0 +1,393 @@
...
+        agg = dict()
+        for k in old:
+            agg[k] = (old[k] + new[k]) / 2
+        return agg
```
chris3torek: Shorter (not really any better):

    return { k: (old[k] + new[k]) / 2 for k in old }

More important question: why is the aggregate always a straight unweighted average?

melvinw:
Thanks for the quick review @chris3torek ! Will address shortly

bessctl/measurement_utils.py
melvinw: Agreed!

bessctl/measurement_utils.py
melvinw: Definitely cleaner. Switched

bessctl/measurement_utils.py
melvinw: Switched to `foo.copy()` where I could. Most of the copies here are happening on protobufs.

bessctl/measurement_utils.py
melvinw: Fixed! :)

bessctl/measurement_utils.py
melvinw: Agreed! Switched

bessctl/measurement_utils.py
melvinw: Nice! Much easier to read. Replaced

bessctl/measurement_utils.py
melvinw: Fixed

bessctl/measurement_utils.py
```diff
@@ -88,38 +88,31 @@ def rtt(self, percentile=None):
         Returns the dictionary of all RTT value if `percentile` is none.
         Otherwise returns the `percentile`th percentile RTT.
         """
-        if self.rtt is None:
-            return None
-        if percentile is None:
-            return self.rtt
+        if self.rtt is None or percentile is None:
+            self.rtt
```
chris3torek: lost a `return` keyword (here and below)...

melvinw: Doh. Fixed

sangjinhan:
tumblr_mbarwefvw61qh66wqo1_1280

Issue #717, opened 2017-11-08
Possible bug in weighted_fair when resouce is 'bit'


nemethf:
I think the current behavior of the wighted_fair scheduler when the resource is set to 'bit' is counter-intuitive.  Take a look at the following bess script:
```python
bess.add_tc('root', 'weighted_fair', wid=0, resource='bit')

tagger= SetMetadata(attrs=[{'name': 'tag', 'size': 4, 'value_int': 0}])
split = Split(attribute='tag', size=4)

s::Source() -> tagger -> split

for i in range(2):
  q = Queue()
  name = 't-%s' % i
  bess.add_tc(name=name, parent='root', policy='round_robin', share=1)
  q.attach_task(parent=name)
  split:i -> q -> Sink()

bess.attach_task(s.name, wid=0)
```
Running the script results in a dead-lock because the split module never sends a packet to the second queue, and therefore the queue never forwards a packet although it always 'runnable'.
```
+--------+                   +-------------+                   +----------+                   +-----------+             +-------+
|   s    |                   |  setattr0   |                   |  split0  |                   |  queue0   |             | sink0 |
| Source |  :0 81434528 0:   | SetMetadata |  :0 81475136 0:   |  Split   |  :0 81579712 0:   |   Queue   |  :0 32 0:   | Sink  |
|        | ----------------> |             | ----------------> |          | ----------------> | 1023/1024 | ----------> |       |
+--------+                   +-------------+                   +----------+                   +-----------+             +-------+
                                                                 |
                                                                 | :1 0 0:
                                                                 v
                                                               +----------+                   +-----------+
                                                               |  queue1  |                   |   sink1   |
                                                               |  Queue   |  :0 0 0:          |   Sink    |
                                                               |  0/1024  | ----------------> |           |
                                                               +----------+                   +-----------+

<worker 0>
  +-- !default_rr_0            round_robin
      +-- root                 weighted_fair
      |   +-- t-0              round_robin         share: 1
      |   |   +-- !leaf_queue0:0 leaf
      |   +-- t-1              round_robin         share: 1
      |       +-- !leaf_queue1:0 leaf
      +-- !leaf_s:0            leaf
```
I somehow expected that the scheduler would choose queue0 because queue1 was empty and queue0 would opportunistically work all the time.  Now that I understand the current behavior I can live with it, but still wonder whether it is intended.  Thanks.

sangjinhan:
[...]

PR #729, opened 2017-11-16
Fix a bug that starves all children except an unblocked child in weight_fair TC

sangjinhan:
In the current implementation of stride scheduling for weighted fair sharing, the "pass" value of unblocked (joined again) child TCs are initialized to 0. Then only that TC will be scheduled until its pass value has caught up with the that of other TCs. This caused scheduling bug reported in #717.

In addition, a minor bug, where frequently blocked TCs are more scheduled than they should be, has also been fixed (`client->remain` in Fig 3 in [the original paper](http://www.read.seas.harvard.edu/~kohler/class/aosref/waldspurger95stride.pdf))

### DIFF
--- a/core/traffic_class.cc
+++ b/core/traffic_class.cc
@@ -192,8 +192,7 @@ bool WeightedFairTrafficClass::AddChild(TrafficClass *child,
   }
 
   child->parent_ = this;
-  ChildData child_data{
-      .stride = STRIDE1 / share, .pass = NextPass(), .c = child};
+  ChildData child_data{STRIDE1 / share, {NextPass()}, child};
   if (child->blocked_) {
     blocked_children_.push_back(child_data);
   } else {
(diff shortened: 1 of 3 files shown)


core/traffic_class.cc
```diff
@@ -187,21 +187,12 @@ std::vector<TrafficClass *> WeightedFairTrafficClass::Children() const {
...
-  }
-
-  if (!share) {
+  if (child->parent_ || share == 0) {
     return false;
   }
 
   child->parent_ = this;
-  WeightedFairTrafficClass::ChildData child_data{STRIDE1 / share, pass, child};
+  ChildData child_data{STRIDE1 / share, {NextPass()}, child};
```
barath: Since this is initializing a struct with a union, for clarity would you be able to use designated initializers?

sangjinhan: Yeah, I wanted to do that, but g++5 refused with "non-trivial designated initializers are not supported" 😞

barath: Hrm... I was hoping that that wouldn't happen with unions because it'd be smart enough to know that one option among the union is set.

melvinw:
Thanks!

cat

PR #733, opened 2017-11-21
modules/hash_lb: add configurable n-tuple mode

melvinw:


melvinw:
It's backwards compatible with the old `l2,l3,l4` modes. Using `bessctl/samples/conf/samples/hash_lb.bess` as a benchmark, it is a bit slower now (33.9 Mpps vs 41.8 Mpps on my machine). I took out the specialized `Lb*` functions, but I'm happy to put them back if performance is important here.

melvinw:
Marking not ready while I put back the 'Lb*` functions

melvinw:
Ready to go now. Impact on `l2`, `l3` and `l4` is negligible now.

sangjinhan:
tumblr_m140vwamm61r0bljro1_1280

chris3torek:
(I wondered about the template specializations here too, but the only other thing I thought we could do was make this use a function pointer, and that would probably be worse/slower.)

melvinw:
Ah, I agree-- it doesn't really buy us anything-- it just felt slightly cleaner to me.

PR #736, opened 2017-11-29
Fix a bug on module crash testing

shinae-woo:
- Previously, bessd crash from a module running detected at the moments
of running tests for next module. As a results, error reporting is
cascaded and incorrect.
- This commit adds a check whether bess have been crashed in the same test case.


melvinw:
Thanks!

cat

PR #738, opened 2017-11-30
Initialize byte count in Track gate hook

dtnaylor:


sangjinhan:
tumblr_m172ukjirm1qz5dg8o1_1280

PR #739, opened 2017-12-02
Iteratively scheduling modules by a task 

shinae-woo:
This PR changes a module scheduling within a task from recursive way within scheduler to iterative way. This provides following benefits.

1. Functionality: Each modules' execution boundary become clear as a Task fully control and know which modules will be being executing. As a result, sharing a module from multiple threads (or Workers in BESS contexts) can be implemented more efficiently as locking cost become smaller, or we can even remove locking. This can enable various future functionality such as do not reject sharing of a module between different workers (currently it is rejected), minimally interrupting a data-plane processing while update per-module configurations, Batching sequentially allocated packets to free at once, etc. 

2. Performance: Get the benefit of packet batching back whenever possible. Previously, once a PacketBatch split into multiple packet batches, they will process separately and not merged back. However, now if the splitted packets paths are merged back together, the scheduler merged back the packet to regain the benefit of packet batching.

There should be following PRs to address more issues very soon,
1. Make both worker context and tasks arguments from modules as arguments, and move some per-task member variable (such as per-gate batch, PacketBatch pool) into workers.
2. Adjust the size of packet batch pool for a task before resuming. 
3. Makes the tasks arguments non-const.


shinae-woo:
Sorry for such huge things, but I can't split out this functional changes. But, the core changes are mostly in task.cc, module.h, which needs most efforts of the reviews.

core/task.cc
```diff
@@ -44,13 +45,76 @@ void Task::Attach(bess::LeafTrafficClass *c) {
...
+  bess::PacketBatch init_batch;
+  ClearPacketBatch();
+
+  // Start from the first module (task module)
+  struct task_result result = module_->RunTask(this, &init_batch, arg_);
+
+  // next_gate_: Continuously run if modules are chainned
+  // igates_to_run_ : If next module connection is not chained (merged),
+  // check priority to choose which module run next
+  while (1) {
```
sangjinhan: very minor: existing code seems to prefer `while (true)` over `while (1)`.

core/gate.h
```diff
@@ -170,17 +170,34 @@ class OGate;
...
+        priority_(),
+        g_idx_(),
+        mergeable_(false) {}
 
   const std::vector<OGate *> &ogates_upstream() const {
     return ogates_upstream_;
   }
 
+  void SetPriority(uint32_t priority) { priority_ = priority; }
+  void SetUniqueIdx(uint32_t g_idx) { g_idx_ = g_idx; }
```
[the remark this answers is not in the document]
shinae-woo: Sure.

core/gate.h
```diff
@@ -170,17 +170,34 @@ class OGate;
...
+  uint32_t priority() const { return priority_; }
+  uint32_t g_idx() const { return g_idx_; }
+  bool mergeable() const { return mergeable_; }
+
   void PushOgate(OGate *og);
   void RemoveOgate(const OGate *og);
 
  private:
   std::vector<OGate *> ogates_upstream_;  // previous ogates connected with
+  uint32_t priority_;
```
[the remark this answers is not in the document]
shinae-woo: I am sorry, `uint32_t priority_;`..

core/gate.h
```diff
@@ -170,17 +170,34 @@ class OGate;
...
+  uint32_t g_idx_;  // a globally unique igate index
```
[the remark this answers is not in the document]
shinae-woo: Right

core/module.h
```diff
@@ -234,18 +241,32 @@ class alignas(64) Module {
...
+   * NOTE:
+   *   1. Order is preserved for packets with the same gate.
+   *   2. No ordering guarantee for packets with different gates.
+   *
+   * Update on 11/27/2017, by Shinae Woo
+   * This interface will become DEPRECATED.
+   * Consider using new interfafces supporting faster data-plane support
+   * DropPacket()/EmitPacket()
+   * */
+  inline void RunSplit(const Task *task, const gate_idx_t *ogates,
```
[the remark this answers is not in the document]
shinae-woo: I tried to use `[[dprecated]]`, and interestingly, g++ supports it even with c++11, but clang does not. Once we can safely switch into c++14, then I can amend it. Do you want me to execute the switching?

core/module.h
```diff
@@ -465,20 +487,112 @@ inline void Module::RunChooseModule(gate_idx_t ogate_idx,
...
+}
+
+inline void Module::RunNextModule(const Task *task, bess::PacketBatch *batch) {
+  RunChooseModule(task, 0, batch);
+}
+
+inline void Module::DropPacket(const Task *task, bess::Packet *pkt) {
+  task->dead_batch()->add(pkt);
+  if (static_cast<size_t>(task->dead_batch()->cnt()) >=
+      bess::PacketBatch::kMaxBurst)
```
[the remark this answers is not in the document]
shinae-woo: ike.

core/task.h
```diff
@@ -33,55 +33,149 @@
...
+  // Do not track used/unsued for efficiency
+  bess::PacketBatch *AllocPacketBatch() const {
+    CHECK_LT(pbatch_idx_, MAX_PBATCH_CNT);
+    bess::PacketBatch *batch = &pbatch_[pbatch_idx_++];
+    batch->clear();
+    return batch;
+  }
+
+  void UpdatePerGateBatch(uint32_t igate_cnt, uint32_t ogate_cnt) const {
+    if (igate_batch_.capacity() < igate_cnt) {
```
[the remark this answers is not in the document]
shinae-woo: Yes. This is intended. The reason for doing this is that the way I access the vector not trigger automatic vector resizing.

core/module_graph.cc
```diff
@@ -81,19 +85,148 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+  ogate_cnt_ = 0;
+  while (!ogates_queue.empty()) {
+    bess::OGate *ogate = ogates_queue.top();
+    ogates_queue.pop();
+
+    ogate->SetUniqueIdx(++ogate_cnt_);
+  }
+}
+
+void ModuleGraph::ConfigureTasks() {
```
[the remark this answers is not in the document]
shinae-woo: That's correct.

shinae-woo:
Thank you for the review, I will push an amendment commit for the review, and will push following PRs as I described.

core/module.h
```diff
@@ -465,20 +489,113 @@ inline void Module::RunChooseModule(gate_idx_t ogate_idx,
...
+  if (unlikely(ogates_.size() <= ogate_idx) || unlikely(!ogates_[ogate_idx])) {
+    DropPacket(task, pkt);
+    return;
+  }
+
+  // Put a packet into the ogate
+  bess::OGate *ogate = ogates_[ogate_idx];
+  bess::PacketBatch *batch = task->get_gate_batch(ogate);
+  if (!batch) {
+    if (ogate->hooks().size()) {
```
barath: `!ogate->hooks().empty()`

core/module.h
```diff
@@ -465,20 +489,113 @@ inline void Module::RunChooseModule(gate_idx_t ogate_idx,
...
+      } else {
+        task->set_gate_batch(ogate, task->get_gate_batch(ogate->igate()));
+      }
+      gate_without_hook[gate_without_hook_cnt++] = ogate_idx;
+    }
+    batch = task->get_gate_batch(ogate);
+  }
+
+  if (static_cast<size_t>(batch->cnt()) >= bess::PacketBatch::kMaxBurst) {
+    if (ogate->hooks().size()) {
```
barath: `!ogate->hooks().empty()`

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+
+      bess::IGate *next_igate = ogates[i]->igate();
+      if (visited_igates.count(next_igate) != 0 ||  // This is a loop or
+          next_igate->priority() >= priority) {     // visited by longer path
+        continue;
+      }
+
+      visited_igates.insert(next_igate);
+      next_igate->SetPriority(priority);
+      priority++;
```
barath: Is there a reason to increment `priority` and then decrement again?  Why not replace these three lines with:

`PropagateIGatePriority(next_igate, visited_igates, priority+1);`

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+      PropagateIGatePriority(next_igate, visited_igates, priority);
+      priority--;
+      visited_igates.erase(next_igate);
+    }
+  }
+}
+
+void ModuleGraph::SetIGatePriority(Module *task_module) {
```
barath: Some code in this function seems the same as above -- could it be consolidated?

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+void ModuleGraph::SetUniqueGateIdx() {
+  bess::utils::extended_priority_queue<bess::OGate *> ogates_queue;
+  bess::utils::extended_priority_queue<bess::IGate *, IGateGreater>
+      igates_queue;
+  std::unordered_set<bess::IGate *> igates_pushed;
+
+  for (auto const &e : all_modules_) {
+    std::vector<bess::OGate *> ogates = e.second->ogates();
+    for (size_t i = 0; i < ogates.size(); i++) {
+      if (!ogates[i])
```
barath: Braces around `if` block.

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+        continue;
+
+      ogates_queue.push(ogates[i]);
+
+      bess::IGate *igate = ogates[i]->igate();
+      if (igates_pushed.count(igate) != 0)
```
barath: Braces around `if` block.

core/task.h
```diff
@@ -33,55 +33,150 @@
...
 // Functor used by a leaf in a Worker's Scheduler to run a task in a module.
 class Task {
+ private:
+  // Used by operator().
+  Module *module_;
+  void *arg_;                  // Auxiliary value passed to Module::RunTask().
+  bess::LeafTrafficClass *c_;  // Leaf TC associated with this task.
+
+  struct GateBatchGreater {
+    bool operator()(std::pair<bess::IGate *, bess::PacketBatch *> left,
```
barath: Should be passed as const references.

core/task.h
```diff
@@ -33,55 +33,150 @@
...
+      : module_(m),
+        arg_(arg),
+        c_(nullptr),
+        igates_to_run_(),
+        next_gate_(),
+        next_batch_(),
+        pbatch_idx_(),
+        pbatch_(
+            new bess::PacketBatch[MAX_PBATCH_CNT]),  // XXX Need to adjust size
+        gate_batch_(std::vector<bess::PacketBatch *>(64, 0)) {
```
barath: Could you add `current_igate_` (and in general all members) to the initializer list?

core/task.h
```diff
@@ -33,55 +33,150 @@
...
+    } else {
+      bess::PacketBatch *ibatch = get_gate_batch(ig);
+      if (ibatch && (static_cast<size_t>(ibatch->cnt() + batch->cnt()) <
+                     bess::PacketBatch::kMaxBurst)) {
+        // merge two batches
+        get_gate_batch(ig)->add(batch);
+      }
+      // set the input as new batch
+      set_gate_batch(ig, batch);
+      igates_to_run_.push(std::make_pair(ig, batch));
```
barath: Maybe use `emplace` instead? (And then I believe you shouldn't need `std::make_pair`, though we may still run into a compiler constraint there...)

core/task.h
```diff
@@ -33,55 +33,150 @@
...
+  bess::PacketBatch *AllocPacketBatch() const {
+    DCHECK_LT(pbatch_idx_, MAX_PBATCH_CNT);
+    bess::PacketBatch *batch = &pbatch_[pbatch_idx_++];
+    batch->clear();
+    return batch;
+  }
+
+  void UpdatePerGateBatch(uint32_t gate_cnt) const {
+    if (gate_batch_.capacity() < gate_cnt) {
+      gate_batch_.resize(gate_cnt, nullptr);
```
barath: Dumb question: do you really want `reserve` here, not `resize`?  I ask because you're checking `capacity` not `size`.

core/modules/replicate.cc
```diff
@@ -64,26 +64,17 @@ CommandResponse Replicate::CommandSetGates(
...
-  bess::PacketBatch out_gates[ngates_];
-  for (int i = 0; i < ngates_; i++) {
-    out_gates[i].clear();
-  }
-
+void Replicate::ProcessBatch(const Task *task, bess::PacketBatch *batch) {
   for (int i = 0; i < batch->cnt(); i++) {
     bess::Packet *tocopy = batch->pkts()[i];
-    out_gates[0].add(tocopy);
+    EmitPacket(task, tocopy, 0);
```
chris3torek: We could stand to have a promise here that EmitPacket() doesn't drop the packet before we can copy it. I think that's _mostly_ true, but if we put a copy into a dead batch and the dead batch fills up, we'll call `deadend()`, which might free it prematurely here. The obvious alternative is to copy before calling `EmitPacket` here (which might be a better way to code this anyway).

core/task.cc
```diff
@@ -45,12 +46,58 @@ void Task::Attach(bess::LeafTrafficClass *c) {
...
 
 struct task_result Task::operator()(void) const {
-  return module_->RunTask(arg_);
+  bess::PacketBatch init_batch;
+  ClearPacketBatch();
+
+  // Start from the first module (task module)
+  struct task_result result = module_->RunTask(this, &init_batch, arg_);
+
+  // next_gate_: Continuously run if modules are chainned
```
chris3torek: trivial typo: `chained`

shinae-woo:
Thank you for your great reviews @barath @chris3torek @sangjinhan. I will push additional commits.

core/modules/replicate.cc
shinae-woo: Great point!! Thanks!

core/module.h
```diff
@@ -226,6 +227,12 @@ class alignas(64) Module {
 
   static const Commands cmds;
 
+  int gate_with_hook_cnt = 0;
+  gate_idx_t gate_with_hook[bess::PacketBatch::kMaxBurst];
+
+  int gate_without_hook_cnt = 0;
+  gate_idx_t gate_without_hook[bess::PacketBatch::kMaxBurst];
```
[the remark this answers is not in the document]
shinae-woo: These variables are gone with my next PR I am preparing, but should be fixed here anyway. Thanks!

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+void ModuleGraph::PropagateIGatePriority(
+    bess::IGate *igate, std::unordered_set<bess::IGate *> &visited_igates,
+    uint32_t priority) {
+  if (igate->module()->is_task()) {
+    return;
+  } else {
+    std::vector<bess::OGate *> ogates = igate->module()->ogates();
+    for (size_t i = 0; i < ogates.size(); i++) {
+      if (!ogates[i]) {
+        break;
```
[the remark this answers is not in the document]
shinae-woo: OMG, I thought I have fixed! There are some sparse use cases as well (e.g., vlan split). Thank you for pointing out.

core/module_graph.cc
```diff
@@ -81,19 +84,133 @@ void ModuleGraph::UpdateSingleTaskGraph(Module *task_module) {
...
+    igate->SetUniqueIdx(gate_cnt_++);
+  }
+
+  while (!ogates_queue.empty()) {
+    bess::OGate *ogate = ogates_queue.top();
+    ogates_queue.pop();
+
+    ogate->SetUniqueIdx(gate_cnt_++);
+  }
+}
```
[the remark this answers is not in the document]
shinae-woo: This set of functions in module_graph.cc can definitely be integrated and cleared. I plan to integrate them including back-pressure, gate priority in next PR.

core/modules/buffer.h
```diff
@@ -40,7 +40,7 @@ class Buffer final : public Module {
 
   void DeInit() override;
 
-  void ProcessBatch(bess::PacketBatch *batch) override;
+  void ProcessBatch(const Task* task, bess::PacketBatch* batch) override;
```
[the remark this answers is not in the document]
shinae-woo: That is what happens in my git hook. only for drr.h file... I will figure it out. Thank you!
image

shinae-woo:
Note: The module-side interface would be changed one more time, to remove access from thread_local storage (currently per-thread `Worker *ctx`). As a result, modules within a task passing by not only `Task *task` as an input param, but but `Context *context` which has all temporary contexts of running a task including worker, task, temporary pkt batches, etc.

core/module.h
```diff
@@ -465,20 +505,113 @@ inline void Module::RunChooseModule(gate_idx_t ogate_idx,
...
+  // Check if valid ogate is set
+  if (unlikely((ogates_.size() <= ogate_idx) || (!ogates_[ogate_idx]))) {
+    DropPacket(task, pkt);
+    return;
+  }
+
+  // Put a packet into the ogate
+  bess::OGate *ogate = ogates_[ogate_idx];
+  bess::PacketBatch *batch = task->get_gate_batch(ogate);
+  if (!batch) {
```
sangjinhan: You can rearrange the code like this
```c++
   if (!batch) {
     if (!ogate->hooks().empty()) {
       // Having separate batch to run ogate hooks
      batch = task->AllocPacketBatch();
      task->set_gate_batch(ogate, batch);
       gate_with_hook_[gate_with_hook_cnt_++] = ogate_idx;
     } else {
       // If no ogate hooks, just use next igate batch
      batch = task->get_gate_batch(igate);
      if (batch == nullptr) {
        batch = task->AllocPacketBatch();
        task->AddToRun(igate, batch);
        task->set_gate_batch(ogate, batch);
       }
       gate_without_hook_[gate_without_hook_cnt_++] = ogate_idx;
     }
   }
```
to eliminate use of temporary variable and make it easier to see when the `batch` pointer is assigned.

core/task.h
```diff
@@ -33,55 +33,152 @@
...
+    if (next_gate_ == nullptr &&
+        !ig->mergeable()) {  // optimization for chained
+      next_gate_ = ig;
+      next_batch_ = batch;
+    } else {
+      bess::PacketBatch *ibatch = get_gate_batch(ig);
+      if (ibatch && (static_cast<size_t>(ibatch->cnt() + batch->cnt()) <
+                     bess::PacketBatch::kMaxBurst)) {
+        // merge two batches
+        get_gate_batch(ig)->add(batch);
```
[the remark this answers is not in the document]
sangjinhan: or you wanted to put the two lines below into a `else { }` block?

core/modules/exact_match.cc
```diff
@@ -241,9 +235,11 @@ void ExactMatch::ProcessBatch(bess::PacketBatch *batch) {
     return pkt->head_data<uint8_t *>() + f.offset;
   };
   table_.MakeKeys(batch, buffer_fn, keys);
-  table_.Find(keys, out_gates, cnt, default_gate);
 
-  RunSplit(out_gates, batch);
+  for (int i = 0; i < batch->cnt(); i++) {
```
sangjinhan: Unfortunately the compiler doesn't seem to do very well with `batch->cnt()`. Manually doing `int cnt=batch->cnt();` before the loop is much faster, at least on my machine. You may want to check other modules as well.

core/task.h
```diff
@@ -33,55 +33,152 @@
...
   void Attach(bess::LeafTrafficClass *c);
 
+  inline void AddToRun(bess::IGate *ig, bess::PacketBatch *batch) const {
+    if (next_gate_ == nullptr &&
+        !ig->mergeable()) {  // optimization for chained
+      next_gate_ = ig;
+      next_batch_ = batch;
+    } else {
+      bess::PacketBatch *ibatch = get_gate_batch(ig);
+      if (ibatch && (static_cast<size_t>(ibatch->cnt() + batch->cnt()) <
```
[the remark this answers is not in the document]
shinae-woo: <= is the correct way. thanks.

core/task.h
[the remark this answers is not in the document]
shinae-woo: line 134 is merging `batch` into `ibatch`. I will change the code `get_gate_batch(ig)->add(batch)` to `ibatch->add(batch)` (More clearer).

those two lines should be in else{} block

shinae-woo:
@sangjinhan thank you for such detailed comments. Will fix them all.

sangjinhan:
Rebased to the latest master branch

sangjinhan:
Merged (finally), thanks!
e57957eacdb22c3c3595524986d143a8_1484629829_9664

PR #740, opened 2017-12-04
bessctl: Add tcpdump request-only command

nlsun:


nlsun:
Yes, it is if something else has created a pipe and is listening on it. This can then be used to write to such a specified pipe.

bessctl/commands.py
```diff
@@ -502,6 +506,9 @@ def bind_var(cli, var_type, line):
         if re.match(r'^(\*|[_a-zA-Z][\w]*)$', val) is None:
             raise cli.BindError('"name" must be "*" or [_a-zA-Z][_a-zA-Z0-9]*')
 
+    elif var_type == 'fifo':
+        pass
+
```
shinae-woo: If 'fifo' is for already existing one, then how about setting var_candidates as existing filenames?

bessctl/commands.py
```diff
@@ -1899,6 +1930,16 @@ def tcpdump_module(cli, module_name, direction, gate, opts):
...
 
+@cmd('tcpdump_request MODULE FIFO [DIRECTION] [GATE] [TCPDUMP_OPTS...]',
+     'Request packet capture on a gate')
+def tcpdump_request_module(cli, module_name, fifo, direction, gate, opts):
+    # TODO(nlsun) It may be better to implement this as a flag in the `tcpdump`
+    # command, but there doesn't seem to be a mechanism for adding flags at
+    # this time.
+    _capture_request_module(cli, module_name, direction,
+                            gate, opts, 'tcpdump', cli.bess.tcpdump, fifo)
+
```
shinae-woo: I agree with your TODO comments. In my opinion, clearing up those commands is quite important as it is open to users.  Is it possible that 'tcpdump_request'  integrated with 'tcpdump' along with integration of 'capture_request_module' and 'capture_module'? And follows the existing command syntax? For example, existing cmd does not use '_' but multiple tokens.

shinae-woo:
If #744 PR merged, then it will make your things easier?

nlsun:
Well, this is more for sending bess the message to start/stop the tcpdump. The deferrable/reconnectable tcpdump seems to solve a different issue

nlsun:
Closing, since this seems like this is a rather niche use case

PR #742, opened 2017-12-04
Upgrade from C++11 to C++14

sangjinhan:
g++ 5, which is our minimum requirement, supports C++14. So why not?


shinae-woo:
Looks good, and no side effect detected from CI

melvinw:
Thanks! cat

PR #743, opened 2017-12-05
Retire nvport and snbuf

shinae-woo:
R.I.P.
- It was being stale for a while.
- I was the last user of nvport, and my project has been ported into new interfaces (vhost-user).
- Maintaining this set of code requires extra efforts without no gain.

BYE!


apanda:
LGTM, thanks

apanda:
image

PR #744, opened 2017-12-05
deferrable, reconnect-able fifos

chris3torek:
This has only been lightly tested (I need to make sure it doesn't break tcpdump and pcapng in particular), but I'm putting it up for review before I get stuck on the freeway for hours...

### DIFF
--- a/core/utils/fifo_opener.cc
+++ b/core/utils/fifo_opener.cc
@@ -150,10 +150,8 @@ void FifoOpenerThread::Run() {
     fd = open(fifopath_, O_WRONLY);
   } while (fd < 0 && !IsExitRequested());
 
-  // It's open, so we're almost done.  Defer exit requests so
-  // we can make system calls without getting interrupted by
-  // a knock thread if there is one (or one starts here).
-  PushDefer();
+  // It's open, so we're irrevocably on the way out.
+  BeginExiting();
 
   // Set up the file descriptor: it should be non-blocking.
   // It also needs additional setup, which is something our
--- a/core/utils/fifo_opener.cc
+++ b/core/utils/fifo_opener.cc
@@ -1,0 +1,304 @@
+// Copyright (c) 2017, Nefeli Networks, Inc.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are met:
+//
+// * Redistributions of source code must retain the above copyright notice, this
+// list of conditions and the following disclaimer.
+//
+// * Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the following disclaimer in the documentation
+// and/or other materials provided with the distribution.
+//
+// * Neither the names of the copyright holders nor the names of their
+// contributors may be used to endorse or promote products derived from this
+// software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+// IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+// ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+// LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <cassert>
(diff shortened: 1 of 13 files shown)


chris3torek:
(rebased with autosquash)

core/utils/syscallthread.h
```diff
@@ -0,0 +1,443 @@
...
+      thread_.join();
+    }
+    if (knock_thread_.joinable()) {
+      // Kick the knock thread now, to make it finish early.
+      KickThread(&knock_thread_);
+      knock_thread_.join();
+    }
+  }
+
+  bool IsExitRequested() { return exit_requested_; }
```
chris3torek: (minor) should be `bool IsExitRequested() const`, likewise for `Done` just below - though a const instance of `*this` makes little sense for either one anyway.

core/utils/syscallthread.h
```diff
@@ -0,0 +1,443 @@
...
+ * into a sub-object that is given to the thread, so that we can
+ * use thread_.detach() and knock_thread_.detach()).  If we do this
+ * we need new/dispose, std::unique_ptr<...>, and/or refcounts.
+ */
+template <bool ReliableSignals>
+class SyscallThread {
+ public:
+  // NB: order matters here.  Thread state progresses linearly
+  // (except for Reset() which must not be allowed to race).
+  enum ThreadState { TSNotStarted, TSStarting, TSReady, TSExiting, TSDone };
```
chris3torek: Style guide says these should be `kFoo` constants: https://google.github.io/styleguide/cppguide.html#Enumerator_Names.  Not sure if TS (short for thread state) should be Ts, or maybe just rely on them being inside the `ThreadState` namespace in the first place and call them `kNotStarted`, `kStarting`, etc.?

core/utils/syscallthread.h
```diff
@@ -0,0 +1,443 @@
...
+  sigset_t allmask_;
+  sigset_t sigmask_;
+
+  // We only catch the exit signal once, process wide, so we want
+  // a flag to tell if we've done that.  Likewise, the per-thread
+  // signal mask is the same across all such threads, so we only
+  // need one instance.
+  static bool signals_initialized_;
+};
+
```
shinae-woo: We may move this class `ExitSigMask` into `syscallthread.cc`,  if this class is meant to be only accessible and visible to SyscallThread. Then inside .cc, this class can become simpler form as we don't need access control as well.

core/utils/syscallthread.h
sangjinhan: A new namespace with `enum class` sounds good.

The guide says `Ts` is right. My soul dies whenever I have to type `Tcp`...

core/utils/syscallthread.h
```diff
@@ -0,0 +1,443 @@
...
+        // This means that if it occurs in any other code path
+        // in the user's Run(), it will be deferred until the
+        // point at which the user calls pselect/ppoll.
+      } else {
+        // The user promises to call IsExitRequested() after
+        // each blocking system call.
+        sigdelset(&sigset, ExitSigMask::SIG_THREAD_EXIT);
+      }
+      ThreadSetSigmask(&sigset);
+
```
shinae-woo: May `ExitSigMask::allmask`, `ExitSigMask::sigmask` be used here for `sigset` as well?

core/utils/fifo_opener.cc
```diff
@@ -0,0 +1,304 @@
...
+    // While holding the lock (preventing other opens),
+    // mark "not open" and close any open FIFO.  Reset
+    // the thread so that new opens can happen if desired.
+    // Then we're done with this shutdown.
+    SetNewFdLocked(kNotOpen);
+    thread_.Reset();
+    shutting_down_ = false;
+  }
+}
+
```
shinae-woo: May I don't understand the code Push/PopDefer() clearly. Should the `PopDefer()` be called when shutdown?

chris3torek: It's not required, the push/pop is just in case there's some future need for this.  For instance, if you had a thread where you needed to make some system call (`socket`, `open`, whatever) that was blocking but should get interrupted, but then if that call _works_ you also need to make several more system calls that must not be interrupted, then go back into the interruptible loop, you could `PushDefer(); syscall1(arg); syscall2(arg); PopDefer(); continue;` for instance.

(In other words, right now this is just over-engineered 😈 but since they're inline it should not be adding any actual runtime overhead)

core/utils/syscallthread.h
chris3torek: This is a good idea, if I can get it to work out.

core/utils/syscallthread.h
chris3torek: I think so, this was what's left after I consolidated some earlier versions that were more scattered (which is also why the class is in the header and not in the `.cc` file).

chris3torek:
Reworked per @shinae-woo comments, fixing a few minor `const` issues too; and I tweaked the fifo handler slightly in a new commit.

chris3torek:
(I missed one `override` keyword, which clang was strict about - fixed, re-rebased, etc)

chris3torek:
The remaining failures are because I have run into a somewhat obscure rule about not calling virtual functions from destructors. I will have to re-work this yet again to work around this rule.

shinae-woo:
Should someone confirm this PR? @sangjinhan @melvinw

sangjinhan:
Merged, thanks!
tumblr_m6m622ixbr1qjev1to1_1280

core/utils/syscallthread.cc
```diff
@@ -0,0 +1,178 @@
...
+ * Waits for the knock thread to exit.
+ *
+ * Note that WaitForKnockThread is called after the regular thread
+ * is join()ed, so by definition the knock thread doesn't need to
+ * run any more.
+ */
+void SyscallThreadAny::WaitForKnockThread() {
+  if (knock_thread_.joinable()) {
+    // Kick the knock thread now to make it finish early if it's
+    // stuck in nanosleep (see syscallthread.cc).
```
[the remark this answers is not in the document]
chris3torek: I've never gotten doxygen to run cleanly and I'm not entirely sure what should be doxygen-ed.  I'd be happy enough to just use // comments everywhere...

(That comment, and code, used to be in the header, but it moved to the `.cc` file 😺 )

PR #745, opened 2017-12-05
List gate hooks 

pxgao:
To know the current BESS gate hook state, I added a grpc called ListGateHooks

### DIFF
--- a/protobuf/bess_msg.proto
+++ b/protobuf/bess_msg.proto
@@ -494,26 +494,23 @@ message PcapngArg {
   bool reconnect = 7; /// If set, we'll reconnect after failure.
 }
 
-message ConfigureGateHookRequest {
+
+message GateHookInfo {
   string hook_name = 1;         /// Name of the hook
   string module_name = 2;       /// Name of module
-  bool enable = 3;              /// Installs the hook if True, else uninstalls
   oneof gate {
-    int64 igate = 4;            /// Input gate index. All input gates if -1
-    int64 ogate = 5;            /// Output gate index. All output gates if -1
+    int64 igate = 3;            /// Input gate index. All input gates if -1
+    int64 ogate = 4;            /// Output gate index. All output gates if -1
   }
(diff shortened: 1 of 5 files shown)


pxgao:
Rebased to chris3torek:fifo

protobuf/bess_msg.proto
```diff
@@ -522,8 +519,8 @@ message ListGateHooksResponse {
 // Enable/Disable a global ResumeHook.
 message ConfigureResumeHookRequest {
   string hook_name = 1;         /// Name of the hook
-  bool enable = 2;              /// Installs the hook if True, else uninstalls
-  google.protobuf.Any arg = 3;  /// Hook-specific arguments
+  bool enable = 5;              /// Installs the hook if True, else uninstalls
+  google.protobuf.Any arg = 6;  /// Hook-specific arguments
 }
```
shinae-woo: Is there any reason for changing the field number in a way that not sequential? I do not exactly know how internal gRPC works, but it should probably fine as they are unique. Just very personal curious.

pxgao: Nice catch. I deleted a field. I will make it sequential.

shinae-woo: No worry!

pxgao: Actually its because I patched from the pangolin bess_msg.proto, which is wrong. That was caused by someone else. I fixed both anyway.

chris3torek: There's a gap left over in some protobufs in bess from when those field numbers meant something else. The general idea is that if you delete something you might leave in a comment, something like "// 4: was foo, now reserved" and avoid re-using it so that anyone still using an old protobuf definition will send field 4 which you'll ignore, rather than using as something new.

We haven't been very good about this. 😺

sangjinhan:
LGTM. Rebased to the master branch.

sangjinhan:
Thanks!

tumblr_m5ai413urg1qzex9io1_1280

PR #747, opened 2017-12-08
SSE4_2 has become the minimal requirements as with DPDK

shinae-woo:
SSE4.2 has become the minimal requirements for DPDK. So as to for BESS.

### DIFF
--- a/core/utils/http_parser.cc
+++ b/core/utils/http_parser.cc
@@ -29,12 +29,10 @@
 #include <assert.h>
 #include <stddef.h>
 #include <string.h>
-#ifdef __SSE4_2__
 #ifdef _MSC_VER
 #include <nmmintrin.h>
 #else
 #include <x86intrin.h>
-#endif
 #endif
 
 /* $Id$ */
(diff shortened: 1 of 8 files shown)


core/utils/simd.h
```diff
@@ -41,8 +41,8 @@
 #define __ymm_aligned __attribute__((aligned(32)))
 #define __zmm_aligned __attribute__((aligned(64)))
 
-#if !__SSSE3__
-#error CPU must be at least Core 2 or equivalent (SSSE3 required)
+#if !__SSE4_2__
+#error CPU must be at least corei5 or equivalent (SSE4.2 required)
```
sangjinhan: I think `Intel Nehalem` is a more accurate description than `corei5` (i.e., all Core i3 processors also support SSE4_2)

core/utils/http_parser.cc
```diff
@@ -69,29 +67,29 @@
...
-          *ret = -1;                                                          \
-          return NULL;                                                        \
-        }                                                                     \
-      }                                                                       \
-      ++buf;                                                                  \
-      CHECK_EOF();                                                            \
-    }                                                                         \
-    tok = tok_start;                                                          \
-    toklen = buf - tok_start;                                                 \
+#define ADVANCE_TOKEN(tok, toklen)                                       \
```
sangjinhan: Perhaps it's better to minimize changes (except bug patches) of this file, which was copied from some external open-source projects? Deviation from the original code may make code maintenance later.

sangjinhan:
Note: the SSE4.2 requirement is not yet strictly required for the current version of BESS, but we are planning to update to DPDK 17.11 soon.

core/utils/http_parser.cc
shinae-woo: I see. I will roll-back the changes for this file.

core/utils/simd.h
shinae-woo: Oh, right.

sangjinhan:
Thanks!
tumblr_m94zyfvd6f1qzbvzho1_500

PR #748, opened 2017-12-08
BESSD should not exit(), when failed to load modules

shinae-woo:
Bessd doesn't need to exit when faiedl to load external plugins.


sangjinhan:
tumblr_m99bjtz3aa1qejbiro1_1280

PR #749, opened 2017-12-12
Update DPDK version to 17.11

shinae-woo:
* libnuma becomes mandatory
* port_id becomes to 16 bits
* Extending supports for various RX offloads and associated API changes
* A few minior API changes


core/Makefile
```diff
@@ -83,7 +83,7 @@ endif
 # user requests a static build.
 ALWAYS_DYN_LIBS := -lpthread -ldl
 # These libraries are not supported by pkg-config.
-ALWAYS_LIBS := -lpcap -lgflags
+ALWAYS_LIBS := -lpcap -lgflags -lnuma
 # If pkg-config is available, we just need a list of the dependecies.
```
shinae-woo: @ddiproietto  Could you confirm this change are appropriate?

shinae-woo:
The container should be re-build. ❓ ❓

core/Makefile
ddiproietto: The change makes sense, thanks

shinae-woo:
@sangjinhan Thank you for building new containers.

shinae-woo:
@ddiproietto Thank you for the review! I will fix that point!

sangjinhan:
In `core/modules/ip_lookup.cc`, there is some temporary code to work around an issue of DPDK 17.02. Given that the issue has gone in the new version, could you clean up the hack I made in the file?

shinae-woo:
@sangjinhan This is a very weird bug on g++ side... Anyway, DPDK 17.11 have the similar workaround codes inside `rte_lpm_lookupx4`, so I remove the temporary code from the module code.

shinae-woo:
image

sangjinhan:
That's the code I pushed to the DPDK upstream 😸

sangjinhan:
Merged, thanks!
tumblr_m83dqfx91m1qzex9io1_1280

PR #751, opened 2017-12-14
Add delete command for IPLookup module

levaitamas:
Mostly done as cargo cult programming, but it works for me.
Comments welcome.

### DIFF
--- a/core/modules/ip_lookup.cc
+++ b/core/modules/ip_lookup.cc
@@ -149,6 +149,8 @@ static inline int is_valid_gate(gate_idx_t gate) {
 const Commands IPLookup::cmds = {
     {"add", "IPLookupCommandAddArg", MODULE_CMD_FUNC(&IPLookup::CommandAdd),
      Command::THREAD_UNSAFE},
+    {"delete", "IPLookupCommandDeleteArg", MODULE_CMD_FUNC(&IPLookup::CommandDelete),
+     Command::THREAD_UNSAFE},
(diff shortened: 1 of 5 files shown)


core/modules/ip_lookup.cc
```diff
@@ -302,6 +304,47 @@ CommandResponse IPLookup::CommandAdd(
...
+                          arg.prefix().c_str());
+  }
+
+  uint64_t prefix_len = arg.prefix_len();
+  if (prefix_len > 32) {
+    return CommandFailure(EINVAL, "Invalid prefix length: %" PRIu64,
+                          prefix_len);
+  }
+
+  net_mask = be32_t(~((1ull << (32 - prefix_len)) - 1));
```
melvinw: nit: Could you use the [SetBitsLow()](https://github.com/NetSys/bess/blob/master/core/utils/bits.h#L193) utility here instead? There's an example of its usage with the big-endian types [here](https://github.com/NetSys/bess/blob/master/core/utils/ip.cc#L78)

sangjinhan:
LGTM. Perhaps you can refactor the common code in `Add` and `Delete` regarding argument processing?

levaitamas:
I tried to address  @melvinw's and @sangjinhan's  comments with a new commit. Thanks.

sangjinhan:
Merged, thanks!
tumblr_m48dsbx5rt1qz4ueho1_1280

PR #753, opened 2017-12-21
Use lookup table in Rewrite to solve unfairness

sangjinhan:
... among templates. Also see #436

### DIFF
--- a/core/modules/rewrite.cc
+++ b/core/modules/rewrite.cc
@@ -75,6 +75,7 @@ CommandResponse Rewrite::CommandAdd(const bess::pb::RewriteArg &arg) {
     size_t j = i % num_templates_;
     bess::utils::Copy(templates_[i], templates_[j], template_size_[j]);
     template_size_[i] = template_size_[j];
+    jump_[i] = j;
   }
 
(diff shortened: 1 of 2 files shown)


core/modules/rewrite.cc
```diff
@@ -119,10 +120,7 @@ inline void Rewrite::DoRewrite(bess::PacketBatch *batch) {
     bess::utils::CopyInlined(ptr, templates_[start + i], size, true);
   }
 
-  next_turn_ = start + cnt;
-  if (next_turn_ >= bess::PacketBatch::kMaxBurst) {
```
melvinw: I'm fine with this approach, but could the unfairness may actually have come from the fact that we compare `next_turn_` to `PacketBatch::kMaxBurst` here instead of `kNumSlots`, which is set to `bess::PacketBatch::kMaxBurst * 2 - 1`?

sangjinhan: We can't compare `next_turn_` with `kNumSlots` here; if it is larger, then it implies out-of-bound access happened in the for loop.

melvinw:
Thanks! cat

PR #754, opened 2017-12-27
minor histogram changes

chris3torek:
NB: I originally put in a static assert with is_integral, and
then histogram_test.cc failed.  I'm not sure which one we should fix!

---

While examining the measure module I noted a few things
to improve slightly in the histogram code:

 - Comments say that the type must be integral.  It probably should
   be but at least one test uses double, and any arithmetic type
   will work up to the point where it runs out of bits, at least.
   Adjust the comment and assert that the type is arithmetic.  If
   we want to constrain the type to integral, we'll need to fix
   the test.

 - The Reset() function constructs an all-new vector of the type,
   to reset to all-zero values.  We can just memset() the data now
   that we have C++11 which gives us a vector .data() operator,
   since we only run on machines with all-zero-bits = 0 (integral)
   or 0.0 (float/double).

### DIFF
--- a/core/utils/histogram.h
+++ b/core/utils/histogram.h
@@ -33,18 +33,21 @@
 
 #include <cstdint>
 #include <cstdlib>
+#include <cstring>
 #include <limits>
 #include <vector>
 
 #include <glog/logging.h>
 
-// Class for general purpose histogram. T must be an integral type.
+// Class for general purpose histogram. T generally should be an
+// integral type, though floating point types will also work.
 // A bin b_i corresponds for the range [i * width, (i + 1) * width)
 // (note that it's left-closed and right-open), and (i * width) is used for its
 // representative value.
 template <typename T = uint64_t>
 class Histogram {
  public:
+  static_assert(std::is_arithmetic<T>::value, "Arithmetic type required.");
   struct Summary {
     size_t count;        // # of all samples. If 0, min, max and avg are also 0
     size_t above_range;  // # of samples beyond the histogram range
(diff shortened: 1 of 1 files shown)


core/utils/histogram.h
```diff
@@ -131,9 +134,12 @@ class Histogram {
     return ret;
   }
 
+  // Resets all counters, and the count of such counters, to zero.
+  // Note that the number of buckets remains unchanged.  We assume
+  // memset-to-zero works for all arithmetic types.
   void Reset() {
     count_ = 0;
-    buckets_ = std::vector<size_t>(buckets_.size());
+    memset(buckets_.data(), 0, buckets_.size() * sizeof(T));
```
barath: Why not the more C++ style approach?

`std::fill(buckets_.begin(), buckets_.end(), 0);`

My understanding is that it will be optimized to be roughly the same performance.

chris3torek: If it does optimize properly, that would be better.  I'll check it out.  It's not very important since bess doesn't use the Reset() function anyway though...

chris3torek:
Switched to `std::fill`.  Testing shows that both g++ and clang++ emit tail calls to `memset` here.  Curiously, both of them check for an empty vector first, rather than blindly calling `memset`, perhaps to allow for the pointer to be invalid if the size is zero.

(I'm still not sure if the test that uses `double` is OK, or if this really should be restricted to integer types.)

chris3torek:
BTW, if this looks reasonable, the next step is to make a split-up "prepare reset"/"do reset as fast as possible"/"finish reset" sequence, so that BESS's internal speed hack using `std::move` can be hidden away.  BESS is trying to avoid an expensive operation (big allocation or memset) while holding a spinlock, in the `Measure` module `Clear` function.  It currently makes a second histogram and then assigns using `std::move`, which makes the internal vector use a move-and-free.  We could make it use `std::swap` instead, perhaps (and otherwise be pretty similar to how it's set up now).

barath:
That seems good to me, though Melvin or Sangjin may have more to say about
it.

PR #755, opened 2018-01-09
Task to context

shinae-woo:
With this PR, modules do not access directly into the thread-local storage, but modules can access the current execution contexts 'Context *' such as task, worker id,  current cycles/time, etc.

!! This PR should be merged with #739. This requires changing module interface as well. So I would recommend deferring merging #739 until this PR is approved and is ready to merge.

### DIFF
--- a/core/module.h
+++ b/core/module.h
@@ -529,24 +529,26 @@ inline void Module::EmitPacket(const Task *task, bess::Packet *pkt,
 
   // Put a packet into the ogate
   bess::OGate *ogate = ogates_[ogate_idx];
+  bess::IGate *igate = ogate->igate();
   bess::PacketBatch *batch = task->get_gate_batch(ogate);
   if (!batch) {
     if (!ogate->hooks().empty()) {
       // Having separate batch to run ogate hooks
-      task->set_gate_batch(ogate, task->AllocPacketBatch());
+      batch = task->AllocPacketBatch();
+      task->set_gate_batch(ogate, batch);
       gate_with_hook_[gate_with_hook_cnt_++] = ogate_idx;
     } else {
       // If no ogate hooks, just use next igate batch
-      if (task->get_gate_batch(ogate->igate()) == nullptr) {
-        bess::PacketBatch *tmp = task->AllocPacketBatch();
-        task->AddToRun(ogate->igate(), tmp);
-        task->set_gate_batch(ogate, tmp);
+      batch = task->get_gate_batch(igate);
+      if (batch == nullptr) {
+        batch = task->AllocPacketBatch();
+        task->AddToRun(igate, batch);
+        task->set_gate_batch(ogate, batch);
       } else {
-        task->set_gate_batch(ogate, task->get_gate_batch(ogate->igate()));
+        task->set_gate_batch(ogate, task->get_gate_batch(igate));
       }
       gate_without_hook_[gate_without_hook_cnt_++] = ogate_idx;
     }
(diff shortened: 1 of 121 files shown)


core/worker.cc
```diff
@@ -300,7 +300,7 @@ void *Worker::Run(void *_arg) {
 
   current_tsc_ = rdtsc();
 
-  pframe_pool_ = bess::get_pframe_pool();
+  pframe_pool_ = bess::get_pframe_pool_socket(socket_);
```
chris3torek: Should this be a separate thing? It's good to be NUMA aware here but isn't that independent of the thread-local cleanup...?

core/module.h
```diff
@@ -53,6 +52,24 @@ using bess::gate_idx_t;
 #define MAX_TASKS_PER_MODULE 32
 #define UNCONSTRAINED_SOCKET ((0x1ull << MAX_NUMA_NODE) - 1)
 
+struct Context {
```
chris3torek: I wonder if we should make this more strict about access to internal fields (using class instead of struct, and private or protected members with acccessors).  I'm not exactly militant about it though, perhaps from too much old school C programming 😈

core/module.h
```diff
@@ -53,6 +52,24 @@ using bess::gate_idx_t;
...
+  // Set by task scheduler, read by modules
+  uint64_t current_tsc;
+  uint64_t current_ns;
+  int wid;
+  Task *task;
+
+  // Set by module scheduler, read by a task scheduler
+  uint64_t silent_drops;
+
+  // Temproray variables to be accessed and updated by module scheduler
```
chris3torek: (trivial typo in comment)

core/module.h
shinae-woo: Definitely, I can change it to class.

core/modules/drr.cc
```diff
@@ -260,14 +261,14 @@ DRR::Flow* DRR::GetNextFlow(int* err) {
   return f;
 }
 
-uint32_t DRR::GetNextPackets(bess::PacketBatch* batch, Flow* f, int* err) {
+uint32_t DRR::GetNextPackets(bess::PacketBatch *batch, Flow *f, int *err) {
   uint32_t total_bytes = 0;
-  bess::Packet* pkt;
+  bess::Packet *pkt;
```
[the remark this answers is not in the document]
shinae-woo: No... I don't know why yet, I haven't investigated much..

core/modules/drr.h
```diff
@@ -45,34 +45,34 @@
...
-  * PARAMETERS:
-  *    * quantum: the number of bytes allocated to each flow on each round
-  *    * Max Number of flows: max number of flows the module will handle
-  *    * Max Flow Queue Size: the maximum size that any Flows queue can get
-  *          before the module will start dropping the flows packets
-  * COMMANDS
-  *    update quantum: cannot not be done live
-  *    update Max Flow Queue Size: can be done live
-*/
+//
```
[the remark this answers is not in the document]
shinae-woo: Yes :)

core/modules/drr.cc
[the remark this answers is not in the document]
sangjinhan: clang-format is fine with either style, but it tries to make all consistent. The current majority style is chosen, so it will depend on which one is majority.

sangjinhan:
Merged to the `scheduler_1103` branch. Thanks!

cute cat 16

Issue #756, opened 2018-01-09
compile issue


ziruizhuang:
A strange problem occurred during building process on master branch.

**If using gcc 6 or 7**
```
user@user-Lenovo-PC:~/bess$ ./build.py 
Building DPDK...
Generating protobuf codes for pybess...
Building BESS daemon...
Log:
 make: Entering directory '/home/user/bess/core'
[CXX]       packet.o
[CXX]       main.o
[CXX]       debug.o
[CXX]       gate_hooks/tcpdump.o
[CXX]       gate_hooks/track.o
[CXX]       gate_hooks/pcapng.o
[CXX]       resume_hooks/task_graph.o
[CXX]       resume_hooks/metadata.o
debug.cc:397:10: error: declaration of ‘struct bess::debug::TrapHandler(int, siginfo_t*, void*)::ucontext’ shadows a parameter [-Werror=shadow]
   struct ucontext *uc;
          ^~~~~~~~
debug.cc:395:61: note: shadowed declaration is here
 static void TrapHandler(int sig_num, siginfo_t *info, void *ucontext) {
                                                             ^~~~~~~~
debug.cc: In function ‘void bess::debug::TrapHandler(int, siginfo_t*, void*)’:
debug.cc:405:35: error: expected primary-expression before ‘;’ token
   uc = (struct ucontext *)ucontext;
                                   ^
debug.cc:410:40: error: invalid use of incomplete type ‘struct bess::debug::TrapHandler(int, siginfo_t*, void*)::ucontext’
   trap_ip = reinterpret_cast<void *>(uc->uc_mcontext.gregs[REG_RIP]);
                                        ^~
debug.cc:397:10: note: forward declaration of ‘struct bess::debug::TrapHandler(int, siginfo_t*, void*)::ucontext’
   struct ucontext *uc;
          ^~~~~~~~
debug.cc:395:61: error: unused parameter ‘ucontext’ [-Werror=unused-parameter]
 static void TrapHandler(int sig_num, siginfo_t *info, void *ucontext) {
                                                             ^~~~~~~~
debug.cc: In function ‘void bess::debug::GoPanic()’:
debug.cc:390:1: error: ‘noreturn’ function does return [-Werror]
 }
 ^
cc1plus: all warnings being treated as errors
Error: debug.o
[...]

PR #757, opened 2018-01-10
Use ucontext_t in place of ucontext

sangjinhan:
debug.cc fails to build on Ubuntu 17.10, where its ucontext.h file does not declare `struct ucontext` any longer. Use typedef'ed `ucontext_t` instead. Reported by #756.


ddiproietto:
gatto

PR #760, opened 2018-01-19
Gatehookcmds

chris3torek:
This is a fairly complicated series of 9 commits that:

 * cleans up a few minor bugs here and there
 * reorganizes the code to instantiate gate hooks
 * adds the concept of a "gate hook command"
 * adds a sample gate hook command ("reset" to the "track" hook)
 * adds code to invoke the track hook reset from bessctl

(but there is no test for this reset, and I should probably write one).

It may make sense to collapse the last three commits into one, but the rest probably should stay independent since they are a good idea whether or not gate hook commands are _also_ a good idea... :smiley_cat:

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -1349,6 +1349,43 @@ class BESSControlImpl final : public BESSControl::Service {
     return Status::OK;
   }
 
+  Status GateHookCommand(ServerContext*, const GateHookCommandRequest* request,
+                         CommandResponse* response) override {
+    // No need to look up the hook factory: the gate either
+    // has a hook instance with the right name, or doesn't.
+    const bess::pb::GateHookInfo& rh = request->hook();
+    const auto& it = ModuleGraph::GetAllModules().find(rh.module_name());
+    if (it == ModuleGraph::GetAllModules().end()) {
+      return return_with_error(response, ENOENT, "No module '%s' found",
+                               rh.module_name().c_str());
+    }
+    Module* m = it->second;
+    bool is_igate = rh.gate_case() == bess::pb::GateHookInfo::kIgate;
+    gate_idx_t gate_idx = is_igate ? rh.igate() : rh.ogate();
+    bess::Gate* g = module_gate(m, is_igate, gate_idx);
+    if (m == nullptr) {
+      return return_with_error(
+          response, EINVAL, "%s: %cgate '%hu' does not exist",
+          m->name().c_str(), is_igate ? 'i' : 'o', gate_idx);
+    }
+
+    bess::GateHook* hook = g->FindHook(rh.hook_name());
+    if (hook == nullptr) {
+      return return_with_error(response, ENOENT,
+                               "%s: %cgate '%hu' has no hook named '%s'",
+                               m->name().c_str(), is_igate ? 'i' : 'o',
+                               gate_idx, rh.hook_name().c_str());
+    }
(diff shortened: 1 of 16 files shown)


core/bessctl.cc
```diff
@@ -91,46 +91,41 @@ static inline Status return_with_errno(T* response, int code) {
...
+      return CommandFailure(EINVAL, "'%s': %cgate '%hu' does not exist",
+                            m->name().c_str(), is_igate ? 'i' : 'o', gate_idx);
+    }
+    bess::GateHook* hook = factory.CreateGateHook();
+    CommandResponse init_ret = factory.InitGateHook(hook, gate, arg);
+    if (init_ret.error().code() != 0) {
+      delete hook;
+      return init_ret;
+    }
+    if ((ret = gate->AddHook(hook))) {
```
barath: Is `ret` used anywhere?  If not, maybe this assignment in `if` is unnecessary.

core/gate.cc
```diff
@@ -124,7 +124,26 @@ void IGate::RemoveOgate(const OGate *og) {
...
 // Add internally-generated Track() hook to this ogate.
 void OGate::AddTrackHook() {
-  this->AddHook(new Track());
+  static const GateHookFactory *track_factory;
+  static google::protobuf::Any arg;
+
+  // If we haven't located the track hook factory yet, do that first.
+  if (track_factory == nullptr) {
+    const auto it =
+        bess::GateHookFactory::all_gate_hook_factories().find(Track::kName);
```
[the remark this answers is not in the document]
chris3torek: Well, we're in the same source file so we could cheat and access the variable directly, but that's more or less the same as what `bessctl.cc` was doing.

core/gate.h
```diff
@@ -150,8 +144,9 @@ class Gate {
 
   const std::vector<GateHook *> &hooks() const { return hooks_; }
 
-  // Inserts hook in priority order and returns 0 on success.
-  int AddHook(GateHook *hook);
+  // Creates, initializes, and then inserts gate hook in priority order.
```
barath: Given that `gate` is modified by this method, perhaps it should be the last arg?

chris3torek: I don't have any strong feeling either way. Do we have a convention? The Google style sheet doesn't seem to say anything about modified vs unmodified, except that `T&` requires the `const`, so that if we're modifying the thing we have to use `T*` rather than `T&`.

barath: It's minor, so fine with me.  The style guide does say outputs come last, though whether this is an "output" is a little ambiguous:

https://google.github.io/styleguide/cppguide.html#Function_Parameter_Ordering

core/gate.cc
```diff
@@ -124,7 +124,26 @@ void IGate::RemoveOgate(const OGate *og) {
...
+    if (it == bess::GateHookFactory::all_gate_hook_factories().end()) {
+      DCHECK(0) << "track gate hook factory is missing";
```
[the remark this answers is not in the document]
chris3torek: BTW I now think I had `CHECK` initially and it failed to compile as there was no ostream formatting available for at least one argument. Maybe I can just use `<< ...` to override the need for it...

chris3torek:
Updated to use `CHECK(0)`. Using `CHECK_NE` fails as the map iterator and/or `.end()` does not support `operator<<` on an ostream.

PR #761, opened 2018-01-20
pybess: Fix _configure_gate_hook()

ddiproietto:
The `ConfigureGateHook` protobuf message has changed: update its usage.
This fixes tcpdump in bessctl.

Reported by @gatanu


ddiproietto:
I should point out that the exception generated by `protobuf_to_dict()` was not shown because bessctl was stuck here:

https://github.com/NetSys/bess/blob/ed9c01683f2cc646749692c78d12cf4c3254639f/bessctl/commands.py#L1880

PR #762, opened 2018-01-27
Make BESS easier to test without root

ddiproietto:
I'd like to use BESS in a set of tests that could be run in parallel and without root privileges.

* As long as there are no PMD devices and hugepages, bessd should work as non root, so this PR adds a command line flags to skip the root check.

To be able to run in parallel, I'd like each BESS instance to listen on a unix domain socket. gRPC makes it easy to do that, so:

* This commits adds a `--grpc_url` option to bessd. the `-p` and `-b` options are kept for backward compatibility.
* The bessctl `daemon connect` command is changed: it takes a grpc URL instead of the host and the port, to make it more generic. This is _not_ backward compatible

### DIFF
--- a/pybess/bess.py
+++ b/pybess/bess.py
@@ -152,6 +152,7 @@ class BESS(object):
         pass
 
     DEF_PORT = 10514
+    DEF_GRPC_URL = "localhost:" + str(DEF_PORT)
     BROKEN_CHANNEL = "AbnormalDisconnection"
 
     def __init__(self):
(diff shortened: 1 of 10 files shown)


pybess/bess.py
```diff
@@ -182,16 +183,16 @@ def _update_status(self, connectivity):
         else:
             self.status = connectivity
 
-    def connect(self, host='localhost', port=DEF_PORT):
+    def connect(self, grpc_url=DEF_GRPC_URL):
```
chris3torek: Aha, this causes test failures in pybess/test_bess.py (but that seems to be the only user of the old `port=` argument so we should just fix that one up).

ddiproietto: Oops, thanks for figuring that out, I will fix it

sangjinhan:
Thanks! (sorry, million years late)
tumblr_m1nxtehynk1qjc1a7o1_1280

PR #764, opened 2018-02-01
Add support for initial buffers to allocate, allocate memory properly with no_huge

nrdmao:
Add a -buffers argument that specifies the initial number of buffers to allocate per
numa socket.

Properly allocate mb_per_socket when specifying -no-huge by passing
-m (mb_per_socket * numa_count) to rte_eal_init().

### DIFF
--- a/core/opts.cc
+++ b/core/opts.cc
@@ -35,6 +35,7 @@
 
 #include "bessd.h"
 #include "worker.h"
+#include "packet.h"
 
 // Port this BESS instance listens on.
 // Panda came up with this default number
@@ -95,17 +96,22 @@ static const bool _m_dummy[[maybe_unused]] =
     google::RegisterFlagValidator(&FLAGS_m, &ValidateMegabytesPerSocket);
 
 static bool ValidateBuffersPerSocket(const char *, int32_t value) {
-  if (value <= 0) {
-    LOG(ERROR) << "Invalid buffer count: " << value;
+  if (value < bess::minimum_try) {
+    LOG(ERROR) << "Invalid buffer count: " << value <<
+      " must be >= " << bess::minimum_try;
     return false;
   }
-  if (value % 512 != 0) {
-    LOG(ERROR) << "Number of buffers must be a multiple of 512: " << value;
-    return false;
+  int32_t check = value;
+  while (check != 1) {
+    if (check % 2 != 0) {
+      LOG(ERROR) << "Number of buffers must be a power of 2: " << value;
+      return false;
+    }
+    check = check / 2;
   }
   return true;
 }
 DEFINE_int32(buffers, 262144, "Specifies how many buffers to allocate per socket,"
-	     " must be a multiple of 512");
+	     " must be a power of 2.");
 static const bool _buffers_dummy[[maybe_unused]] =
     google::RegisterFlagValidator(&FLAGS_buffers, &ValidateBuffersPerSocket);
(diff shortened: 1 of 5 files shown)


core/opts.cc
```diff
@@ -93,3 +93,19 @@ static bool ValidateMegabytesPerSocket(const char *, int32_t value) {
 DEFINE_int32(m, 1024, "Specifies how many megabytes to use per socket");
 static const bool _m_dummy[[maybe_unused]] =
     google::RegisterFlagValidator(&FLAGS_m, &ValidateMegabytesPerSocket);
+
+static bool ValidateBuffersPerSocket(const char *, int32_t value) {
+  if (value <= 0) {
+    LOG(ERROR) << "Invalid buffer count: " << value;
+    return false;
+  }
+  if (value % 512 != 0) {
```
chris3torek: The comments in `packet.cc` suggest that the `rte_mempool_create` argument should be (2**n)-1 for some integral n, so perhaps we should check "is power of 2" here instead (so that dividing by 2 leaves inputs as a power of 2)? I'll leave more analysis to @sangjinhan but other than this, things seem OK here.

(Note that another option would be to round down to the nearest power of two in `packet.cc` after each divide-by-2.)

nrdmao: Makes sense. I will post an update.

core/packet.h
```diff
@@ -60,6 +60,8 @@ static_assert(SNBUF_SCRATCHPAD_OFF == 320,
 
 namespace bess {
 
+const int minimum_try = 16384; // For packet buffer allocation.
```
sangjinhan: According to the Google style guide it should be renamed something like `kMinimumTry`. However, I have a different suggestion:
* This `minimum_try` stays in the `init_mempool_socket()`
* `ValidateBuffersPerSocket()` does not check if the value is no less than `minimum_try`.
* `init_mempool_socket()` remains the same. The resulting behavior is that the user specified value will be tried once, but not any further if it is less than `minimum_try`.

core/opts.cc
```diff
@@ -93,3 +94,24 @@ static bool ValidateMegabytesPerSocket(const char *, int32_t value) {
...
+  while (check != 1) {
+    if (check % 2 != 0) {
+      LOG(ERROR) << "Number of buffers must be a power of 2: " << value;
+      return false;
+    }
+    check = check / 2;
+  }
+  return true;
+}
+DEFINE_int32(buffers, 262144, "Specifies how many buffers to allocate per socket,"
```
sangjinhan: `buffers` -> `packet buffers`?

core/opts.cc
```diff
@@ -93,3 +94,24 @@ static bool ValidateMegabytesPerSocket(const char *, int32_t value) {
...
 static const bool _m_dummy[[maybe_unused]] =
     google::RegisterFlagValidator(&FLAGS_m, &ValidateMegabytesPerSocket);
+
+static bool ValidateBuffersPerSocket(const char *, int32_t value) {
+  if (value < bess::minimum_try) {
+    LOG(ERROR) << "Invalid buffer count: " << value <<
+      " must be >= " << bess::minimum_try;
+    return false;
+  }
+  int32_t check = value;
```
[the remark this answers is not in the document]
nrdmao: Will do.

core/opts.cc
nrdmao: OK. will change.

core/packet.h
nrdmao: Will do.

sangjinhan:
Thank you!
6k3k26y647ssor4txbenirb0_500

PR #766, opened 2018-02-02
Add a command line option to control core dump generation

nrdmao:
Add a command line option to force bessd to drop a core file on a fatal condition.
When the this option is set, do not generate /tmp/bessd_crash.log.

### DIFF
--- a/core/debug.cc
+++ b/core/debug.cc
@@ -380,13 +380,15 @@ static bool SkipSymbol(char *symbol) {
   if (oops_msg == "")
     oops_msg = DumpStack();
 
-  // Create a crash log file
-  try {
-    std::ofstream fp(P_tmpdir "/bessd_crash.log");
-    fp << oops_msg;
-    fp.close();
-  } catch (...) {
-    // Ignore any errors.
+  // Create a crash log file if we are not dropping a core file.
+  if (!FLAGS_core_dump) {
+    try {
+      std::ofstream fp(P_tmpdir "/bessd_crash.log");
+      fp << oops_msg;
+      fp.close();
+    } catch (...) {
+      // Ignore any errors.
+    }
   }
 
   // Set SIGABRT back to the default to avoid catching the abort used to
--- a/core/debug.cc
+++ b/core/debug.cc
@@ -56,6 +56,7 @@
 #include "packet.h"
 #include "scheduler.h"
 #include "traffic_class.h"
+#include "opts.h"
 #include "utils/format.h"
 
 namespace bess {
@@ -371,6 +372,10 @@ static bool SkipSymbol(char *symbol) {
   _exit(EXIT_FAILURE);
 }
 
+[[noreturn]] static void abort_failure() {
+  abort();
+}
+
 [[ gnu::noinline, noreturn ]] void GoPanic() {
   if (oops_msg == "")
     oops_msg = DumpStack();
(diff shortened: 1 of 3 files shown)


core/debug.cc
```diff
@@ -371,21 +372,37 @@ static bool SkipSymbol(char *symbol) {
...
 
-  // Create a crash log file
-  try {
-    std::ofstream fp(P_tmpdir "/bessd_crash.log");
-    fp << oops_msg;
-    fp.close();
-  } catch (...) {
-    // Ignore any errors.
+  // Create a crash log file if we are not dropping a core file.
+  if (!FLAGS_core_dump) {
```
sangjinhan: I guess the new behavior here is to either create a crashlog file or exit with `abort()`. I think these two options are not really exclusive to each other, i.e., one may want to create a core dump yet still have a crashlog file as well. Maybe I am misunderstood but this `if` statement looks unnecessary.

core/debug.cc
```diff
@@ -371,21 +372,37 @@ static bool SkipSymbol(char *symbol) {
...
-  google::InstallFailureFunction(exit_failure);
+  // Set SIGABRT back to the default to avoid catching the abort used to
+  // generate the core file.
+  struct sigaction sa;
+  memset(&sa, 0, sizeof(sa));
+  sa.sa_handler = SIG_DFL;
+  sigaction(SIGABRT, &sa, 0);
+
+  if (FLAGS_core_dump) {
+    google::InstallFailureFunction(abort_failure);
```
sangjinhan: If I remember correctly, this line is somewhat redundant (`LOG(FATAL) << ...` below will invoke `abort()` eventually). But this explicit code looks better.

core/debug.cc
```diff
@@ -371,21 +372,37 @@ static bool SkipSymbol(char *symbol) {
...
+    } catch (...) {
+      // Ignore any errors.
+    }
   }
 
-  // The default failure function of glog calls abort(), which causes SIGABRT
-  google::InstallFailureFunction(exit_failure);
+  // Set SIGABRT back to the default to avoid catching the abort used to
+  // generate the core file.
+  struct sigaction sa;
```
sangjinhan: These lines can move into the `if` block below, since they are irrelevant for the `else` block?

core/debug.cc
nrdmao: Well you are right that I am tying together the the control of the crashlog file generation with the control of the core file. I suppose my real desire is to eliminate the creation of the crashlog file because: 1. It is already logged with LOG(FATAL) so it is superfluous. 2. Unlike glog, I can not control where is it dumped. I don't want my logs dumped to /tmp but prefer to dump them to /var/log/. With glog, this is controlled with an environment variable. 3. The crashlog is redundant with the core file.

Anyway, I am open to suggestions. I can either:
1. Add another command line argument to control whether or not the crash log is generated.
2. Document the -core-dump argument to indicate that when a core is dumped a crashlog file will not be generated.

core/debug.cc
nrdmao: Will do.

core/debug.cc
nrdmao: This is what was commented but if I did not call InstallFailureFunction with abort_failure I got some other strange behavior. Given that you like the explicit code anyway I will keep it as it is.

core/debug.cc
sangjinhan: I like the first option better. The argument could be a boolean or a string (path to the crashlog, empty string if disabled). Either is fine by me.

As you said the crashlog is quite superfluous. `bessctl` would check if the file exists and show its content when the grpc connection abruptly has been disconnected. I was too lazy to make it find and read the log file and show the last log messages, so created a new file.

nrdmao: I will add a boolean argument. I did not realize the relationship with bessctl. Thanks for the explanation.

sangjinhan:
jjkybd3nsagfo3jvslivsy5r_500

PR #770, opened 2018-02-08
Only erase members from event_modules PreResume set after iteration done

nrdmao:
It seems that while iterating through a set if you erase a member the iteration may not visit all the
members of the set. Issue found under the following conditions:

```
$ uname -a
Linux bess-ubuntu16 4.4.0-109-generic #132-Ubuntu SMP Tue Jan 9 19:52:39 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

$ g++ --version
g++ (Ubuntu 5.4.1-2ubuntu1~16.04) 5.4.1 20160904
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/core/resume_hook.cc
+++ b/core/resume_hook.cc
@@ -72,12 +72,16 @@ void run_global_resume_hooks(bool run_modules) {
   }
 
   if (run_modules) {
+    std::set<Module *> erase_modules; // Erase modules after completing iteration
     auto &resume_modules = event_modules[Event::PreResume];
     for (Module *m : resume_modules) {
       int ret = m->OnEvent(Event::PreResume);
       if (ret == -ENOTSUP) {
-        resume_modules.erase(m);
+        erase_modules.insert(m);
       }
+    }
+    for (Module *m : erase_modules) {
+      resume_modules.erase(m);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


core/resume_hook.cc
```diff
@@ -72,13 +72,17 @@ void run_global_resume_hooks(bool run_modules) {
   }
 
   if (run_modules) {
+    std::set<Module *> erase_modules; // Erase modules after completing iteration
     auto &resume_modules = event_modules[Event::PreResume];
     for (Module *m : resume_modules) {
       int ret = m->OnEvent(Event::PreResume);
       if (ret == -ENOTSUP) {
-        resume_modules.erase(m);
+        erase_modules.insert(m);
```
barath: In both of these cases, might it be simpler to use iterators rather than colon notation and a second `set`?  e.g., `resume_modules.erase(it++);` should work fine here.

chris3torek: (Oh, you, er, @barath, meant _ditch_ the range-based `for` loop.  I seem to be extra slow this morning...)

nrdmao: OK, will use iterators. Thanks.

chris3torek:
Merged, thanks!

image

PR #773, opened 2018-02-20
gate hook fixes

chris3torek:
This has three separate commits for three small bugs:

 * we couldn't build plugin gate hooks at all
 * we invoked gate hooks incorrectly when trying to use the "in" direction
 * we caught the wrong error (bad module name) instead of the right one (bad gate hook)

The last error was only triggered by the middle error, but we now have them all fixed.


chris3torek:
Something about the last commit (modifying `core/Makefile`) breaks the etherencap test.  It's still a mystery exactly what this is.

chris3torek:
Found the problem: two `metadata.cc` files.  Redid the third commit.

melvinw:
Thanks! cat

PR #774, opened 2018-02-21
Separate coverage and debug builds.

barath:
With this change, DEBUG only sets -O0 and COVERAGE turns on coverage and -O0.


melvinw:
Thanks! cat

PR #775, opened 2018-03-03
build: allow DPDK_REPO, DPDK_TAG to be overidden

melvinw:
git://dpdk.org/dpdk seems to be less flakey than http://dpdk.org/git/dpdk and it would be nice to be able to use it when the user's network allows for it.


sangjinhan:
qgib8terim4ulxoksj28a38jo1_400

PR #776, opened 2018-03-09
histogram: allow atomic and non-atomic usage

chris3torek:
Allow atomic operations so that a histogram can be used
without a lock.

For backwards compatibility, the default is to use non-atomic
operations, requiring that the caller use a lock if the histogram
will be maniuplated in multiple threads.

Note that internally, this changes the counters from the
machine-dependent size_t to uint64_t (which affects 32-bit
implementations).

### DIFF
--- a/core/utils/histogram.h
+++ b/core/utils/histogram.h
@@ -32,6 +32,7 @@
 #define BESS_UTILS_HISTOGRAM_H_
 
 #include <algorithm>
+#include <atomic>
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


core/utils/histogram.h
```diff
@@ -64,13 +65,60 @@ class Histogram {
...
+  // Swap operator allows clean summarizing with external lock,
+  // when using a histogram on an active data stream.
+  // Create a new, empty histogram of the appropriate size
+  // while NOT holding a lock, then use tmphist.swap(datahist)
+  // while holding a lock.  Release the lock, and summarize
+  // from tmphist while new data accumulate into datahist.
+  void swap(Histogram &other) noexcept {
+    using std::swap;
+    swap(bucket_width_, other.bucket_width_);
+    // atomic count_ field requires manual swap
```
sangjinhan: While it's not directly related to this PR; perhaps we can remove `count_` from the class entirely? It would reduce code verbosity regarding atomic access of the counter, also eliminating the non-atomicity between the counter and the actual buckets. The variable doesn't seem to serve anything useful (`Summarize()` can be modified to not rely on `count_` at all).

core/utils/histogram.h
```diff
@@ -147,24 +196,28 @@ class Histogram {
...
-    buckets_.clear();
-    buckets_.resize(num_buckets + 1);
+    buckets_ = std::move(std::vector<std::atomic<uint64_t>>(num_buckets + 1));
     bucket_width_ = bucket_width;
-    count_ = 0;
+    count_.store(0, std::memory_order_relaxed);
   }
 
  private:
+  void nonatomic_inc(std::atomic<uint64_t> *p) {
```
sangjinhan: minor nit: if we ditch the counter it would be better to embed the body of these functions in `(Atomic)Insert`.

core/utils/histogram.h
chris3torek: I like that idea, so I implemented it.  It simplifies the rest of the code a bunch.  See what you think...

chris3torek:
(rebased and added a fixup commit, to be squashed before merging)

sangjinhan:
Merged, thanks.
(Forgot to squash. Ugh...)

tumblr_m4814clzru1qbe5pxo1_1280

PR #778, opened 2018-03-17
Make protobuf headers prerequisites for modules and tests.

nrdmao:
Dependency failures can be seen on massively parallel builds on heavily loaded servers.


sangjinhan:
Merged, thanks.

tumblr_m14599p5nx1r8y5dso1_1280

PR #779, opened 2018-03-17
core/Makefile: fix DPDK check indentation

melvinw:
Make on my machine is really picky and barfs in strange ways because of
the current indentation. Targets like `make protobuf` will fail, but
`make`/`make all` don't.


melvinw:
I admit that the second reason for this commit might be a bit abusive on my part. I have a strange use case for which this is the quickest/easiest fix.

The proper solution here might be to break the `protoc` targets into a separate makefile in `core/pb`.

melvinw:
Nevermind the comment above. Fond a cleaner solution for my use case. This just fixes the indentation now.

PR #782, opened 2018-03-30
Make build.py Python 3 compatible

sangjinhan:
In Python 3, by default subprocess.communicate()/check_output() returns a bytes obejct, not str. Also `dict.keys()` must not be used for a loop in Python 3 if `dict` is modified.


chris3torek:
image

PR #785, opened 2018-04-03
pybess: replace basestring with str

sangjinhan:
In python3, every string is unicode and basestring does not exist any longer.


sangjinhan:
@chris3torek addressed the issue. thanks!

PR #793, opened 2018-04-26
bug fix in EmitPacket()

inhocho89:
In Module::EmitPacket() method, it is possible to add a packet to the packetBatch (`batch->add(pkt)`) which is already full. `batch` should be newly created one.


shinae-woo:
image

PR #797, opened 2018-05-10
grpc: do not use global callbacks to serialize

sangjinhan:
BESS gRPC server handlers are not thread-safe, so we use global
callbacks to acquire/release a mutex before/after method invocation.
For whatever reason, modules may want to run their own gRPC server.
This can cause very subtle issues. For example, the module's gRPC server
may terminate while holding the mutex, thus blocking all BESS gRPC
server indefinitely.

The gRPC global callback is an undocumented feature, so it's good
to avoid using it anyway. Unfortunately there is no per-server callback
mechanism provided by gRPC. Instead this patch adds a lock for every
method.

This fix was made and tested by @shinae-woo


melvinw:
cat

PR #799, opened 2018-05-16
ignore *.o.ur-safe files

chris3torek:
Some recent Ubuntu update generates these files when linking
kernel modules.  Since they are build artifacts, ignore them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ env/.vagrant
 *.symvers
 *.mod.c
 modules.order
+*.o.ur-safe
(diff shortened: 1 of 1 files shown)


ddiproietto:
gatto

justinemarie:
What's with the cat with the eye masks???

ddiproietto:
Alright, I'll reveal my sources:

https://www.nanopress.it/curiosita/foto/gatti-con-occhi-da-cartone-animato_4787_41.html

PR #800, opened 2018-05-16
Stop pausing workers for every gatehook command

### LINKED ISSUE #798: gatehook commands pause workers even if thread-safe
@ddiproietto and @pxgao discovered that thread-safe gatehooks still pause workers.

This appears to be the right fix:
```
diff --git a/core/bessctl.cc b/core/bessctl.cc
index 75616b4..30270e0 100644
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -1408,8 +1408,6 @@ class BESSControlImpl final : public BESSControl::Service {
                                gate_idx, rh.hook_name().c_str());
     }
 
-    WorkerPauser wp;
-
     // DPDK functions may be called, so be prepared
     current_worker.SetNonWorker();
```
but it may take me a bit to make sure (and it would be nice to auto-pause if NOT thread-safe, which this won't do).

chris3torek:
Thread-safe gatehook commands paused workers.  Now we just
don't pause at all; non-thread-safe gate hook commands
will produce an error like non-thread-safe module commands.

Closes #798.

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -1420,8 +1420,6 @@ class BESSControlImpl final : public BESSControl::Service {
                                gate_idx, rh.hook_name().c_str());
     }
 
-    WorkerPauser wp;
-
     // DPDK functions may be called, so be prepared
     current_worker.SetNonWorker();
 


ddiproietto:
gatto

PR #804, opened 2018-06-14
port: Return const reference from conf() accessor

ddiproietto:
According to Google C++ Style Guide we should return a const reference here
instead of making a copy.

I fixed this because I noticed that the accessor is used very often in the
fastpath ({Port,Queue}{Inc,Out}::RunTask call that once per batch). I haven't
noticed any performance impact, I suspect the compiler is smart enough to avoid
the copy, but I think it's better to change it.

### DIFF
--- a/core/port.h
+++ b/core/port.h
@@ -278,7 +278,7 @@ class Port {
                      const queue_t *queues, int num);
 
   const std::string &name() const { return name_; }
-  Conf conf() const { return conf_; }
+  const Conf &conf() const { return conf_; }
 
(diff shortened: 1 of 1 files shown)


sangjinhan:
I am not sure if the style guide mentions which to favor between const ref vs. copy for a return value (please correct me if I am wrong). I do agree const ref makes more sense here.

sangjinhan:
Thanks!
9483d1246510de872ffc4023c9253fc7

PR #805, opened 2018-06-20
pybess: add plugib_pb to path

melvinw:


melvinw:
In python3 not having `plugin_pb` on path seems to break `bessctl`. While this change fixes my use case, it does contradict the original comment about a similar import of `builtin_pb`. @chris3torek have any thoughts?

PR #806, opened 2018-07-04
clang-format: Remove Objective-C related options

sangjinhan:
... otherwise it causes "key errors" on some clang-format versions.


shinae-woo:
Merged, Thanks.
image

PR #807, opened 2018-07-07
Switch to C++17 / g++7

sangjinhan:
Currently BESS is based on C++14, and this PR makes C++17 the minimum compiler requirement. This is to take advantage of "modern" features provided by C++17. One example included in this PR is that it removes the custom memory allocator(`mem_alloc.cc`) for alignment, as C++17 now supports a new feature called over-alignment new. g++5/6 does not fully support C++17 features, so this PR makes BESS require g++-7 at least.

We also drop supporting i386. As recent releases of Ubuntu do not provide i386 version, it has been burdensome to maintain both 32-bit ans 64-bit compatibility. The existing master branch (g++5 and 32bit support) will be archived to c++14-legacy, with only bug patches to be added.

The Travis CI is going to fail since this PR requires some changes in the build container. They will be another PR.


shinae-woo:
Merged. Thanks. 

image

PR #808, opened 2018-07-07
Rework scripts in the env directory

sangjinhan:
* Use APT to install gRPC on Ubuntu 18.04, instead of building from source.
* Reflect the changes in minimum software requirements in the docker container and vagrant VM.
* Break the Ansible script into smaller ones, each with a different purpose (runtime, build dependency, dev, etc.)


shinae-woo:
Thanks, Merged. Let's drink!

image

PR #809, opened 2018-07-08
vport_zc: remove ZeroCopyVPort

sangjinhan:
ZeroCopyVPort has been provided but never really well tested,
documented, or maintained. Given the security concern caused by zero
copy, perhaps it is time for ZeroCopyVPort to retire.

If you want high-performance packet I/O for kernel-bypassing applications,
you can use PMDPort (with vhost-user in BESS, and virtio-user in client
ports) as an alternative.


shinae-woo:
Please rebase so that Travis config updated 👍

shinae-woo:
Merged, thanks!

image

PR #811, opened 2018-07-13
ansible: fix broken link and missing package

mukerjee:


sangjinhan:
Oh, it's my bad. `bess.yml` was meant to be removed, but I forgot to do it in my PR. The file should be gone.

The kernel header required to build the vport module (in `core/kmod`) is now installed by `kmod.yml` (`linux-generic` includes `linux-headers-generic`) `build-dep.yml` is responsible for building the bess daemon itself, but nothing more.

sangjinhan:
It might be useful to add `linux-headers-{{ansible_kernel}}` to `kmod.yml` as you did, in case the running kernel version is different from the installed one (a newer kernel package installed but not rebooted yet).

mukerjee:
Okay-- I can make that fix. If I remove bess.yml, what other ansible playbook should I call to build bess?

sangjinhan:
I deliberately removed the code that clones the repo and runs `./build.py` since I felt it somewhat unsolicited/intrusive. Perhaps showing some message like "now you are ready to build by running `build.py` script" at the end of `build-dep.yml` should be enough...?

sangjinhan:
Thanks!
tumblr_mch3w3j2pb1rgnmlzo1_500

PR #812, opened 2018-07-13
utils/cuckoo_map: add Emplace, move-based Insert

melvinw:


core/utils/cuckoo_map.h
```diff
@@ -206,6 +206,73 @@ class CuckooMap {
     return entry;
   }
 
+  // Insert/update a key value pair
+  // Return the pointer to the inserted entry
+  Entry *Insert(const K &key, V &&value, const H &hasher = H(),
+                const E &eq = E()) {
+    Entry *entry;
```
shinae-woo: Just for clarifying,
In comments, you said that 'move Insert()', but here you create another Insert with Rvalue-Reference. Which one do you intended?

melvinw: Ah, my intention was the latter. I'll update the commit message to be clearer

shinae-woo: Got it

core/utils/cuckoo_map_test.cc
```diff
@@ -45,6 +45,35 @@ TEST(CuckooMapTest, Insert) {
...
+  int a;
+  int b;
+  int c;
+};
+
+// Test Move-Insert function
+TEST(CuckooMapTest, MoveInsert) {
+  CuckooMap<uint32_t, Foo> cuckoo;
+  Foo expected = Foo(1, 2, 3);
+  const Foo &x = cuckoo.Insert(1, expected)->second;
```
sangjinhan: I think it ends up calling the copy one, when used without `std::move(expected)`

core/utils/cuckoo_map_test.cc
```diff
@@ -45,6 +45,35 @@ TEST(CuckooMapTest, Insert) {
   EXPECT_EQ(cuckoo.Insert(1, 1)->second, 1);
 }
 
+struct Foo {
+  Foo() : a(), b(), c() {}
+
```
sangjinhan: Could you delete the default copy constructor and copy assignment operator? For the purpose of the test, it would be better for Foo to avoid accidental copy.

core/utils/cuckoo_map_test.cc
shinae-woo: That's what I try to understand and point out! :)

core/utils/cuckoo_map_test.cc
melvinw: Ah, sure-- this will require a few non-trivial changes to CuckooMap, though. It currently assumes value types are copyable and assignable (and now moveable)

sangjinhan:
Added notready label waiting for @melvinw's new copy/move insert & emplace implementation without code redundancy

melvinw:
Ok, finally found some time to polish this up. It now supports copy/move insertion and emplace w/ no code duplication

core/utils/cuckoo_map.h
```diff
@@ -173,26 +174,24 @@ class CuckooMap {
...
+  template <typename... Args>
+  Entry* DoEmplace(const K& key, const H& hasher, const E& eq, Args&&... args) {
     Entry* entry;
     HashResult primary = Hash(key, hasher);
 
     EntryIndex idx = FindWithHash(primary, key, eq);
     if (idx != kInvalidEntryIdx) {
       entry = &entries_[idx];
-      entry->second = value;
+      new (&entry->second) V(std::forward<Args>(args)...);
```
chris3torek: General note-to-the-universe (not an issue with this change): I hate placement-new. 😝

core/utils/cuckoo_map.h
```diff
@@ -501,12 +518,14 @@ class CuckooMap {
...
   // Resize the space of buckets, and rehash existing entries
+  template <typename VV>
   void ExpandBuckets(const H& hasher, const E& eq) {
     CuckooMap<K, V, H, E> bigger(buckets_.size() * 2, entries_.size());
 
-    for (const auto& e : *this) {
-      // While very unlikely, this insert() may cause recursive expansion
-      bool ret = bigger.Insert(e.first, e.second, hasher, eq);
+    for (auto& e : *this) {
+      // While very unlikely, this Insert() may cause recursive expansion
```
chris3torek: trivial: this isn't `Insert()` any more, it's `DoEmplace` now.

Overall this looks fine, but the whole "VV might be `V&&`, or might be `const V&`" thing (where `V` is already also a template) gives me a headache.

core/utils/cuckoo_map.h
```diff
@@ -201,11 +200,27 @@ class CuckooMap {
...
+      ExpandBuckets<std::conditional_t<std::is_move_constructible<V>::value,
+                                       V&&, const V&>>(hasher, eq);
     }
     return entry;
   }
 
+  // Insert/update a key value pair
+  // Return the pointer to the inserted entry
+  template <typename VV>
+  Entry* Insert(const K& key, VV&& value, const H& hasher = H(),
```
chris3torek: `template <typename VV> ... VV&& ... std::forward<VV>`: deep C++14 voodoo: https://en.cppreference.com/w/cpp/utility/forward

Discussed offline: maybe should comment that constructor might not be called if the `Insert` fails (returns `nullptr`). We don't depend on this, but if someone does, it could be surprising.

melvinw:
Thanks for the review @chris3torek! Made the adjustments we talked about offline

core/utils/cuckoo_map.h
```diff
@@ -222,7 +222,7 @@ class CuckooMap {
   // called.
   template <typename... Args>
   Entry* Emplace(const K& key, Args&&... args) {
-    return DoEmplace(key, H(), E(), args...);
+    return DoEmplace(key, H(), E(), std::forward<Args>(args)...);
```
chris3torek: Ergh, I missed this one. 😿  Sorry about that!

melvinw:
Thanks for the fixes! If you want to merge this now, go ahead. Otherwise I'd like to fix our dependence on default constructors and add move support for keys, too.

sangjinhan:
Merged, thanks! And sorry about the default constructor issue; I made a bad choice to use a vector as a backup store in the first place.

jjkybd3ns9i6yyi07gqpgybc_500

PR #814, opened 2018-07-16
travis: Use prebuilt DPDK when building BESS

sangjinhan:
Now DPDK is built with -march=corei7 (Nehalem), not -march=native. Since the minimum CPU requirement for BESS is also Nehalem, we can safely assume that the compiled DPDK can be used on any (including Travis VMs) machines if only they can run BESS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,13 +34,14 @@ before_install:
 
 install:
   # note that if you are building a slightly different dpdk you will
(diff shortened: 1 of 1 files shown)


sangjinhan:
Removing "not ready" label, now DPDK is built for Nehalem indeed (#829).

shinae-woo:
image

PR #815, opened 2018-07-16
Fix CLI bugs regarding command auto-completion

shinae-woo:
- flushing `fout` after matching partial prompted input to possible commands so that bessctl shell is not broken
(somehow, it was fine in Python2, but not anymore in Python3)
- Removing command ambiguity when the exact match and partial matches are coexists
In case of following a set of commands
```
\tcommand
\tcommand_extends
```
Before,
```
$ command
--> returns ambiguity
```
Now,
```
$ command
--> running the command
```

### DIFF
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -341,27 +341,28 @@ class CLI(object):
 
     def find_cmd(self, line):
         matched, matched_low = self.list_matched(line, ['full'])
+        line_stripped = line.strip()
 
         if len(matched) == 1:
             return matched[0]
 
         elif len(matched) >= 2:
-            for m in  matched: # return if exact match exists
-                if line.strip() == m[0]:
+            for m in matched: # return if exact match exists
+                if line_stripped == m[0]:
                     return m
 
-            self.err('Ambiguous command "%s". Candidates:' % line.strip())
+            self.err('Ambiguous command "%s". Candidates:' % line_stripped)
             for cmd, desc, _ in matched + matched_low:
                 self.ferr.write('  %-50s%s\n' % (cmd, desc))
 
         elif len(matched) == 0:
             matched, matched_low = self.list_matched(line, ['partial'])
             if len(matched) > 0:
-                self.err('Incomplete command "%s". Candidates:' % line.strip())
+                self.err('Incomplete command "%s". Candidates:' % line_stripped)
                 for cmd, desc, _ in matched + matched_low:
                     self.ferr.write('  %-50s%s\n' % (cmd, desc))
             else:
-                self.err('Unknown command "%s".' % line.strip())
+                self.err('Unknown command "%s".' % line_stripped)
 
         raise self.InvalidCommandError()
 
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -307,6 +307,7 @@ class CLI(object):
             self.fout.write('\n')
             self.fout.write(''.join(buf))
             self.fout.write('%s%s' % (self.get_prompt(), line))
+            self.fout.flush()
 
         return []
 
--- a/bessctl/cli.py
+++ b/bessctl/cli.py
@@ -345,6 +345,10 @@ class CLI(object):
             return matched[0]
 
         elif len(matched) >= 2:
+            for m in  matched: # return if exact match exists
+                if line.strip() == m[0]:
+                    return m
+
             self.err('Ambiguous command "%s". Candidates:' % line.strip())
             for cmd, desc, _ in matched + matched_low:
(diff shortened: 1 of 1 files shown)


bessctl/cli.py
```diff
@@ -345,6 +345,10 @@ def find_cmd(self, line):
             return matched[0]
 
         elif len(matched) >= 2:
+            for m in  matched: # return if exact match exists
```
chris3torek: Trivial, nitty: there's a double space here (`for m in<space><space>matched`).

Q (also pretty trivial): should we capture `line.strip()` once instead of calling it once per match?

More seriously: it seems like a bad design in general to have complete commands that are prefixes of longer commands. But maybe we have to allow it in general anyway.

shinae-woo: @chris3torek It was tab 😈...

About the line.strip(), the space at the end here plays a important role, deciding whether it is the end of a token or partial of a token. The design-wise not super clean. ;ㅁ;

shinae-woo:
I need additional review, for my last commit, please.
It's for `sh gatehook` makes running without ambiguity.

bessctl/cli.py
```diff
@@ -231,11 +238,16 @@ def list_matched(self, line, filters):
 
         ret = []
         ret_low = []
-        for cmd, score in matched_list:
-            if score == max_score:
-                ret.append(cmd)
-            else:
-                ret_low.append(cmd)
+
```
chris3torek: I like the new scheme.  The two assignments right above here are redundant now...

bessctl/cli.py
```diff
@@ -247,7 +259,7 @@ def _do_complete(self, line, partial_word):
         for cmd in self.cmdlist:
             syntax = cmd[0]
             match_type, sub_candidates, \
-                syntax_token, score = self.match(syntax, line)
+                syntax_token, _ = self.match(syntax, line)
```
chris3torek: Doesn't `self.match` return a 5-tuple now?  If so there should be two `_`s (or use the new Python3 feature, `*_`).

.gitignore
```diff
@@ -7,6 +7,8 @@ core/.deps
 core/extra.mk
 core/extra.dpdk.mk
 *.retry
+.mypy_cache
+*.cache.mk
```
sangjinhan: This one was included in 57f5a2

bessctl/cli.py
```diff
@@ -207,22 +214,22 @@ def match(self, syntax, line):
...
-                return 'full', ['\n'], '', score
-            return 'full', candidates, syntax_token, score
-        return 'nonmatch', [], '', score
+                return 'full', ['\n'], '', score, exact_score
+            return 'full', candidates, syntax_token, score, exact_score
+        return 'nonmatch', [], '', score, exact_score
 
-    # filters is a list of 'full', 'partial', 'nonmatch'
-    def list_matched(self, line, filters):
+    # filter is one list of 'full', 'partial', 'nonmatch'
```
sangjinhan: one of

bessctl/cli.py
```diff
@@ -231,11 +238,16 @@ def list_matched(self, line, filters):
...
-            else:
-                ret_low.append(cmd)
+
+        ret = [m[0] for m in matched_list if m[1] == max_score]
+        ret_low = [m[0] for m in matched_list if m[1] != max_score]
+
+        # Find exact matches without ambiguity
+        if filter == 'full' and len(ret) > 1:
+            full_matches = [m for m in matched_list if m[1] == max_score]
+            full_matches.sort(key=itemgetter(2), reverse=True) # sorted by exact score
```
sangjinhan: Maybe it's just me, but `key=itemgetter(2)` (which requires external module) sounds less familiar to me than `key=lambda m: m[2]`

chris3torek: Using itemgetter (and attrgettr) in sort is kind of a standard idiom (see https://docs.python.org/2/howto/sorting.html). I'm not sure if the sort code optimizes these specially. If it does that might be a reason to use itemgetter, if not they're kind of six-of-one half-a-dozen-of-the-other.

shinae-woo:
@chris3torek @sangjinhan Fixed! What I am confused a bit is that is `bessctl` is expected to support both python2/phthon3? Or on the way of moving somewhere?

sangjinhan:
It's supposed to support the both versions (not that we are doing it very well at the moment). Travis runs both Python2 and 3 for the same scripts.

chris3torek:
Ah, I wasn't sure about py2k vs py3k at this point—I defer to @sangjinhan here...

PR #818, opened 2018-07-21
pybess: handle temporary HTTP2 disconnection

sangjinhan:
... during gRPC server launch. This race often happen `daemon start`; when bessd starts, if gRPC client connection is made prematurely, it is disconnected causing client to block forever.

### DIFF
--- a/pybess/bess.py
+++ b/pybess/bess.py
@@ -105,6 +105,7 @@ def _import_modules(name, subdir):
         raise SystemExit(1)
     return mod
 
+
 module_pb = _import_modules('module_pb', None)
 port_msg = _import_modules('port_msg', 'ports')
 sys.path = old_path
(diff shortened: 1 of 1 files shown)


pybess/bess.py
```diff
@@ -192,20 +199,24 @@ def connect(self, grpc_url=DEF_GRPC_URL):
...
-                raise self.APIError(
-                    'Connection to {} failed'.format(grpc_url))
-            time.sleep(0.1)
+            self.status = None
+            self.peer = grpc_url
+            self.channel = grpc.insecure_channel(grpc_url)
+            self.channel.subscribe(self._update_status, try_to_connect=True)
+            self.stub = service_pb2.BESSControlStub(self.channel)
+
+            while not self.is_connected():
```
chris3torek: The logic with these two nested `while not self.is_connected()` seems a bit tortured. Would it be possible to restructure this as:

    while not self.is_connected():
        if <some condition>:
            do the status, peer, channel, stub setup
        if <other conditions>: # or `elif`?
            disconnect, maybe raise error
        time.sleep(0.1)

?

shinae-woo: Maybe changing the first `while not` can be replaced with `if` also good.

sangjinhan: Thanks, revised. The `while` cannot be replaced with `if` in case of involuntary disconnection.

shinae-woo:
image

PR #821, opened 2018-07-25
Support multiple gatehooks of the same type installed per gate

shinae-woo:
This PR is for support of installing multiple gatehooks of the same class type for a single gate. 
1) Now gatehooks also have it's own name with class name
2) If gatehook name is not given with creation request, the default name using name template is assigned
3) Multiple gatehooks for the same class type can be installed now
4) Necessary changes for gatehook request/response are made accordingly

### DIFF
--- a/core/gate.cc
+++ b/core/gate.cc
@@ -42,60 +42,60 @@ namespace bess {
 
 const GateHookCommands GateHook::cmds;
 
-bool GateHookFactory::RegisterGateHook(GateHook::constructor_t constructor,
+bool GateHookBuilder::RegisterGateHook(GateHook::constructor_t constructor,
                                        const std::string &class_name,
                                        const std::string &name_template,
                                        const std::string &help_text,
                                        const GateHookCommands &cmds,
                                        GateHook::init_func_t init_func) {
-  return all_gate_hook_factories_holder()
+  return all_gate_hook_builders_holder()
       .emplace(std::piecewise_construct, std::forward_as_tuple(class_name),
                std::forward_as_tuple(constructor, class_name, name_template,
                  help_text, cmds, init_func))
       .second;
 }
 
-std::map<std::string, GateHookFactory>
-    &GateHookFactory::all_gate_hook_factories_holder(bool reset) {
-  // Maps from hook names to hook factories. Tracks all hooks (via their
-  // GateHookFactorys).
-  static std::map<std::string, GateHookFactory> all_gate_hook_factories;
+std::map<std::string, GateHookBuilder>
+    &GateHookBuilder::all_gate_hook_builders_holder(bool reset) {
+  // Maps from hook names to hook builders. Tracks all hooks (via their
+  // GateHookBuilders).
+  static std::map<std::string, GateHookBuilder> all_gate_hook_builders;
 
   if (reset) {
-    all_gate_hook_factories.clear();
+    all_gate_hook_builders.clear();
   }
 
-  return all_gate_hook_factories;
+  return all_gate_hook_builders;
 }
 
-const std::map<std::string, GateHookFactory>
-    &GateHookFactory::all_gate_hook_factories() {
-  return all_gate_hook_factories_holder();
+const std::map<std::string, GateHookBuilder>
+    &GateHookBuilder::all_gate_hook_builders() {
+  return all_gate_hook_builders_holder();
 }
 
-// Creates new gate hook from the given factory, initializes it,
+// Creates new gate hook from the given builder, initializes it,
 // and adds it to the given gate.  If any of these steps go wrong
 // the gatehook instance is destroyed and we return a failed
 // CommandResponse, otherwise we return a successful one.
-CommandResponse Gate::NewGateHook(const GateHookFactory *factory, Gate *gate,
+CommandResponse Gate::CreateGateHook(const GateHookBuilder *builder, Gate *gate,
                                   bool is_igate, const std::string &name,
                                   const google::protobuf::Any &arg) {
-  bess::GateHook *hook = factory->hook_constructor_();
-  CommandResponse init_ret = factory->hook_init_func_(hook, gate, arg);
+  bess::GateHook *hook = builder->hook_constructor_();
+  CommandResponse init_ret = builder->hook_init_func_(hook, gate, arg);
   if (init_ret.error().code() != 0) {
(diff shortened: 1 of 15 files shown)


core/gate.cc
```diff
@@ -42,67 +43,92 @@ namespace bess {
...
 
-// Creates new gate hook from the given factory, initializes it,
+const std::string Gate::GenerateDefaultName(const GateHookBuilder *builder,
+    Gate *gate) {
+  const std::string &name_template = builder->name_template();
+
+  for (int i = 0;; i++) {
+    const std::vector<GateHook *> hooks = gate->hooks();
+    std::string name = name_template + std::to_string(i);
+    auto it = std::find_if(hooks.begin(), hooks.end(),
```
sangjinhan: Just a matter of style, but you may want to simplify here like this (not tested). I personally find using lambdas more intuitive than using functors...
```
if (std::none_of(hooks.begin(), hooks.end(), [&](const GateHook *hook){ return name == hook->name(); }) {
  return name;
}
```

There are some `find_if`-then-check or manual `for` loops on many occasions of the code, which can be also updated.

core/gate.cc
```diff
@@ -42,67 +43,92 @@ namespace bess {
...
+  return all_gate_hook_builders_holder();
 }
 
-// Creates new gate hook from the given factory, initializes it,
+const std::string Gate::GenerateDefaultName(const GateHookBuilder *builder,
+    Gate *gate) {
+  const std::string &name_template = builder->name_template();
+
+  for (int i = 0;; i++) {
+    const std::vector<GateHook *> hooks = gate->hooks();
```
sangjinhan: This may be taken out of the loop...? Also, `hooks` could be a reference to avoid copying the vector.

protobuf/bess_msg.proto
```diff
@@ -552,20 +556,26 @@ message PcapngArg {
 
 
 message GateHookInfo {
-  string hook_name = 1;         /// Name of the hook
-  string module_name = 2;       /// Name of module
+  string class_name = 1;         /// Name of the hook
```
sangjinhan: Misalignment (sorry)

core/gate.cc
```diff
@@ -42,67 +43,92 @@ namespace bess {
...
-                          gate->module()->name().c_str(), is_igate ? 'i' : 'o',
-                          gate->gate_idx());
+    return nullptr;
   }
-  return CommandSuccess();
+  return hook;
 }
 
-int Gate::AddHook(GateHook *hook) {
+int Gate::AddHook(GateHook *hook, pb_error_t *error) {
```
sangjinhan: I think it is fine to keep setting `error` by the caller, to minimize the footprint of Protobuf-related code.

core/gate.cc
```diff
@@ -136,8 +162,29 @@ void Gate::RemoveHook(const std::string &name) {
...
+  for (const auto &hook : hooks_) {
+    if (hook->class_name() == class_name) {
+      return hook;
+    }
+  }
+  return nullptr;
+}
+
+void Gate::RemoveHookByClass(const std::string &class_name) {
+  for (auto it = hooks_.begin(); it != hooks_.end(); ++it) {
```
sangjinhan: `std::remove_if`? (although it changes the semantics from "remove first one" to "remove all of them")

sangjinhan: There seems no code that calls this method, so I thought the semantic change wouldn't matter. Anyway, it would be good to comment the expected behavior in the header file.

core/gate.h
```diff
@@ -186,30 +210,37 @@ class Gate {
   const std::vector<GateHook *> &hooks() const { return hooks_; }
 
   // Creates, initializes, and then inserts gate hook in priority order.
-  CommandResponse NewGateHook(const GateHookFactory *factory, Gate *gate,
-                              bool is_igate, const google::protobuf::Any &arg);
+  GateHook *CreateGateHook(const GateHookBuilder *builder, Gate *gate,
+                              const std::string &name,
```
sangjinhan: Misalignment here.

protobuf/bess_msg.proto
shinae-woo: Nope, I'm sorry. 😈

core/gate.cc
shinae-woo: Good one!!!

core/gate.h
shinae-woo: 😈

core/gate.cc
shinae-woo: I think removing all the hooks with the specified class would be more appropriate. Thanks!

protobuf/bess_msg.proto
```diff
@@ -400,18 +400,22 @@ message GetModuleInfoRequest {
...
-    uint64 bytes = 5;            /// # of bytes seen
-    double timestamp = 6;       /// The time that cnt/pkts counters were read
-    repeated string hook_name = 7;         /// List of hooks per gate
+    uint64 igate = 1;                /// Input gate ID
+    repeated OGate ogates = 2;       /// The list of upstream output gates
+    uint64 cnt = 3;                  /// # of packet batches seen
+    uint64 pkts = 4;                 /// # of packets seen
+    uint64 bytes = 5;                /// # of bytes seen
+    double timestamp = 6;            /// The time that cnt/pkts counters were read
+    repeated GateHook gatehooks = 7;  /// List of gate hook
```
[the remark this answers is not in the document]
shinae-woo: Sorry. I didn't get why we should renumber or reserve. 😿 Could you explain in more detail, please?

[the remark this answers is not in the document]
chris3torek: We talked about this offline but I'll just record the summary here for posterity 😸 ... in google protobufs, there's no automatic backwards compatibility, but you can reserve the numeric value that was used in an older protocol message, and hence detect client/server protocol mismatches. If you re-use the number, an old client sending to a new server (or vice versa) can misinterpret the message.

(In this case it's not very important since the interior parts of the item won't line up—the old one was a String and the new one is a sub-message—so the client and server will notice that they're not communicating successfully anyway.  Reserving would just make it clear that there was a change at some point.  We have not been good about this in the past...)

sangjinhan:
Merged, thanks!
jjkybd3nsejr02wbn9sfjel3o1_500

PR #822, opened 2018-07-25
Thread-safe track gatehook having per-worker statistics

shinae-woo:
Track data-path can be accessed from multiple threads. With this PR, `track` has a per-thread storage to update and aggregate them when requesting the actual number.

### DIFF
--- a/core/gate_hooks/track.cc
+++ b/core/gate_hooks/track.cc
@@ -68,9 +68,11 @@ void Track::ProcessBatch(const bess::PacketBatch *batch) {
     return;
   }
 
+  uint64_t bytes = 0;
   for (size_t i = 0; i < cnt; i++) {
-    stat->bytes += batch->pkts()[i]->data_len() + kEthernetOverhead;
+    bytes += batch->pkts()[i]->data_len() + kEthernetOverhead;
   }
+  stat->bytes += bytes;
 }
 
 ADD_GATE_HOOK(Track, "count the packets and batches")
--- a/core/gate_hooks/track.cc
+++ b/core/gate_hooks/track.cc
@@ -44,9 +44,7 @@ const GateHookCommands Track::cmds = {{"reset", "EmptyArg",
 Track::Track()
     : bess::GateHook(Track::kName, Track::kPriority),
       track_bytes_(),
-      cnt_(),
-      pkts_(),
-      bytes_() {}
+      worker_stats_() {}
 
 CommandResponse Track::Init(const bess::Gate *, const bess::pb::TrackArg &arg) {
   track_bytes_ = arg.bits();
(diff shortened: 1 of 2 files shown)


core/gate_hooks/track.cc
```diff
@@ -44,33 +44,32 @@ const GateHookCommands Track::cmds = {{"reset", "EmptyArg",
 Track::Track()
     : bess::GateHook(Track::kName, Track::kPriority),
       track_bytes_(),
-      cnt_(),
-      pkts_(),
-      bytes_() {}
+      worker_stats_() {}
```
ddiproietto: My C++ knowledge is not enough to understand why this is initialized to 0, but I tested it and it works well

sangjinhan: This is something like `int a{};`, which will be initialized to zero. If a type doesn't have anything but default constructor, `a()` or `a{}` will trigger value initialization (i.e., zeroed), while simply `a` incurs default initialization (i.e., doing nothing for `int` or `TrackStats`).

core/gate_hooks/track.cc
```diff
@@ -44,33 +44,32 @@ const GateHookCommands Track::cmds = {{"reset", "EmptyArg",
...
-  pkts_ += cnt;
+  stat->cnt += 1;
+  stat->pkts += cnt;
 
   if (!track_bytes_) {
     return;
   }
 
   for (size_t i = 0; i < cnt; i++) {
-    bytes_ += batch->pkts()[i]->data_len() + kEthernetOverhead;
```
sangjinhan: My hunch is that here `stat->bytes` may be loaded and stored every iteration in the loop, instead of doing accumulation with a register. Could you use a local variable and make a quick performance comparison, with `perftest/chain.bess` and byte counting enabled?

shinae-woo:
```
for (size_t i=0; i < cnt; i++)
   stat->bytes += batch->pkts()[i]->data_len() + kEthernetOverhead;
}
```
Chain/ 1 pps:     26390664	counts/s:       824708
Chain/ 2 pps:     13860911	counts/s:       433153
Chain/ 4 pps:      7153635	counts/s:       223551
Chain/ 8 pps:      3719552	counts/s:       116236
Chain/16 pps:      1879005	counts/s:        58719
Chain/32 pps:       941429	counts/s:        29420

```
uint64_t bytes = 0;
for (size_t i = 0; i < cnt; i++) {
  bytes += batch->pkts()[i]->data_len() + kEthernetOverhead;
}
stat->bytes += bytes;
```

Chain/ 1 pps:     26657973	counts/s:       833062
Chain/ 2 pps:     14204597	counts/s:       443894
Chain/ 4 pps:      7272441	counts/s:       227264
Chain/ 8 pps:      3719219	counts/s:       116226
Chain/16 pps:      1878811	counts/s:        58713
Chain/32 pps:       942076	counts/s:        29440

Guess Compilers do the rock here!

shinae-woo:
@chris3torek 🐱  May I get a picture of your cat? 🐈

chris3torek:
64 worker kittens (well, I didn't count - also, not _my_ kittens...)

image

Issue #825, opened 2018-08-02
Command `track disable` is no longer working


levaitamas:
Hi!

It seems the `track disable` command has no effect after commit bc55bb39.

Consider the following snippet:

```sh
./bessctl daemon start -- show version \
-- add module Source source0 -- add module Sink sink0 \
-- add connection source0 sink0 \
-- show pipeline \
-- show gatehook \
-- track disable source0 out 0\
-- show pipeline \
-- show gatehook; \
./bessctl daemon stop
```

##### Output of bc55bb39:

```
v0.3.0-625-gbc55bb39
+---------+                  +-------+
| source0 |  :0 1000192 0:   | sink0 |
| Source  | ---------------> | Sink  |
+---------+                  +-------+
Track::track0    source0:0
+---------+                   +-------+
| source0 |  :0 19105344 0:   | sink0 |
| Source  | ----------------> | Sink  |
+---------+                   +-------+
Track::track0    source0:0
```
Tracking still works, however, it should be disabled.

##### Output of 9ebc5651 (parent of bc55bb39):

```
v0.3.0-615-g9ebc5651
+---------+                  +-------+
| source0 |  :0 1516672 0:   | sink0 |
| Source  | ---------------> | Sink  |
+---------+                  +-------+
Track            source0:0
+---------+            +-------+
| source0 |  :0 ? 0:   | sink0 |
| Source  | ---------> | Sink  |
+---------+            +-------+
*** Error: There is no active gatehook to show.
  Command failed: show gatehook
```
Track disable works as desired.

shinae-woo:
Hi Tamas! Thank you for reporting! I'm gonna make a PR to fix the problem you reported.

PR #826, opened 2018-08-03
Fix bugs for removing gatehook by classname

shinae-woo:
This PR allows removing gate hook by class name. Reported by @levaitamas #825


chris3torek:
image

PR #827, opened 2018-08-06
module_test: support non-Ethernet packet capture

sangjinhan:
In the module_test framework, `run_module` and `run_pipeline` implicitly assumes the output packets are always Ethernet frames. This doesn't always hold since some modules may produce IP datagrams (or any other L3/L4 protocols), not L2 Ethernet frame. This patch is to accommodate such modules.


shinae-woo:
Merged, Thanks!

(p.s. We keep losing coverage... TT)
image

chris3torek:
image

chris3torek:
Oops, didn't see that @shinae-woo had already merged it. I guess you get TWO kitties!

PR #833, opened 2018-08-14
Cleanup .gitignore file

sangjinhan:
* Ignore all dot files by default
* Add [compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html)
* Separate `core/.gitignore` from the main one


core/.gitignore
```diff
@@ -0,0 +1,22 @@
+extra.mk
+extra.dpdk.mk
+
+# Executables
+bessd
+
+# Unittest/benchmark executables
+test_all
```
ddiproietto: If I'm not mistaken, `test_all` doesn't exist anymore, we only have `all_test`, right?

core/.gitignore
```diff
@@ -0,0 +1,22 @@
...
+
+# Code coverage files
+*.gcda
+*.gcno
+*.gcov
+coverage.info
+coverage_html
+htmlcov
+
+# Auto-genereated by make
```
ddiproietto: s/genereated/generated/

core/.gitignore
```diff
@@ -0,0 +1,22 @@
+extra.mk
+extra.dpdk.mk
```
ddiproietto: Maybe we should have `extra*.mk`, to mimic what's in `Makefile`?

core/.gitignore
sangjinhan: Thanks! All updated.

.gitignore
```diff
@@ -1,5 +1,5 @@
 # Temporary/hidden files
-.*.swp
+.*
```
chris3torek: This is minor, because Git never ignores files that are actually committed (well, technically, those that are in the index), but...

If you mark `.*` like this it might be good to explicitly un-ignore the dot files we have, i.e.:

    !.codecov.yml
    !.gitattributes
    !.gitignore
    !.hooks/
    !.travis.yml

(In particular this means if someone runs `git rm -r .; git add .` someday, the wanted dot-files get re-added to the index.)

sangjinhan: I wouldn't worry about this, since git add .` will warn that files are not being ignored.

ddiproietto:
cat

PR #834, opened 2018-08-16
build.py: Do not delete bess symlink

ddiproietto:
Since commit 8293a9fa195c("build: do not create bessd symlink at compile
time"), the symlink is committed to the repo.


chris3torek:
Still waiting for Travis before merging, but here, have a kitty 🐈
image

PR #835, opened 2018-08-17
makefile: make "all" be the default target

sangjinhan:


chris3torek:
(need to take some better kitty pictures, here's an old one that's a little blown out)
image

Issue #836, opened 2018-08-17
FIFO test fails sporadically (or sometimes consistently) on Ubuntu 18.04


chris3torek:
Something has happened internally timing-wise, and the 20 ms magic timeout has become too short.

Even 60 ms can still fail, as demonstrated here (some output snipped, gtest arguments from earlier failure-reducer).  70 or 80 ms works, but to be safe, we should probably go higher.  (Fix incoming, just creating an issue for it so that the fix is more searchable.)
```
$ cd core
$ FIFO_TEST_TIMEOUT=60 ./all_test --gtest_shuffle \
  --gtest_random_seed=58276 --gtest_filter=EmTableTest.*:FifoTestFixture.*
Note: Google Test filter = EmTableTest.*:FifoTestFixture.*
Note: Randomizing tests' orders with a seed of 58276 .
[==========] Running 11 tests from 2 test cases.
[----------] Global test environment set-up.
[----------] 7 tests from EmTableTest
[ RUN      ] EmTableTest.FindMakeKeysPktBatch
...
child 2: failed getting message from parent
utils/fifo_test.cc:309: Failure
Value of: ok
  Actual: false
Expected: true
failed to send <close> to child 2
[  FAILED  ] FifoTestFixture.MultipleFancyFifos (615 ms)
[----------] 4 tests from FifoTestFixture (1043 ms total)

[----------] Global test environment tear-down
[==========] 11 tests from 2 test cases ran. (1317 ms total)
[  PASSED  ] 10 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] FifoTestFixture.MultipleFancyFifos
```

chris3torek:
closed via #837

PR #837, opened 2018-08-17
tests: raise FIFO test default delay

chris3torek:
On at least some Ubuntu 18.04 systems, the FIFO test sometimes
fails when a 20 ms thread-sleep does not allow the parent process
to run to the point of having data ready in the test-control socket.

Raising the default delay to 100 ms seems reliable (70 ms suffices
in my tests).

Fixes issue #836.


ddiproietto:
gatto

PR #840, opened 2018-08-18
makefile: allow plugins to inject .mk files

sangjinhan:
Plugins may require some special software prerequisites, target dependencies, or compiler flags to build their own modules. Since it would be impractical to provide generic mechanisms to meet such plugin-specific requirements without some heavy lifting, this PR instead introduces a hacky way; it enables plugins to implant their extra*.mk files into the main Makefile.

### DIFF
--- a/core/Makefile
+++ b/core/Makefile
@@ -180,6 +180,9 @@ endif
 
 -include extra*.mk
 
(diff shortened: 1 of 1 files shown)


shinae-woo:
image

PR #854, opened 2018-09-28
Add protobuf package to pip installation list

sangjinhan:
Somewhat surprisingly, `protobuf` is not a prerequisite of `grpcio`, so we have to list it explicitly.


shinae-woo:
image

PR #859, opened 2018-10-08
add csv output to `monitor port`, `monitor tc`

melvinw:
This should make benchmarking stuff a bit easier by removing the need to scrape the screen. I'm also smuggling in some formatting clean up/bug fixes

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -48,6 +48,7 @@ import traceback
 import tempfile
 import signal
 import collections
+import contextlib
 import sugar
 
 try:
@@ -67,6 +68,12 @@ CONF_EXT = 'bess'
 # errors in configuration file
 class ConfError(Exception):
     pass
+
+
+# useful for creating dummy file objects for use with `with`
+@contextlib.contextmanager
+def noop():
+    yield None
 
 
 @staticmethod
@@ -1806,6 +1813,7 @@ TcCounterRate = collections.namedtuple('TcCounterRate',
 
 def _monitor_tcs(cli, *tcs):
     GUTTER_WIDTH = 5
+    FIELDS = ('CPU MHz', 'scheduled', 'Mpps', 'Mbps', 'pkts/sched', 'cycles/p')
 
     def get_delta(old, new):
         sec_diff = new.timestamp - old.timestamp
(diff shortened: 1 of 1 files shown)


bessctl/commands.py
```diff
@@ -1817,37 +1818,30 @@ def get_delta(old, new):
 
     def print_header(timestamp, name_len):
         cli.fout.write('\n')
-        fmt = '%-{}s%12s%12s%12s%12s%12s%12s\n'.format(name_len)
-        cli.fout.write(fmt %
-                       (time.strftime('%X') + str(timestamp % 1)[1:8],
-                        'CPU MHz', 'scheduled', 'Mpps', 'Mbps',
-                        'pkts/sched', 'cycles/p'))
+        fmt = '{:<%d}{:>12}{:>12}{:>12}{:>12}{:>12}{:>12}\n' % (name_len,)
```
chris3torek: Mixing old formatting with new formatting. I know some disapprove, but it sure is a lot simpler than `'{{:<{:d}}}{{:>12}}...'.format(name_len)`... 😺

bessctl/commands.py
```diff
@@ -1881,13 +1875,21 @@ def print_delta(tc, delta, name_len):
...
+                print_delta(now[tc].timestamp, 'W{} {}'.format(wids[tc], tc),
+                            get_delta(last[tc], now[tc]), max_len, csv)
 
             print_footer(max_len)
 
             for tc in tcs:
                 last[tc] = now[tc]
+    try:
+        csv_path = os.getenv('CSV', None)
+        csv_f = open(csv_path, 'w+') if csv_path is not None else None
```
chris3torek: Why open for `w+`?

In any case this is better as a `with` to make sure the file is closed even on interrupt, but `with` is slightly hard to use here since `csv_f` might be None.  You can use a no-op context manager: https://stackoverflow.com/questions/28680442/python-context-for-file-or-none

bessctl/commands.py
```diff
@@ -1774,18 +1774,28 @@ def get_total(arr):
...
                     get_total(list(last.values())),
-                    get_total(list(now.values()))))
+                    get_total(list(now.values()))), csv_f)
 
             for port in ports:
                 last[port] = now[port]
+
+    try:
+        csv_path = os.getenv('CSV', None)
+        csv_f = open(csv_path, 'w+') if csv_path is not None else None
```
chris3torek: (same comment as the other csv file)

bessctl/commands.py
melvinw: Yeah, it feels a bit dirty. I agree it's much easier to read than the alternative, though :)

bessctl/commands.py
melvinw: Nice! Implemented this. Much cleaner/safer now-- thanks!

bessctl/commands.py
```diff
@@ -1713,34 +1713,34 @@ def get_delta(old, new):
 
     def print_header(timestamp):
         cli.fout.write('\n')
-        cli.fout.write('%-20s%14s%10s%10s        %14s%10s%10s\n' %
-                       (time.strftime('%X') + str(timestamp % 1)[1:8],
-                        'INC     Mbps', 'Mpps', 'dropped',
-                        'OUT     Mbps', 'Mpps', 'dropped'))
+        cli.fout.write('{:<20}{:>14}{:>10}{:>10}        {:>14}{:>10}{:>10}\n'.format(
+                       time.strftime('%X') + str(timestamp % 1)[1:8],
+                       'INC     Mbps', 'Mpps', 'dropped', 'OUT     Mbps', 'Mpps', 'dropped'))
```
shinae-woo: Tiny thing: Could you change `dropped` to `Dropped`?

bessctl/commands.py
```diff
@@ -1781,18 +1781,26 @@ def get_total(arr):
...
             for port in ports:
                 last[port] = now[port]
+
+    try:
+        csv_path = os.getenv('CSV', None)
+        with open(csv_path, 'w') if csv_path is not None else noop() as csv_f:
+            if csv_f is not None:
+                csv_f.write('{}\n'.format(','.join(
+                    ('Timestamp', 'Port', 'Mbps_In', 'Mpps_In', 'Dropped_In', 'Mbps_Out', 'Mpps_Out', 'Dropped_Out'))))
+            print_loop(csv_f)
```
shinae-woo: Tiny things again, The output to `file` and `stdout` are slightly different in format and header name. Do you make it on purpose?

bessctl/commands.py
```diff
@@ -1863,15 +1864,15 @@ def print_delta(tc, delta, name_len):
...
 
     last = {}
     now = {}
 
     for tc in tcs:
         last[tc] = cli.bess.get_tc_stats(tc)
 
-    try:
+    def print_loop(csv=None):
         while True:
```
shinae-woo: How about moving this function right below the function `get_total`? Now the code flow is not easy to follow, as caller statements and function definitions are mixed.

bessctl/commands.py
```diff
@@ -1763,15 +1763,15 @@ def get_total(arr):
...
 
     last = {}
     now = {}
 
     for port in ports:
         last[port] = cli.bess.get_port_stats(port)
 
-    try:
+    def print_loop(csv_f=None):
         while True:
```
shinae-woo: Likewise

bessctl/commands.py
melvinw: The difference in the headers is intentional, but only because the text sent to `stdout` is more tabular and doesn't map cleanly to a csv. The values as read from left-to-right is meant to be equivalent, though

bessctl/commands.py
melvinw: Sure thing

bessctl/commands.py
melvinw: Good idea!

PR #860, opened 2018-10-08
dpdk: support newer versions of Mellanox OFED

sangjinhan:
Newer versions of OFED do not seem to have `verbs_exp.h` any longer, so it should not be used to detect whether OFED is installed. Also, `libibverbs` does not reference `libmlx4` or `libmlx5` now, so we should explicitly include them when linking `bessd`.

### DIFF
--- a/build.py
+++ b/build.py
@@ -227,9 +227,11 @@ def check_bnx():
 
 
 def check_mlx():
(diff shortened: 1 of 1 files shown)


shinae-woo:
Merged.

image

PR #864, opened 2018-10-22
gatehook: add interface for  gatehook commands in bessctl

shinae-woo:
```
$ command gatehook measuretshook out_outmod out 258 get_summary \
  MeasureCommandGetSummaryArg {'clear': False, 'latency_percentiles': [50, 99]}
```

Sorry, it's not having fancy auto-completion of commands at now.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -383,6 +383,14 @@ def get_var_attrs(cli, var_token, partial_word):
             except:
                 pass
 
+        elif var_token == 'GATEHOOK':
+            var_type = 'name'
+            var_desc = 'name of an existing gatehook instance'
+
+        elif var_token == 'GATEHOOK_CMD':
+            var_type = 'name'
+            var_desc = 'module command to run (see "show gatehookclass")'
+
         elif var_token == '[ENV_VARS...]':
             var_type = 'map'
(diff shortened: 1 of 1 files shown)


bessctl/commands.py
```diff
@@ -992,6 +1000,21 @@ def command_module(cli, module, cmd, arg_type, args):
         cli.bess.resume_all()
 
 
+@cmd('command gatehook GATEHOOK MODULE DIRECTION GATE GATEHOOK_CMD ARG_TYPE [CMD_ARGS...]',
+     'Send a command to a gatehook')
+def command_gatehook(cli, name, module, direction, gate, cmd, arg_type, args):
+    if args is None:
+        args = {}
+
+    cli.bess.pause_all()
```
chris3torek: Should we always pause here? I guess we do for all commands, it's only the add module/connection code that has a conditional pause. It's definitely not _wrong_ anyway.

chris3torek: Right - I think it would be good to have, in the interface that collects commands, a flag for "this command can be run while workers run" and then bessctl could use that to control its auto-pause, for instance. But that's way beyond this PR...

chris3torek:
image

PR #869, opened 2018-11-26
copy_bench: Fix compilation with libbenchmark 1.4.1

ddiproietto:
With libbenchmark 1.4.1 the semicolon is required.

https://github.com/google/benchmark/pull/495


ddiproietto:
Oh, I see what you're worried about, I didn't even think about problems with libbenchmark < 1.4.0.

Turns out we should be good to go, since we were using it with the semicolon in other places: https://github.com/NetSys/bess/blob/566a38b885a2d8aee3e68642628c46ad8f15869a/core/traffic_class_bench.cc#L237

PR #870, opened 2018-11-27
scheduler: Reset silent_drops in context before running task

ddiproietto:
silent_drops were not reset, causing a constant increase of deadend
packets even if no packets were processed:

```
localhost:10514 $ show worker
   Worker ID    Status  CPU core  # of TCs    Deadend pkts
           0   RUNNING         0         6     14946328395
```

This commit fixes it for the DefaultScheduler.

The ExperimentalScheduler doesn't appear to account for those at all.


ddiproietto:
I haven't benchmarked this, I hope it's not too expensive

shinae-woo:
Merged!
image

PR #872, opened 2018-12-13
 kmod: fix ndo_select_queue prototype for linux-4.19 

kevmitch:
Just add a branch to the existing #ifdef. There is no use of the changed
argument inside the function, so there are no substantive changes
required.

I verified it now builds on 4.9, 4.18 and 4.19

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,3 +57,4 @@ Please add your name to the end of this file and include this file to the PR, un
 * Eran Gampel
 * Tamás Lévai
 * Matthew Mussomele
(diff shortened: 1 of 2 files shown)


sangjinhan:
Thank you very much for this patch. If you don't mind, could you sign up for the CLA? (no secret strings attached)

kevmitch:
Sorry for the delay. Was on vacation. Working on the CLA. Need to get clearance from higher up.

kevmitch:
I see it's already fixed. Sorry, I couldn't make this work.

PR #878, opened 2019-01-11
moduletests: Force run bessd

sunghew:
Thanks to @shinae-woo for the proposed fix :)


sunghew:
Oops.`-k` is the default.

shinae-woo:
Sorry about it! But you definitely deserve a cat!

image

PR #879, opened 2019-01-11
bessctl: add reconnection to 'daemon start'

sangjinhan:
When bessctl tries to connects to the bessd gRPC server, it may be rejected with TCP RST if it is not fully up yet. Retry for 3 seconds in case of this transient failure. Since the time window of this race condition is very short (might happen only for a few milliseconds), 3 seconds should be enough.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -441,7 +441,8 @@ def get_var_attrs(cli, var_token, partial_word):
         else:
             raise
 
-    except cli.bess.APIError:
+    except (cli.bess.Error, cli.bess.APIError, cli.bess.RPCError):
+        # ignore errors, this is just auto completion
         pass
 
     if var_type is None:
@@ -692,10 +693,6 @@ def _do_start(cli, opts):
                            '\n')
         subprocess.check_call(cmd, shell='True')
     except subprocess.CalledProcessError:
-        try:
-            cli.bess.connect()  # reconnect to the old instance, if any
-        except cli.bess.APIError:
-            pass
         raise cli.CommandError('Cannot start BESS daemon')
     else:
         start = time.time()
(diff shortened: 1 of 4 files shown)


bessctl/commands.py
```diff
@@ -694,11 +694,18 @@ def _do_start(cli, opts):
     except subprocess.CalledProcessError:
         try:
             cli.bess.connect()  # reconnect to the old instance, if any
-        except:
+        except cli.bess.APIError:
```
chris3torek: Can RPCError occur here?

It might be nice to have a distinguished "connection failed or rejected" error as APIError is also used for bad arguments, but since we only wait 3 seconds that's not critical.

bessctl/commands.py
```diff
@@ -694,11 +694,18 @@ def _do_start(cli, opts):
...
     else:
-        cli.bess.connect()
+        start = time.time()
+        while time.time() - start < 3:
+            try:
+                cli.bess.connect()
+                break
+            except cli.bess.APIError:
+                # bessd is on, but its gRPC server may be not yet. Retry.
+                time.sleep(0.1)
```
chris3torek: We should probably test whether we're connected once we get out of the loop.  Other than that, this change itself seems fine.

More generally, I'm not sure but I think maybe this whole startup process should be reconstructed a bit.  I don't understand why we catch `subprocess.CalledProcessError` and try to connect to the old bessd - shouldn't failure to start a new one just be failure?

bessctl/commands.py
sangjinhan: `RPCError` cannot occur here, but maybe `APIError` should have been `RPCError` as you pointed out.

bessctl/commands.py
sangjinhan: Oh I thought I added it. Fixed.

Reconnecting to the old instance has hardly helped, since it would have already been killed by `bessd -k` with a 99.9% chance :) I removed it.

sangjinhan:
@chris3torek Thank you for the review! Addressed the issues.

sangjinhan:
where is my cat..?

sangjinhan:
@sunghew

PR #880, opened 2019-01-11
module_test: fix invalid IPv4 address format

sangjinhan:
recent versions of scapy seem to use libc `inet_pton` whenever possible
in place of `inet_aton`. `inet_pton` is much stricter when it comes to
formatting; "65.43.21.00" is illegal because of two 0s.


shinae-woo:
Merged, Thanks!
image

sangjinhan:
My guess is that pip is now fetching a newer version of scapy...

PR #883, opened 2019-01-22
unittest: adjust unittest timeout

sangjinhan:
Some default timeouts (about 2 seconds) turn out to be too short, especially when the system is highly overloaded.


shinae-woo:
image

PR #884, opened 2019-01-22
module_test: do not use hugepage for module test

sangjinhan:
hugepage is irrelevant since no physical NIC port is involved


shinae-woo:
image

PR #885, opened 2019-01-25
build: Use a make jobserver if available

sangjinhan:
build.py launches make to build DPDK and BESS. If build.py is launched
from a parent make process, the -jN option (along a few others) given
to the parent would not propagate to the submake processes.

This commit checks if build.py script was invoked from parent make,
and if so, propagates make options to submake processes.

### DIFF
--- a/build.py
+++ b/build.py
@@ -316,7 +316,10 @@ def build_dpdk():
 
     print('Building DPDK...')
     nproc = int(cmd('nproc', quiet=True))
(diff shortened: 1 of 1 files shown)


melvinw:
oh, yikes-- I didn't refresh before I merged and missed your comment, @chris3torek. sorry about that

melvinw:
Thanks! cat

PR #886, opened 2019-01-26
build: handle GNU make $MAKEFLAGS oddity

chris3torek:
It turns out that $MAKEFLAGS is sometimes spelled, e.g.,
"w -j4 --jobserver..." rather than "-w -j4 --jobserver...".
Check for this case and handle it.

Cache the result of the first makeflags() call in subsequent
calls.


sangjinhan:
file 2

PR #887, opened 2019-01-26
kmod: defer free_netdev() call to kernel on teardown for versions 4.11+

gomesj:
According to https://www.spinics.net/lists/netdev/msg439014.html,
priv_destructor should not perform free_netdev(), needs_free_netdev
should be set instead.


sangjinhan:
abc

PR #888, opened 2019-01-26
 build: disable `close_fds` for submake call

sangjinhan:
Unlike Python 2, the default value for the close_fds argument for Popen is true. This is problematic in case we want to use the GNU make jobserver, since it requires that file descriptors (pipe fds to jobserver) remain open in the submake process.


chris3torek:
image

PR #892, opened 2019-02-06
drivers/pmd: fix stats reporting for net_e1000_igb

melvinw:
I've been testing against I350 and I354 and they appear to not report per-queue stats. I'm not sure about 82575, 82576, 82580, I210, I211, and DH89xx though.


sangjinhan:
tumblr_mck3fdafmy1qze0hyo1_1280

PR #896, opened 2019-02-26
kmod: adapt ndo_select_queue prototype to linux-4.19

nrdmao:
Resubmitting this patch with since I have a signed CLA.

There is no use of the changed argument in the function, so there are no
substantive changes required.


nrdmao:
Any reason why this patch can't merge now?

sangjinhan:
Oh sorry. It seems that the CLA thing requires "all contributors" have singed it, as indicated by the error message below.

Issue #845, opened 2018-08-30
Page fault on vport delete


muraliran:
This is pretty consistent. If I recall well, this wasn't happening before vport refactor. This is ubuntu 18.04 vbox vm using generic pio. 

Cleaned and recompiled all (including kmod) but problem remains.

[  899.289069] BUG: unable to handle kernel paging request at ffffb185057118c0
[  899.289095] IP: sn_release_netdev+0x17/0x40 [bess] 

Steps to repro

1. Create docker
2. Attach vport (stadard bess procedure)
3. Delete docker instance (stop & rm)
4. bess hangs

Full trace:
~~~~~~~~~
[  829.645939] bess - sn_register_netdev():895 v_mlocal_1: registered - e6:8d:e7:03:44:92 txq 1 rxq 1
[  831.100724] bess - sn_register_netdev():895 v_mlocal_2: registered - 96:f7:38:c9:e2:bf txq 1 rxq 1
[  878.074034] docker0: port 1(veth4f435fe) entered disabled state
[  878.075258] veth3c9ec4b: renamed from eth0
[  878.108181] docker0: port 1(veth4f435fe) entered disabled state
[  878.111852] device veth4f435fe left promiscuous mode
[  878.111855] docker0: port 1(veth4f435fe) entered disabled state
[  878.216952] bess - sn_netdev_destructor():750 v_mlocal_1: releasing netdev...
[  899.289069] BUG: unable to handle kernel paging request at ffffb185057118c0
[  899.289095] IP: sn_release_netdev+0x17/0x40 [bess]
[  899.289106] PGD 217141067 P4D 217141067 PUD 217142067 PMD 20c720067 PTE 0
[  899.289123] Oops: 0000 [#1] SMP PTI
[...]

PR #899, opened 2019-03-09
Fix vport kernel panic when network namespace is closed

sangjinhan:
The network interface may be terminated involuntarily (i.e., other than calling SN_IOC_RELEASE_HOSTNIC). If that has happened, filp->private_data ends up pointing to freed memory space. When it is accessed by SN_IOC_RELEASE_HOSTNIC later, kernel may crash due to page fault. The problem can be easily reproduced when a network namespace is closed. VPort belonging to that namespace are also forcibly closed, thus ioctl(SN_IOC_RELEASE_HOSTNIC) later causes kernel panic.

This patch defers the memory space until its associated file/inode is released. One can test it by running `bessctl run port/vport_latency`.

### DIFF
--- a/core/kmod/sn_netdev.c
+++ b/core/kmod/sn_netdev.c
@@ -902,6 +902,11 @@ fail_free:
 
 void sn_release_netdev(struct sn_device *dev)
 {
+	if (!dev) {
+		log_err("called with a null pointer");
+		return;
+	}
+
 	rtnl_lock();
 
 	/* it is possible that the netdev has already been unregistered */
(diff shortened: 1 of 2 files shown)


sangjinhan:
This should fix #845

core/kmod/sn_netdev.c
```diff
@@ -913,10 +905,16 @@ void sn_release_netdev(struct sn_device *dev)
 	rtnl_lock();
 
 	/* it is possible that the netdev has already been unregistered */
-	if (dev && dev->netdev && dev->netdev->reg_state == NETREG_REGISTERED)
```
melvinw: Is it safe to assume `dev` is always non-null by the time we get here? Would it be worth checking and returning early before taking the lock?

sangjinhan: I believe it's safe... Not sure why I checked it in the past. As per your suggestion I added an early check, it should be better safe than sorry.

PR #900, opened 2019-03-12
Add an HTTP server for pipeline monitoring

sangjinhan:
`show pipeline` / `monitor pipeline` is a useful tool to monitor the BESS dataflow graph. However, the ascii diagram quickly becomes illegible as the graph grows in a non-trivial manner. One can generate an SVG file in the bessctl CLI, but it is cumbersome and not capable of real-time monitoring.

This PR creates a new bessctl command, `bessctl http [address] [port]`, which runs an HTTP server to show the current dataflow graph. Other features, such as port/TC monitoring, may be added later.

Screenshot from 2019-03-11 22-29-29

### DIFF
--- a/bessctl/static/pipeline.js
+++ b/bessctl/static/pipeline.js
@@ -1,0 +1,177 @@
+// colorscheme:
+// #01295f cool black
+// #437f97 queen blue
+// #849324 olive drab
+// #ffb30f dark tangerine
+// #fd151b vivid red
+
+// (node name, gate type, gate ID) -> [(timestamp, pkts, bits, cnt), ...]
+var stats = {};
+
+var opt_field;
+var opt_mode;
+
+function gates_to_str(gates, gate_type) {
+    var ret = '';
+
+    for (i = 0; i < gates.length; i++) {
+        gate_num = gates[i][gate_type]
+        if (gate_type == 'igate') {
+            color = '#437f97'
+        } else {
+            color = '#01295f'
+        }
+        ret += `<td port="${gate_type}${gate_num}" border="0" cellpadding="0" bgcolor="${color}"><font color="#ffffff" point-size="6">${gate_num}</font></td>
+`;
+    }
+
+    return `
+      <tr>
+        <td border="0" cellspacing="0" cellpadding="0">
+          <table border="0" cellborder="0" cellspacing="0" cellpadding="0">
+            <tr>
+              ${ret}
+            </tr>
+          </table>
+        </td>
+      </tr>`;
+}
+
+function add_datapoints(stats, module_name, gates, gate_type) {
+    for (var i = 0; i < gates.length; i++) {
+        var gate = gates[i];
+        key = [module_name, gate_type, gate[gate_type]];
+        value = {timestamp: gate.timestamp,
+                 bits: Number(gate.bytes * 8),
+                 pkts: Number(gate.pkts),
+                 cnt: Number(gate.cnt),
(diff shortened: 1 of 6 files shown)


pybess/bess.py
```diff
@@ -221,15 +222,14 @@ def connect(self, grpc_url=DEF_GRPC_URL):
     # returns no error if already disconnected
     def disconnect(self):
         try:
-            if self.is_connected():
+            if self.channel is not None:
```
chris3torek: I wonder if this is worth a comment - presumably the difference between `is_connected` and `channel is not None` is that `self.channel.unsubscribe` does something in the rare case that you have a channel but its status is not `grpc.ChannelConnectivity.READY`?

env/runtime.yml
```diff
@@ -18,6 +18,7 @@
           - protobuf
           - grpcio
           - scapy
+          - flask
```
chris3torek: commit message on this one is wrong, it should say `add flask` not `add pybess` (also this commit should probably go before the pybess commit, it looks like it's at the end right now)

pybess/bess.py
sangjinhan: Yeah, in such cases, it seemed that grpc keep retrying to connect in the background, with it own timeout. In the meantime, creating a new channel object doesn't do anything until the timeout fires. So I wanted to delete the channel so that a subsequent connection can be tried immediately.

env/runtime.yml
sangjinhan: :+1:

bessctl/static/pipeline.js
```diff
@@ -0,0 +1,177 @@
...
+
+function graph_to_dot(modules) {
+    opt_field = document.querySelector('input[name="metric"]:checked').value;
+    opt_mode = document.querySelector('input[name="mode"]:checked').value;
+    opt_humanreadable = document.querySelector('input[name="humanreadable"]').checked;
+
+    var nodes = ''
+    for (var module_name in modules) {
+        module = modules[module_name];
+        // add_datapoints(stats, module_name, module.igates, 'igate')
```
shinae-woo: Do you intentionally comment this line out? or having a plan to add the option or it?

sangjinhan: Yeah, at the moment it only shows the statistics numbers from output gates. Per-input-gate stats seemed less useful in general, so I didn't bother displaying them... so I didn't want to keep collecting the data. At some point I could add an option to show igate stats, but I have no plan now.

sangjinhan:
@ericcarino Could you please take a quick look at the javascript code to see if anything is (terribly) broken?

bessctl/static/pipeline.js
```diff
@@ -0,0 +1,177 @@
...
+var stats = {};
+
+var opt_field;
+var opt_mode;
+
+function gates_to_str(gates, gate_type) {
+    var ret = '';
+
+    for (i = 0; i < gates.length; i++) {
+        gate_num = gates[i][gate_type]
```
ericcarino: You are declaring variables like `gate_num` w/o the `var` keyword.  This is fine, but just be aware since you're not using `var`, the variable is global.

ericcarino: You do this several times, not sure if intentional or not?

sangjinhan: Oh, of course not. I didn't even know that it becomes global. I thought `var` in js is something optional just like `;` without any semantic difference :+1:

shinae-woo:
Thank you for doing this! 
image

PR #907, opened 2019-04-16
Mapping a logical internal/external ports to actual bess gates in nat

shinae-woo:
In nat module, packet flows are looks like
 internal --> 0:nat:0 --> external (forward direction)
 external --> 1:nat:1 --> internal (reverse direction)

This PR makes the bess gates are correspondint with the logical nat
ports, 0 for internal, and 1 for external, so the new packet flows will
looks like
 internal --> 0:nat:1 --> external (forward direction)
 external --> 1:nat:0 --> internal (reverse direction)

### DIFF
--- a/bessctl/conf/perftest/nat.bess
+++ b/bessctl/conf/perftest/nat.bess
@@ -49,6 +49,6 @@ Source() \
     -> nat::NAT(ext_addrs=nat_config)
 
 if bidirectional:
-    nat -> MACSwap() -> IPSwap() -> 1:nat:1 -> Sink()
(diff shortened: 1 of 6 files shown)


nemethf:
I think the examples in bessctl/conf will be outdated when this is merged.  Can you, please, update them as well?  Thank you.

shinae-woo:
@nemethf Ohh, thank you for catching this. Will update!

sangjinhan:
tumblr_lg90acBdUF1qfyzelo1_1280

PR #910, opened 2019-05-13
Fix vport initialization on recent kernels

kot-begemot-uk:
The default queue discipline on Linux is pfifo. It will fail
to initialize rendering the device unusable on recent kernels
if the device queue depth is zero. All devices with queue depth
of zero must also specify that they are IFF_NO_QUEUE in order
to select the noop scheduler.

This is from some point in 4.x onwards (I have not had the time
to trace the exact kernel revision).


kot-begemot-uk:
I have updated it accordingly.

sangjinhan:
Great, thank you very much.

tumblr_m1adotNrja1qbe5pxo1_1280

PR #911, opened 2019-05-13
Fix gtest include path

apanda:
We currently pass `$(GTEST_DIR)` as an include path using `-isystem`.
This should really be `$(GTEST_DIR)/include` (see
https://github.com/google/googletest/tree/master/googletest), since the
header files live inside the include directory. This doesn't show up as
an issue for many people since the files are also in `/usr/include` when
installed using the system path but is a problem when this is not the
case.


sangjinhan:
:+1: 
iE5F4AHiQl055004D08LgkL0o1_1280

PR #912, opened 2019-05-15
Queue occupancy histogram

dtnaylor:
Adds a histogram to the queue module which can be enabled to track statistics about queue occupancy.


protobuf/module_msg.proto
```diff
@@ -243,22 +243,11 @@ message MeasureCommandGetSummaryArg {
  * may not be a multiple of the resolution.
  */
 message MeasureCommandGetSummaryResponse {
```
shinae-woo: Could we avoid/minimize changing the protobuf message field names? There are many external project/repo using this, it's all broken. At least two of my project is not compatible with these changes.

PR #914, opened 2019-05-18
tests: skip unit tests that require extra capabilities

mukerjee:
5 unit tests require more capabilities than are available in docker build. Let's
skip them if they don't work.


sangjinhan:
thanks!
Jjkybd3nS9drgi2tzky6EcG4_400

PR #916, opened 2019-05-21
Fix kmod in 5.0+ kernels

shague:
5.0 kernels included a new parameter in the dev_open()
function. Fixed by adding a #if/#else block for the
newer kernel and pass new parameter as NULL.

Signed-off-by: Sam Hague <shague@gmail.com>


sangjinhan:
tumblr_m1agbr834T1qze0hyo1_1280

PR #924, opened 2019-06-20
do not consolidate non-worker threads in one core

sangjinhan:
At the moment, all non-worker threads are isolated within a single core, typically in the last core. This is not what the user asked for. Instead, non-worker threads should be able to run on any cores, or some particular set of cores explicitly set by the user (via taskset, etc.)

### DIFF
--- a/core/dpdk.cc
+++ b/core/dpdk.cc
@@ -98,13 +98,13 @@ class CmdLineOpts {
   std::vector<char *> argv_;
 };
 
-void init_eal(int dpdk_mb_per_socket, int default_core) {
+void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
   CmdLineOpts rte_args{
       "bessd",
       "--master-lcore",
       std::to_string(RTE_MAX_LCORE - 1),
       "--lcore",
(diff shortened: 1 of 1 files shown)


shinae-woo:
image

krsna1729:
Nice. This allows me to get rid of the [entrypoint](https://github.com/krsna1729/bess-router/blob/master/entrypoint.sh) script I had just to preserve the task affinity that was set.

krsna1729:
But seems adding a worker pinned to one from the tasket, does not move all the other threads away from the core selected for the worker. Wondering the performance affects

```
root@144b99d9fcd8:/opt/bess/bessctl# ps -T -p 1
  PID  SPID TTY          TIME CMD
    1     1 pts/0    00:00:06 bessd
    1     7 pts/0    00:00:00 vfio-sync
    1    52 pts/0    00:00:00 eal-intr-thread
    1    54 pts/0    00:00:00 bessd
    1    83 pts/0    00:00:00 bessd
    1    88 pts/0    00:05:55 bessd
```
```
pid 1's current affinity mask: 3000
pid 7's current affinity mask: 3000
pid 52's current affinity mask: 3000
pid 54's current affinity mask: 3000
pid 83's current affinity mask: 3000
pid 88's current affinity mask: 1000 # worker
```

PR #927, opened 2019-06-28
Make L2Forward and Replicate changes thread safe and transactional

### LINKED ISSUE #920: Thread safety in configuration requests
Hi,

I was looking into implementing a simple L2 switch PoC using BESS as an engine and ran into the fact that most config changes in the L2Forward and IPForward are marked as non-thread-safe and thus require all workers for the modules to be stopped.

Is stopping and restarting the pipeline for such config change the intended mode of operation or we should look into making at least some of these thread safe?

kot-begemot-uk:
Fixes #920 for L2Forward and Replicate, demonstrates a pattern applicable as a fix to most other modules.

Introduces a very common check for "port hopping" into the L2Forward module and defines it as an option

Adds the clear command to the L2Forward module so it is functionally similar to the IPForward module.

### DIFF
--- a/core/kmod/sn_netdev.c
+++ b/core/kmod/sn_netdev.c
@@ -588,12 +588,6 @@ static int sn_start_xmit(struct sk_buff *skb, struct net_device *netdev)
 		return NET_XMIT_DROP;
 	}
 
-	if (unlikely(skb_shinfo(skb)->frag_list)) {
-		log_err("frag_list is not NULL!\n");
-		dev_kfree_skb(skb);
-		return NET_XMIT_DROP;
-	}
-
 	if (unlikely(txq >= dev->num_txq)) {
 		log_err("invalid txq=%u\n", txq);
 		dev_kfree_skb(skb);
@@ -688,9 +682,8 @@ extern const struct ethtool_ops sn_ethtool_ops;
 
 static void sn_set_offloads(struct net_device *netdev)
 {
-	netif_set_gso_max_size(netdev, SNBUF_DATA);
-
 #if 0
+	netif_set_gso_max_size(netdev, SNBUF_DATA);
 	netdev->hw_features = NETIF_F_SG |
 			      NETIF_F_IP_CSUM |
 			      NETIF_F_RXCSUM |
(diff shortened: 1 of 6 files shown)


sangjinhan:
Thank you very much! At the moment I am quite busy; will soon come back and review the code.

krsna1729:
I have wondered if BESS framework itself could handle `THREAD_UNSAFE` commands in a safe way, since it implements the scheduler in every worker thread. Today I think the master thread receives the commands and executes them. Instead could it queue them to individual worker threads? The scheduler code after every round or num rounds, can dequeue and run the commands? This way we need not stop? This should be good enough if only one worker is accessing the module instance, right?

kot-begemot-uk:
That is possible, but that does not solve transactional configuration changes. For that you need to support it at config level. In the absence of a global config framework each module has to support these individually.

There are also issues in a multi-worker environment. If a worker has finished its batch and there is a config change pending it needs to stop, because the scheduler can allow the config thread to do changes only once all workers have finished using the module. 

All in all, it will be very nice if the scheduler does that by default for all UNSAFE commands as a workaround. 

It is however a workaround - some stuff like route changes in L2 and L3 should have no stops as they are happening many times a second under normal operating conditions.

krsna1729:
I think this loop would be tens of million times a second so should be ok?

https://github.com/NetSys/bess/blob/7c368cfccacc6a2497e975bb87fb68264970fa1a/core/scheduler.h#L326-L339

kot-begemot-uk:
Stopping workers this way and automatically restarting them without a restart request as a default action for UNSAFE is OK as a default behaviour. In fact, we should do it, but it is something orthogonal to this change.
 
1. Turning workers on/off does not make changes transactional. The proposed changes do. 

If bulked configuration changes are allowed they have to be transactional especially for a module which does not have any means to read its current state like L2Forward. 

At present if you try to change and one of the entries fails, you have no idea which one failed, you have absolutely no idea what is the state of the module, you have no means to read it back and no means to clear it short of destroying the module and rewiring the whole pipeline to a new instance.

2. The changes are still significantly faster and with lower impact than a default worker stop/worker start. As I said - we should do that as a default, but that does not mean that individual modules should not mark particular changes as SAFE and use faster methods.

krsna1729:
Agreed :)

PR #930, opened 2019-07-15
Fix driver scatter gather and kernel driver flags

kot-begemot-uk:
The driver was being initialized incorrecttly.

1. Any setting of gso_max_size also implicitly turns on GSO
which is not supported resulting in transfer speeds as low as
85MBit for TCP

2. The code for Scatter Gather is correct, just suboptimal
(extra copy). There is no reason not to enable it.


sangjinhan:
Thanks!
qgIb8tERiqb59t8bFhf7qJnvo1_500

PR #932, opened 2019-07-15
Add Vector TX Support

kot-begemot-uk:
Add support for vector/batched TX.

RX to be done separately as it requires packet pre-allocation.

Increases packet throughput > 10 times under normal conditions, more in some cases.

### DIFF
--- a/core/drivers/unix_socket.cc
+++ b/core/drivers/unix_socket.cc
@@ -158,6 +158,18 @@ CommandResponse UnixSocketPort::Init(const bess::pb::UnixSocketPortArg &arg) {
     return CommandFailure(errno, "unable to start accept thread");
   }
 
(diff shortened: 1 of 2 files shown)


kot-begemot-uk:
I just pushed the second half - the RX portion.

sangjinhan:
Sorry for this late response. Could you check https://github.com/kot-begemot-uk/bess/pull/4 ?

(also, please check the CLA agreement)

kot-begemot-uk:
The corresponding part for User Mode Linux just went into 5.4-rc1. It already had vector paravirtual nic so I only needed to add the protocol. 

It is can now talk the bess unix domain socket protocol at similar speeds (a few Gbit).

sangjinhan:
Thanks! 0f50AIVp99uy4k64bERW0vHJ_250

PR #937, opened 2019-08-23
do not crash even if rte_socket_id() fails

sangjinhan:
In some container scenario, it was observed that rte_socket_id() may return -1 (SOCKET_ID_ANY)

### DIFF
--- a/core/worker.cc
+++ b/core/worker.cc
@@ -291,7 +291,14 @@ void *Worker::Run(void *_arg) {
   wid_ = arg->wid;
   core_ = arg->core;
   socket_ = rte_socket_id();
-  CHECK_GE(socket_, 0);  // shouldn't be SOCKET_ID_ANY (-1)
+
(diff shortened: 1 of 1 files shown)


core/worker.cc
```diff
@@ -291,7 +291,14 @@ void *Worker::Run(void *_arg) {
   wid_ = arg->wid;
   core_ = arg->core;
   socket_ = rte_socket_id();
-  CHECK_GE(socket_, 0);  // shouldn't be SOCKET_ID_ANY (-1)
+
+  // For some reason, rte_socket_id() does not return a correct NUMA ID.
```
chris3torek: This is a bit troubling, I can't see why `rte_socket_id` should fail here, but the change itself seems OK.

sangjinhan: Oh my bad. "does not" should have been "may not"

chris3torek: I figured that - obviously this _usually_ works. Under what condition, I wonder, is it failing?

PR #940, opened 2019-09-17
There are more than 64 Core/Thread CPUs nowdays

kot-begemot-uk:
Rather obvious one. Without it bess crashes on a two socket Xeon which claims to have 36 Cores/72 threads.


sangjinhan:
Thanks!
iE5F4AHiQl055004D08LgkL0o1_1280

PR #945, opened 2019-11-04
Fix kernel module build for kernels >= 5.2.0

kot-begemot-uk:

### DIFF
--- a/core/kmod/sn_netdev.c
+++ b/core/kmod/sn_netdev.c
@@ -610,6 +610,10 @@ static u16 sn_select_queue(struct net_device *netdev,
 			   struct sk_buff *skb,
 			   void *accel_priv,
 			   select_queue_fallback_t fallback)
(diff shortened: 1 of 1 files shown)


core/kmod/sn_netdev.c
```diff
@@ -610,6 +610,10 @@ static u16 sn_select_queue(struct net_device *netdev,
 			   struct sk_buff *skb,
 			   void *accel_priv,
 			   select_queue_fallback_t fallback)
+#elif LINUX_VERSION_CODE >= KERNEL_VERSION(5,2,0)
```
sangjinhan: (million years late, sorry)

IMO the current order is somewhat hard to follow. Perhaps this order would be better?

```
...
#elif version < 4.18.0
   ...
#elif version < 5.2.0
   ...
#else 
   ...
#endif

PR #946, opened 2019-11-04
Allows specifying socket/numa node for BESS

trozet:
Using "-n" flag a user may now specify the desired socket to allocate
huge pages from when using BESS.

Also includes a check for worker core existing on the correct socket
before trying to launch a worker. If worker is launched on a core on an
unused socket GetDefaultPool will crash bessd.

Signed-off-by: Tim Rozet <trozet@redhat.com>

### DIFF
--- a/core/bessctl.cc
+++ b/core/bessctl.cc
@@ -567,6 +567,12 @@ class BESSControlImpl final : public BESSControl::Service {
     if (!is_cpu_present(core)) {
       return return_with_error(response, EINVAL, "Invalid core %d", core);
     }
+   
+    if (rte_lcore_to_socket_id(core) != (unsigned int) FLAGS_n) {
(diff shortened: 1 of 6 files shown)


core/worker.cc
```diff
@@ -290,15 +290,20 @@ void *Worker::Run(void *_arg) {
...
+    LOG(ERROR) << "Core specified: " << core_ << " does not exist on socket: "
+    << socket_ << ". Cannot create worker";
+
+    delete scheduler_;
+    delete rand_;
+
+    return nullptr;
   }
 
+  LOG(INFO) << "Running worker: " << wid_ << " on core " << core_ << " on socket " << socket_;
```
melvinw, line 305: nit: perhaps demote this to `VLOG(1)`

PR #949, opened 2019-12-06
bessctl: expose deadend packets per module

melvinw:


core/module.h
```diff
@@ -441,6 +443,7 @@ class alignas(64) Module {
 
   std::vector<bess::IGate *> igates_;
   std::vector<bess::OGate *> ogates_;
+  uint64_t deadends_;
```
nrdmao: Shouldn't you have a counter per worker as multiple workers may be running through the module at the same time. As such deadends() would return the sum of the counters for each worker.

melvinw: good catch! adjusted

core/module.h
```diff
@@ -495,13 +506,15 @@ inline void Module::RunChooseModule(Context *ctx, gate_idx_t ogate_idx,
...
   if (unlikely(ogate_idx >= ogates_.size())) {
+    deadends_[ctx->wid]++;
     deadend(ctx, batch);
     return;
   }
 
   ogate = ogates_[ogate_idx];
 
   if (unlikely(!ogate)) {
+    deadends_[ctx->wid]++;
```
sangjinhan: ditto

core/module.h
```diff
@@ -327,6 +329,14 @@ class alignas(64) Module {
 
   const std::vector<bess::OGate *> &ogates() const { return ogates_; }
 
+  uint64_t deadends() const {
+    uint64_t ret = 0;
```
sangjinhan: if you like: `return std::accumulate(deadends_, deadends_ + kMaxWorkers, 0);`

core/module.h
```diff
@@ -441,6 +451,7 @@ class alignas(64) Module {
 
   std::vector<bess::IGate *> igates_;
   std::vector<bess::OGate *> ogates_;
+  uint64_t deadends_[Worker::kMaxWorkers];
```
sangjinhan: if you like: `std::array<uint64_t, kMaxWorkers> deadends_`

core/module.h
sangjinhan: (or `.begin()` and `.end()` for an `std::array`)

melvinw:
Thanks @sangjinhan! Updated

core/module.h
```diff
@@ -327,6 +331,10 @@ class alignas(64) Module {
 
   const std::vector<bess::OGate *> &ogates() const { return ogates_; }
 
+  uint64_t deadends() const {
```
shinae-woo: I guess we want per-worker statistics here either array or map. The caller (`bessctl.cc`) able to accumulation if required.

melvinw: In my head `deadends_` is just an array to make the counter thread-safe-- i.e., it's just an implementation detail. I only ever intended to expose a single counter to the user here

melvinw:
Nice catch @shinae-woo! Pushed changes to `RunSplit()` and `ResetActiveWorkers()`

shinae-woo:
Thank you! LGTM for the part I commented on. @sangjinhan Do you wanna take a look again and merge this?

PR #952, opened 2019-12-12
Fix relative imports

mukerjee:
1) allow module_tests and bessctl to use relative paths even when those files are in symlink'd directories
2) re-work how auto-generated protobuf files are imported so that they're treated as external modules

These commits together fix situations where the source tree for bess and the autogenerated protobuf files don't live in the same directory, but are instead symlinked into a third directory. Build systems like bazel do this sort of symlink madness.

tested by running the module test and having bessctl use python2 and python3

### DIFF
--- a/pybess/bess.py
+++ b/pybess/bess.py
@@ -48,15 +48,18 @@ from . import pm_import as _pm
 # Ugh: builtin_pb and plugin_pb must be on path, as protoc generates python code
 # that assumes it can import files in that directory.
 old_path = list(sys.path)
+p = os.path.abspath(os.path.dirname(__file__))
+if p not in sys.path:
+    sys.path.append(p)
 for extra in ('builtin_pb', 'plugin_pb'):
     p = os.path.abspath(os.path.join(__file__, '..', extra))
     if p not in sys.path:
-        sys.path.insert(1, p)
+        sys.path.append(p)
 del extra, p
 
-from .builtin_pb import service_pb2_grpc
-from .builtin_pb import bess_msg_pb2 as bess_msg
-from .builtin_pb import module_msg_pb2 as module_msg
+from builtin_pb import service_pb2_grpc
+from builtin_pb import bess_msg_pb2 as bess_msg
+from builtin_pb import module_msg_pb2 as module_msg
 
 
 def _import_modules(name, subdir):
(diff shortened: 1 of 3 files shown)


sangjinhan:
Thanks for the patch. Certainly this is my original sin w/ writing the related code without fully understanding Python module/package/subpackage/etc., especially when it comes to protobuf auto generation. Admittedly I still don't fully understand them.

#1 (realpath -> abspath) sounds very straightforward to me.

#2 is something I am trying to understand. Could you elaborate on why relative imports now need to be absolute? Also, I am curious whether lines 54-57 are still relevant after the change.

pybess/bess.py
```diff
@@ -48,15 +48,18 @@
 # Ugh: builtin_pb and plugin_pb must be on path, as protoc generates python code
 # that assumes it can import files in that directory.
 old_path = list(sys.path)
+p = os.path.abspath(os.path.dirname(__file__))
+if p not in sys.path:
+    sys.path.append(p)
```
chris3torek: You might want to `insert` at 1 here for efficiency-and-override effect (same as in the loop below), or else change all of them to use `append` - it seems a bit weird/inconsistent now.

mukerjee:
Thanks for taking a look at this guys. I also can't say I'm an expert on python package management.

@sangjinhan 
re L54-57: This aren't necessary for python2 but it looks like import statements are handled differently in python3. Removing these lines breaks `from builtin_pb import service_pb2_grpc` for python3.

In general using absolute imports here allows a user of bessctl to provide additional paths (by means of PYTHONPATH) where the generated protobuf python files might live. This makes bessctl more compatible with other build systems that may place the protobufs in a different folder than the bess source e.g., bazel. 

By the time we get to bess.py , `__file__` has already become a real path (i.e., it's resolved the symlink), so relative imports here (I think) will look in the bess source tree rather than using `sys.path`. It seems that the behavior of `__file__` is notoriously complex: https://stackoverflow.com/questions/7116889/is-module-file-attribute-absolute-or-relative

using absolute imports imho is the way to go here, since it makes it overridable.

pybess/bess.py
mukerjee: Ah-- they probably all should be an `append`, otherwise things a user adds to PYTHONPATH will come after the `insert(1, p)`. This would mean that PYTHONPATH won't properly override. For my specific use-case, I actually only need to override the first one (L53), but we should do all of them for consistency.

pybess/bess.py
```diff
@@ -70,20 +73,22 @@ def _import_modules(name, subdir):
...
+        for path in sys.path:
+            if not os.path.exists(os.path.join(path, 'builtin_pb')):
+                continue
+            for directory in ('builtin_pb', 'plugin_pb'):
+                if subdir:
+                    dirpath = os.path.join(path, directory, subdir)
+                    prefix = '{}.{}'.format(directory, subdir)
+                else:
+                    dirpath = os.path.join(path, directory)
+                    prefix = '{}'.format(directory)
```
sunghew: nit: This can be
```
prefix = directory
```

pybess/bess.py
```diff
@@ -70,20 +73,22 @@ def _import_modules(name, subdir):
...
-                dirpath = os.path.join(self_path, directory)
-                prefix = '..{}'.format(directory)
-            for filename in os.listdir(dirpath):
-                if not filename.endswith('_msg_pb2.py'):
-                    continue
-                import_name = '{}.{}'.format(prefix, filename[:-3])
-                if import_name != '..builtin_pb.bess_msg_pb2':
-                    yield import_name
+        for path in sys.path:
+            if not os.path.exists(os.path.join(path, 'builtin_pb')):
```
sunghew: Shouldn't this be:
```
if not os.path.exists(os.path.join(path, 'builtin_pb')) or os.path.exists(os.path.join(path, 'plugin_pb'))
```

mukerjee: probably `if not A or not B` but, yeah good point.

pybess/bess.py
mukerjee: This one i'm going to leave as-is because it better matches the form of the one above it.

PR #957, opened 2020-01-20
a few config inspection patches

melvinw:


core/modules/ip_checksum.cc
```diff
@@ -34,11 +34,20 @@
 #include "../utils/ether.h"
 #include "../utils/ip.h"
 
+const Commands IPChecksum::cmds = {
+    {"get_initial_arg", "EmptyArg", MODULE_CMD_FUNC(&IPChecksum::GetInitialArg),
+     Command::THREAD_SAFE},
+};
+
+CommandResponse IPChecksum::GetInitialArg(const bess::pb::EmptyArg &) {
+  return CommandSuccess();
```
chris3torek: As Shinae mentioned elsewhere (offline) we can make this the default. There's a minor danger that someone will then forget to implement it when using a non-trivial initial arg, but I'm not sure if it's worth protecting against this case.

core/modules/set_metadata.h
```diff
@@ -34,11 +34,15 @@
...
+using bess::metadata::mt_offset_t;
 
-typedef struct { uint8_t bytes[kMetadataAttrMaxSize]; } value_t;
-typedef struct { uint8_t bytes[kMetadataAttrMaxSize]; } mask_t;
+typedef struct {
+  uint8_t bytes[kMetadataAttrMaxSize];
+} value_t;
+typedef struct {
+  uint8_t bytes[kMetadataAttrMaxSize];
+} mask_t;
```
chris3torek: (Huh, did clang-format change?)

melvinw: oh, boo it did. didn't even noticed when i scanned the diff earlier

core/modules/ip_checksum.cc
melvinw: Pushed a fixup with this. Isn't the cleanest, but I'm not sure we can do much better atm

PR #958, opened 2020-01-23
Add pybess API to configure port

shinae-woo:
This commit 
1) verifies a MTU input ranges and add responses, and
2) adds a pybess API to call `SetPortConfRequest()` which configure a port by name for mac_address, MTU, and enable/disable the port.

### DIFF
--- a/bessctl/commands.py
+++ b/bessctl/commands.py
@@ -1031,6 +1031,17 @@ def command_gatehook(cli, name, module, direction, gate, cmd, arg_type, args):
         cli.bess.resume_all()
 
 
+# Please do not rely on this API. This API may be replaced with `command port PORT`
+# in the same way with gatehook and module
+@cmd('configure port PORT [PORT_ARGS...]', '[Experimental] Update a port configuration')
+def conigure_port(cli, name, args):
+    cli.bess.pause_all()
+    try:
+        cli.bess.ser_port_config(name, args or {})
+    finally:
+        cli.bess.resume_all()
+
+
 @cmd('delete worker WORKER_ID...', 'Delete a worker')
 def delete_worker(cli, wids):
     wids = sorted(list(set(wids)))
@@ -1440,8 +1451,10 @@ def _show_port(cli, port):
     else:
         autoneg = 'OFF'
 
-    cli.fout.write('  %-12s Driver %-10s HWaddr %s\n' %
-                   (port.name, port.driver, port.mac_addr))
+    port_config = cli.bess.get_port_config(port.name)
+
+    cli.fout.write('  %-12s Driver %-10s HWaddr %-18s MTU %-6d\n' %
+                   (port.name, port.driver, port.mac_addr, port_config.conf.mtu))
     cli.fout.write('  %-12s Speed %-11s Link %-5s Duplex %-5s Autoneg %-5s\n' %
                    ('', speed, link, duplex, autoneg))
     stats = cli.bess.get_port_stats(port.name)
--- a/core/drivers/pmd.cc
+++ b/core/drivers/pmd.cc
@@ -324,16 +324,20 @@ CommandResponse PMDPort::Init(const bess::pb::PMDPortArg &arg) {
   return CommandSuccess();
 }
 
-int PMDPort::UpdateConf(const Conf &conf) {
+CommandResponse PMDPort::UpdateConf(const Conf &conf) {
   rte_eth_dev_stop(dpdk_port_id_);
 
   if (conf_.mtu != conf.mtu && conf.mtu != 0) {
+    if (conf.mtu > SNBUF_DATA || conf.mtu < ETHER_MIN_MTU) {
+      return CommandFailure(EINVAL, "mtu should be >= %d and <= %d",
+                            ETHER_MIN_MTU, SNBUF_DATA);
+    }
+
     int ret = rte_eth_dev_set_mtu(dpdk_port_id_, conf.mtu);
     if (ret == 0) {
       conf_.mtu = conf.mtu;
     } else {
-      LOG(WARNING) << "rte_eth_dev_set_mtu() failed: " << rte_strerror(-ret);
-      return ret;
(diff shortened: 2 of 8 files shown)


pybess/port.py
```diff
@@ -52,3 +52,7 @@ def get_port_stats(self):
 
     def get_link_status(self):
         return self.bess.get_link_status(self.name)
+
+    def set_port_config(self, **kwargs):
```
chris3torek: Do we want to add this to `bessctl/commands.py` as well? (Otherwise it can only be invoked from scripts)

core/drivers/pmd.cc
```diff
@@ -324,16 +324,20 @@ CommandResponse PMDPort::Init(const bess::pb::PMDPortArg &arg) {
...
+    }
+
     int ret = rte_eth_dev_set_mtu(dpdk_port_id_, conf.mtu);
     if (ret == 0) {
-      conf_.mtu = conf_.mtu;
+      conf_.mtu = conf.mtu;
     } else {
-      LOG(WARNING) << "rte_eth_dev_set_mtu() failed: " << rte_strerror(-ret);
-      return ret;
+      return CommandFailure(-ret, "rte_eth_dev_set_mtu() failed");
```
melvinw: It would be helpful to include the output of `rte_strerror()` here, too. And shouldn't we keep the original `LOG(WARNING)`?

core/drivers/pmd.cc
```diff
@@ -345,9 +349,7 @@ int PMDPort::UpdateConf(const Conf &conf) {
     if (ret == 0) {
       conf_.mac_addr = conf.mac_addr;
     } else {
-      LOG(WARNING) << "rte_eth_dev_default_mac_addr_set() failed: "
-                   << rte_strerror(-ret);
-      return ret;
+      return CommandFailure(-ret, "rte_eth_dev_default_mac_addr_set() failed");
```
melvinw: same note about `rte_strerror()` and the log here

core/drivers/pmd.cc
```diff
@@ -356,12 +358,11 @@ int PMDPort::UpdateConf(const Conf &conf) {
     if (ret == 0) {
       conf_.admin_up = true;
     } else {
-      LOG(WARNING) << "rte_eth_dev_start() failed: " << rte_strerror(-ret);
-      return ret;
+      return CommandFailure(-ret, "rte_eth_dev_start() failed");
```
melvinw: same note about `rte_strerror()` and the log here

bessctl/commands.py
```diff
@@ -1031,6 +1031,18 @@ def command_gatehook(cli, name, module, direction, gate, cmd, arg_type, args):
         cli.bess.resume_all()
 
 
+@cmd('configure port PORT [PORT_ARGS...]', 'Send a command to a gatehook')
```
melvinw: Command description could use updating :)

core/drivers/pmd.h
```diff
@@ -127,7 +127,7 @@ class PMDPort final : public Port {
 
   LinkStatus GetLinkStatus() override;
 
-  int UpdateConf(const Conf &conf) override;
+  CommandResponse UpdateConf(const Conf &conf) override;
```
melvinw: Are we sure this change won't break any outside drivers?

shinae-woo: Yes :)

core/drivers/pmd.cc
shinae-woo: In a case of RPC call (e.g., through `bessctl`), returning `CommandResponse` without warning is one of the common patterns used in this repo. `CommandFailure()` or `CommandSuccess()` will be returned to the RPC caller, so that the caller should log that result if necessary. (Refer the code structures in `besctl.cc` https://github.com/NetSys/bess/blob/9cfa4bb188ebde987b43a617fdbd096471888314/core/bessctl.cc#L1039)

bessctl/commands.py
```diff
@@ -1031,6 +1031,20 @@ def command_gatehook(cli, name, module, direction, gate, cmd, arg_type, args):
         cli.bess.resume_all()
 
 
+# Please do not rely on this API. This API may replaced with `command port PORT`
+# in the same way with gatehook and module
+@cmd('configure port PORT [PORT_ARGS...]', '[Experimental] Update a port configuration')
+def conigure_port(cli, name, args):
+    if args is None:
```
sangjinhan: style thing, feel free to ignore: `cli.bess.ser_port_config(name, args or {})` is shorter, without hurting readability (imo).

protobuf/service.proto
```diff
@@ -204,7 +204,7 @@ service BESSControl {
   rpc DestroyPort (DestroyPortRequest) returns (EmptyResponse) {}
 
   /// Runtime-updatable configuration
-  rpc SetPortConf (SetPortConfRequest) returns (EmptyResponse) {}
+  rpc SetPortConf(SetPortConfRequest) returns (CommandResponse) {}
```
sangjinhan: nit: hm, a whitespace was removed.

pybess/bess.py
```diff
@@ -393,6 +393,19 @@ def destroy_port(self, name):
         request.name = name
         return self._request('DestroyPort', request)
 
+    def set_port_config(self, name, arg):
+        request = bess_msg.SetPortConfRequest()
+        request.name = name
+        request.conf.mac_addr = arg.pop('mac_addr', '00:00:00:00:00:00')
```
sangjinhan: a single comment line would be nice, something like "no update is done if mac_addr or mtu is not specified".

protobuf/service.proto
chris3torek: (Interesting, we don't run these through clang-format?)

shinae-woo: Oh, I see, maybe we don't lint protobuf file. FYI. running clang-format-5.0 gave me this: (Removing a bunch of spaces). I will add back space here.

image

sangjinhan:
Thanks
tumblr_mcgyk1QuXu1rppsk3o1_1280

PR #966, opened 2020-03-30
build.py: fix make invocations

chris3torek:
The MAKEFLAGS environment variable may hold `--` inside it, so be
sure to put our own `-C` and variable-setting arguments earlier.


sangjinhan:
IMG_20200330_163830

PR #968, opened 2020-04-04
Set IOVA mode explicitly

sangjinhan:
... to PA or VA, based on whether hugepages are being used or not.
Otherwise, what mode is used heuristically depends on the current
current configuration of the system, sometimes ends up with not
what we want (e.g., PA mode is used even when run by non-root users)

https://software.intel.com/en-us/articles/memory-in-dpdk-part-2-deep-dive-into-iova

---

Without this patch, bessd (or unit tests) may crash if
* some NICs are bound to `vfio-pci`,
* `/sys/module/vfio/parameters/enable_unsafe_noiommu_mode` is set,
* bessd is run by non-privileged user without hugepages.

### DIFF
--- a/core/dpdk.cc
+++ b/core/dpdk.cc
@@ -91,6 +91,16 @@ class CmdLineOpts {
 
   int Argc() const { return args_.size(); }
 
+  std::string Dump() {
+    std::ostringstream os;
+    os << "[";
+    for (size_t i = 0; i < args_.size(); i++) {
+      os << (i == 0 ? "" : ", ") << '"' << args_[i].data() << '"';
+    }
+    os << "]";
+    return os.str();
+  }
+
  private:
   // Contains a copy of each argument.
   std::vector<std::vector<char>> args_;
@@ -148,6 +158,7 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
   stdout = fopencookie(nullptr, "w", dpdk_log_init_funcs);
 
   disable_syslog();
+  LOG(INFO) << "Initializing DPDK EAL with options: " << rte_args.Dump();
   int ret = rte_eal_init(rte_args.Argc(), rte_args.Argv());
   if (ret < 0) {
     LOG(FATAL) << "rte_eal_init() failed: ret = " << ret
@@ -218,7 +229,6 @@ void InitDpdk(int dpdk_mb_per_socket) {
 
   if (!is_initialized) {
     is_initialized = true;
-    LOG(INFO) << "Initializing DPDK";
     init_eal(dpdk_mb_per_socket, GetNonWorkerCoreList());
   }
 }
--- a/core/dpdk.cc
+++ b/core/dpdk.cc
@@ -118,9 +118,12 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
       // Do not bother with /var/run/.rte_config and .rte_hugepage_info,
       // since we don't want to interfere with other DPDK applications.
       "--no-shconf",
+      // TODO(sangjin) switch to dynamic memory mode
+      "--legacy-mem",
   };
 
   if (dpdk_mb_per_socket <= 0) {
+    rte_args.Append({"--iova", "va"});
     rte_args.Append({"--no-huge"});
 
     // even if we opt out of using hugepages, many DPDK libraries still rely on
@@ -128,6 +131,8 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
     // memory in advance. We allocate 512MB (this is shared among nodes).
     rte_args.Append({"-m", "512"});
   } else {
+    rte_args.Append({"--iova", "pa"});
+
     std::string opt_socket_mem = std::to_string(dpdk_mb_per_socket);
     for (int i = 1; i < NumNumaNodes(); i++) {
       opt_socket_mem += "," + std::to_string(dpdk_mb_per_socket);
--- a/core/dpdk.cc
+++ b/core/dpdk.cc
@@ -46,6 +46,7 @@
 #include <string>
 
 #include "memory.h"
+#include "opts.h"
 #include "worker.h"
 
 namespace bess {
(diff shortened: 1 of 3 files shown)


krsna1729:
i have had similar issues before, never dug deep, but as a workaround was using --legacy-mem. In the ONF environment we are currently using bess with vfio-pci on baremetal k8s without privileged

krsna1729:
does not seem to be working in our setup

```
I0404 01:10:36.868716     1 dpdk.cc:166] Initializing DPDK EAL with options: ["bessd", "--master-lcore", "127", "--lcore", "127@12-13", "--no-shconf", "--legacy-mem", "--iova", "pa", "--socket-mem", "1024,1024", "--huge-unlink"]
EAL: Detected 88 lcore(s)
EAL: Detected 2 NUMA nodes
EAL: FATAL: Cannot use IOVA as 'PA' since physical addresses are not available
EAL: Cannot use IOVA as 'PA' since physical addresses are not available
F0404 01:10:36.908143     1 dpdk.cc:169] rte_eal_init() failed: ret = -1 rte_errno = 22 (Invalid argument)
*** Check failure stack trace: ***
F0404 01:10:36.924130     1 debug.cc:405] Backtrace (recent calls first) ---
(0): bessd(+0x4db57c) [0x55b7c364257c]
(1): bessd(_ZN4bess8InitDpdkEi+0x6b) [0x55b7c36425fb]
(2): bessd(_ZN4bess10PacketPool18CreateDefaultPoolsEm+0x36) [0x55b7c3636946]
(3): bessd(main+0x239) [0x55b7c353e289]
(4): /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xea) [0x7f1c148be09a]
(5): bessd(_start+0x29) [0x55b7c362fab9]
*** Check failure stack trace: ***
```
Needed CAP_SYS_ADMIN

with `-m 0` works without any capabilities

core/dpdk.cc
```diff
@@ -148,6 +163,7 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
   stdout = fopencookie(nullptr, "w", dpdk_log_init_funcs);
 
   disable_syslog();
+  LOG(INFO) << "Initializing DPDK EAL with options: " << rte_args.Dump();
```
krsna1729, line 166: Helpful :)

core/dpdk.cc
```diff
@@ -108,16 +118,21 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
...
   if (dpdk_mb_per_socket <= 0) {
+    rte_args.Append({"--iova", "va"});
     rte_args.Append({"--no-huge"});
 
     // even if we opt out of using hugepages, many DPDK libraries still rely on
     // rte_malloc (e.g., rte_lpm), so we need to reserve some (normal page)
     // memory in advance. We allocate 512MB (this is shared among nodes).
     rte_args.Append({"-m", "512"});
   } else {
+    rte_args.Append({"--iova", "pa"});
```
krsna1729, line 134: what are the capabilities needed?
does it need `iommu=pt`? today we use `intel_iommu=on` only.

core/dpdk.cc
```diff
@@ -108,16 +118,21 @@ void init_eal(int dpdk_mb_per_socket, std::string nonworker_corelist) {
       // Do not bother with /var/run/.rte_config and .rte_hugepage_info,
       // since we don't want to interfere with other DPDK applications.
       "--no-shconf",
+      // TODO(sangjin) switch to dynamic memory mode
+      "--legacy-mem",
```
krsna1729, line 122: we use this too https://github.com/omec-project/upf-epc/blob/master/patches/bess/0003-legacy-mem.patch
CAP_SYS_NICE otherwise

chris3torek: @krsna1729 Based on your own experiment, it looks like `CAP_SYS_ADMIN` is required.  This is not really surprising but doesn't seem to be clear from the man pages.

chris3torek: (oops, attached this at the wrong place, that's for the `--iova pa` arguments)

core/dpdk.cc
krsna1729: @chris3torek i wanted to know why we should force PA and indirectly SYS_ADMIN when using hugepages? Today we use hugepages without needing root.

sangjinhan:
@krsna1729 

> does not seem to be working in our setup
> ...
> Needed CAP_SYS_ADMIN
> with `-m 0` works without any capabilities

Let me clarify, is your use case like this?
* IOMMU is fully turned on, not just passthrough
* want to use a physical NIC (which model?)
* non-privileged user

Are you able to do the above with vanilla DPDK, let's say, testpmd? If so, did you use VA? PA?

Maybe I should let the user specify which IOVA mode to use, rather than (often incorrectly) guessing it? Want to hear your thoughts.

krsna1729:
> * IOMMU is fully turned on, not just passthrough

Yes

> * want to use a physical NIC (which model?)

X710 VFs (PFs too)

> * non-privileged user

from capabilities perspective yes. root user in a container, still not using userns.

> Are you able to do the above with vanilla DPDK, let's say, testpmd? If so, did you use VA? PA?

```
$ docker run --rm -it \
--ulimit memlock=-1 -v /dev/hugepages:/dev/hugepages \
--device=/dev/vfio/82 --device=/dev/vfio/83 --device=/dev/vfio/vfio \
dpdk-app-testpmd testpmd --legacy-mem
...
EAL: Selected IOVA mode 'VA'
....
```

```
$ docker run --rm -it \
--ulimit memlock=-1 -v /dev/hugepages:/dev/hugepages \
--device=/dev/vfio/82 --device=/dev/vfio/83 --device=/dev/vfio/vfio \
--cap-add SYS_NICE \
dpdk-app-testpmd testpmd
...
EAL: Selected IOVA mode 'VA'
...
```

> Maybe I should let the user specify which IOVA mode to use, rather than (often incorrectly) guessing it? Want to hear your thoughts.

https://doc.dpdk.org/guides/rel_notes/known_issues.html#unsuitable-iova-mode-may-be-picked-as-the-default

Yes we can leave it for DPDK to figure out (not passing anything) and allow user to freely append EAL args that wont have an impact on BESS. Whitelisting PCI is one thing I wanted to pass.

sangjinhan:
@krsna1729 added `--iova <pa|va>` option to force an iova mode. If unspecified, `va` is used for `-m 0`, or `pa` otherwise.

krsna1729:
Can we add the arg only if the user passes? If no, that's fine.
```
if (FLAGS_iova != "")
    rte_args.Append({"--iova", FLAGS_iova});
```

sangjinhan:
@krsna1729 that would be my preference, but DPDK often trying `pa` even if it runs without the required capability, resulting in a crash :( I wish its heuristic to decide default iova mode also considered the current capability.

melvinw:
Thanks! cat

PR #971, opened 2020-04-10
Resolving the invalid RSS configuration problem for vhost PMD ports.

sarsanaee:
There is a flow_type_rss_offloads condition added to the PMD port driver
which makes issues for vhost ports requiring more than one queue. Given
that there is no RSS support for vhost ports yet, it should be considered
in that condition.

Before this commit, we weren't able to have a vhost PMD port with more
than one queue, and even the example bess script in `conf/port/vhost/vhost.bess`
with more than one queue --which isn't the default config-- used to have errors.

Signed-off-by: Alireza Sanaee <sarsanaee@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,3 +61,4 @@ Please add your name to the end of this file and include this file to the PR, un
 * Ryan Standt
 * Tim Rozet
 * Alireza Sanaee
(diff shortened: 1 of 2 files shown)


sangjinhan:
Jjkybd3nSces1mo7ozYc9IKz_500

PR #974, opened 2020-05-05
Fix the vhost-user initialization issue and its examples.

sangjinhan:
It has been recently observed that the PMD driver may often fail to initialize vhost-user ports correctly; in the guest DPDK-based applications (such as testpmd) cannot receive packets from some ports for no apparent reason.

* guest -> host works fine
* host -> guest works fine only if the virtio port in the guest is managed by Linux TCP/IP
* when guest is a container, not a vm, it works

While the root cause of the problem is unclear (just guessing some random bug of the DPDK vhost-user implementation, as it had in 17.02), adjusting the port initialization sequence somehow fixed the problem.

This PR also updates the outdated `launch_vm.py` and `launch_container.py` examples for the latest DPDK version, with some minor tweaks.


melvinw:
Thanks! cat

PR #975, opened 2020-05-07
Add metadata with attribute to the Timestamp module.

ajamshed:
This patch is useful when Timestamp and Measure modules are used
in tandem. In current version, the default location to store the
metadata in Timestamp module is at Eth/IP4/Udp offset. This may
lead to bugs if the pipeline uses encap/decap operations (header
sizes are changed, and timestamp value can get corrupted).

We propose using attribute option to store timestamp value as
default.

Signed-off-by: Muhammad Asim Jamshed <muhammad.jamshed@intel.com>
Signed-off-by: Saikrishna Edupuganti <saikrishna.edupuganti@intel.com>

### DIFF
--- a/core/modules/measure.cc
+++ b/core/modules/measure.cc
@@ -85,7 +85,9 @@ CommandResponse Measure::Init(const bess::pb::MeasureArg &arg) {
   if (arg.offset()) {
     offset_ = arg.offset();
   } else {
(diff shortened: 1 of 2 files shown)


ajamshed:
Okay sure. I will send in a new PR. Closing this one.

PR #976, opened 2020-05-08
Allow 'Debian' to use ansible dependency build script

shinae-woo:
The ansible script assumes 'Ubuntu' only for the logic to check distro and it's version to add ppa source and build from source appropriately as needed. This PR allows 'Debian' to add build dependencies. 

(Tested on Debian10+)

### DIFF
--- a/env/build-dep.yml
+++ b/env/build-dep.yml
@@ -36,7 +36,7 @@
         - libgrpc++-dev
         - libprotobuf-dev
         - protobuf-compiler-grpc
-      when: ansible_distribution == 'Ubuntu' and ansible_lsb.major_release|int >= 18
+      when: ansible_distribution == 'Debian' or (ansible_distribution == 'Ubuntu' and ansible_lsb.major_release|int >= 18)
 
     # for Ubuntu < 18.04, grpc has to be installed manually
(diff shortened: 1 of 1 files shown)


env/build-dep.yml
```diff
@@ -36,7 +36,7 @@
         - libgrpc++-dev
         - libprotobuf-dev
         - protobuf-compiler-grpc
-      when: ansible_distribution == 'Ubuntu' and ansible_lsb.major_release|int >= 18
+      when: ansible_distribution == 'Debian' or ansible_distribution == 'Ubuntu' and ansible_lsb.major_release|int >= 18
```
chris3torek, line 39: minor: I had to hunt for a claim that `and` has higher precedence than `or` (I found one here: https://riptutorial.com/ansible/example/12269/when-condition). Maybe use explicit parentheses to avoid the need to check for this... 😺

shinae-woo: Oh. you're not the first one who asked about the precedence. So it looks like this statement deserve parentheses :)

shinae-woo: (Just curious, is there any well-known language in which `and` and `or` have the same precedence?)

PR #978, opened 2020-05-08
Add CPU flag to customize the `-march` build option

sangjinhan:
BESS versions with DPDK 17.11:
- DPDK was always built with `-march=nehalem`
- BESS was built with `CXXARCHFLAGS`

BESS versions with DPDK 19.11.1:
- DPDK was always built with `-march=native`
- BESS was built with `CXXARCHFLAGS` environment variable

After this PR:
- both DPDK and BESS are built with `CPU` environment variable (`native` if not specified)

### DIFF
--- a/build.py
+++ b/build.py
@@ -288,14 +288,14 @@ def download_dpdk(quiet=False):
 
 
 def configure_dpdk():
-    print('Configuring DPDK...')
+    arch = os.getenv('CPU', 'native')
+    print('Configuring DPDK... (CONFIG_RTE_MACHINE=%s)' % arch)
 
+    set_config('%s/config/defconfig_%s' % (DPDK_DIR, DPDK_TARGET),
+               "CONFIG_RTE_MACHINE", arch)
     check_kernel_headers()
-
     check_mlx()
-
     generate_dpdk_extra_mk()
-
     cmd('make -C %s config T=%s' % (DPDK_DIR, DPDK_TARGET))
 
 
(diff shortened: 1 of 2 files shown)


core/Makefile
```diff
@@ -108,8 +108,8 @@ endif
 # these headers.  Should fix the warnings.  Using -isystem also disables
 # -MMD dependency recording (should we use -MD?).
 COREDIR := $(abspath .)
-CXXARCHFLAGS ?= -march=native
-CXXFLAGS += -std=c++17 -g3 -ggdb3 $(CXXARCHFLAGS) \
+CPU ?= native
+CXXFLAGS += -std=c++17 -g3 -ggdb3 -march=$(CPU) \
```
chris3torek, line 112: I'm fine with this change, I'll just note that the old way allowed one to not specify `-march=...` at all by setting `CXXARCHFLAGS` to something weird. 😺 (This way doesn't do that but seems clearer / cleaner to me)

build.py
```diff
@@ -197,12 +197,17 @@ def set_config(filename, config, new_value):
...
 
+    found = False
     with open(filename, 'w') as fp:
         for line in lines:
             if line.startswith(config + '='):
+                found = True
                 line = '%s=%s\n' % (config, new_value)
             fp.write(line)
 
+    assert found, '"%s" is not found in %s' % (config, filename)
```
chris3torek, line 208: Should we use `assert`, or should we use something that survives if Python optimization is turned on? (This isn't pytest with its weird manipulation of `assert`s...)

(I think it's OK this way, just noting it here.)

melvinw:
Thanks!
cat

PR #979, opened 2020-05-13
port.h: bump kDefault*QueueSize

ravens:
The current value of kDefaultOutQueueSize is too small for some
paravirtualized adapters such as vmxnet3, and as a result those PMDports
fail during attachment. Bumping up the default value.

### DIFF
--- a/core/drivers/pmd.cc
+++ b/core/drivers/pmd.cc
@@ -258,6 +258,23 @@ CommandResponse PMDPort::Init(const bess::pb::PMDPortArg &arg) {
 
   eth_rxconf = dev_info.default_rxconf;
   eth_rxconf.rx_drop_en = 1;
+
+  if (dev_info.rx_desc_lim.nb_min > 0 &&
+      queue_size[PACKET_DIR_INC] < dev_info.rx_desc_lim.nb_min) {
+    int old_size_rxq = queue_size[PACKET_DIR_INC];
+    queue_size[PACKET_DIR_INC] = dev_info.rx_desc_lim.nb_min;
+    LOG(WARNING) << "resizing RX queue size from " << old_size_rxq << " to "
+                 << queue_size[PACKET_DIR_INC];
+  }
+
+  if (dev_info.rx_desc_lim.nb_max > 0 &&
+      queue_size[PACKET_DIR_INC] > dev_info.rx_desc_lim.nb_max) {
+    int old_size_rxq = queue_size[PACKET_DIR_INC];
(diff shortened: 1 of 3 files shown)


sangjinhan:
Thank you for the patch! One potential issue with default queue sizing is that for some QEMU (and other old virtio implementations) versions used to have the queue size limited (hardcoded) to 256. 

How about this?

* We set the default queue size to 1,024 for both inc (rx) and out (tx). 1,024 is just an arbitrary number I chose (many Linux device drivers for hardware NICs have 1k as their default)
* In pmd.cc, we cap the default or user-configured queue size with the `rte_eth_dev_info.*_desc_lim` (with some warning messages when it was out of range)

ravens:
@sangjinhan I fixed the PR with the elements you proposed.

core/drivers/pmd.cc
```diff
@@ -207,6 +207,8 @@ CommandResponse PMDPort::Init(const bess::pb::PMDPortArg &arg) {
 
   int num_txq = num_queues[PACKET_DIR_OUT];
   int num_rxq = num_queues[PACKET_DIR_INC];
+  int size_txq = queue_size[PACKET_DIR_OUT];
```
sangjinhan, line 210: Since only `size_txq` is updated in this function, queue_size[] values may not reflect the reality.

`int &size_txq = ...` might be a quick fix. I am not sure if having a reference of primitive variable is fine with the Google style guide, though.

core/drivers/pmd.cc
```diff
@@ -257,18 +259,37 @@ CommandResponse PMDPort::Init(const bess::pb::PMDPortArg &arg) {
 
   eth_rxconf = dev_info.default_rxconf;
   eth_rxconf.rx_drop_en = 1;
+
+  if (size_rxq < dev_info.rx_desc_lim.nb_min) {
+    size_rxq = dev_info.rx_desc_lim.nb_min;
+    LOG(WARNING) << "resizing RX queue size to " << size_rxq;
```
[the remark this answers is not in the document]
ravens: Good idea, thanks !

core/drivers/pmd.cc
ravens: Yup, so will use queue_size directly, thanks.

ravens:
@sangjinhan just did another pass today to address your comments. Thanks !

sangjinhan:
Merged, thanks!
Jjkybd3nSdcy9ex8vBti8Wi7_500

PR #980, opened 2020-05-13
pmd.cc: add back support for PMD virtio devices

ravens:
Virtio devices are using a unique queue so the current default config
cannot work and return 22 when user attempts to add port to bess. We
fix that by using the number of RX queues to configure the device.

### DIFF
--- a/core/drivers/pmd.cc
+++ b/core/drivers/pmd.cc
@@ -36,11 +36,12 @@
 #include "../utils/ether.h"
 #include "../utils/format.h"
 
-static const rte_eth_conf default_eth_conf(const rte_eth_dev_info &dev_info) {
+static const rte_eth_conf default_eth_conf(const rte_eth_dev_info &dev_info,
+                                           int nb_rxq) {
   rte_eth_conf ret = {};
 
   ret.link_speeds = ETH_LINK_SPEED_AUTONEG;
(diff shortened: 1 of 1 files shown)


core/drivers/pmd.cc
```diff
@@ -50,6 +50,16 @@ static const rte_eth_conf default_eth_conf(const rte_eth_dev_info &dev_info) {
                 dev_info.flow_type_rss_offloads,
   };
 
+  if (nb_rxq < 2)
```
sangjinhan, line 53: It can be simply like this in Line 43?
` ret.rxmode.mq_mode = (nb_rxq > 1) ? ETH_MQ_RX_RSS : ETH_MQ_RX_NONE;`

(I believe `rss_hf` is just ignored unless RSS is used)

sangjinhan: (also please run `clang-format`)

ravens: Thanks, just pushed the corrections after tests. Changing the mode was sufficient to get the port attachment to work again. I also ran clang-format.

ravens:
@sangjinhan I fixed the PR with the elements you proposed.

sangjinhan:
Merged, thanks!

IMG_20200518_103543

PR #986, opened 2020-05-28
Bypass l4/ip4 csum compute/validate for !tcp/!udp

ajamshed:
This patch ensures that ping messages goes through the dataplane pipeline which includes verify checksum module.

Signed-off-by: Muhammad Asim Jamshed <muhammad.jamshed@intel.com>
Signed-off-by: Saikrishna Edupuganti <saikrishna.edupuganti@intel.com>
Signed-off-by: Hyunsun Moon <hyunsun@opennetworking.org>


chris3torek:
Just a note: I'm in the middle of moving to Washington (state) and won't be able to look at this for a while yet.

Issue #917, opened 2019-05-23
[bessctl]: show pipeline fails with python 3.7


krsna1729:
```
Traceback (most recent call last):
  File "/bin/bessctl", line 186, in <module>
    main()
  File "/bin/bessctl", line 183, in main
    run_cli(io.StringIO('\n'.join(cmds)))
  File "/bin/bessctl", line 159, in run_cli
    cli.loop()
  File "/bin/bessctl", line 134, in loop
    super(BESSCLI, self).loop()
  File "/opt/bess/bessctl/cli.py", line 528, in loop
    self.process_one_line()
  File "/opt/bess/bessctl/cli.py", line 440, in process_one_line
    self.call_func(func, args)
  File "/bin/bessctl", line 95, in call_func
    super(BESSCLI, self).call_func(func, args)
  File "/opt/bess/bessctl/cli.py", line 411, in call_func
    func(*args)
  File "/opt/bess/bessctl/commands.py", line 1405, in show_pipeline
    cli.fout.write(_draw_pipeline(cli, 'pkts', '', graph_args=opts))
  File "/opt/bess/bessctl/commands.py", line 1359, in _draw_pipeline
    print('[%s]' % node_labels[m.name], file=f.stdin)
TypeError: a bytes-like object is required, not 'str'
```

JanonWang:
I met the same problem and solved it by shifting to python2.7

krsna1729:
Context - I was using 2.7 and wanted to move to 3.7

fshahinfar1:
I think it is good to be mentioned that BESS needs python2.7 some where in [installation guide](https://github.com/NetSys/bess/wiki/Build-and-Install-BESS).

PR #996, opened 2020-06-24
Fix Python 3.7/3.8 incompatibility

sangjinhan:
This patch fixes miscellaneous compatibility issues of bessctl and its sample scripts for Python 3.7 & Python 3.8.

See #917.


shinae-woo:
IMG_7062

PR #998, opened 2020-06-25
add requirements.txt for pip & virtualenv

sangjinhan:


requirements.txt
```diff
@@ -0,0 +1,4 @@
+scapy
+flask
+grpcio
+protobuf
```
melvinw, line 4: perhaps nail these to the latest known working versions?

shinae-woo: I think, if it has some known limitation for a specific version (e.g., known minimum), we can specify that. If not, I think we shouldn't specify versions, as we don't actually want to force to use that specific version.

PR #1000, opened 2020-06-26
BPF: commands for removing a BPF rule

shinae-woo:
Add `delete` command taking the same input as `add`

### DIFF
--- a/core/modules/bpf.cc
+++ b/core/modules/bpf.cc
@@ -51,6 +51,8 @@
 const Commands BPF::cmds = {
     {"add", "BPFArg", MODULE_CMD_FUNC(&BPF::CommandAdd),
      Command::THREAD_UNSAFE},
+    {"delete", "BPFArg", MODULE_CMD_FUNC(&BPF::CommandDelete),
+     Command::THREAD_UNSAFE},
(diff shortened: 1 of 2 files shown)


core/modules/bpf.cc
```diff
@@ -125,6 +127,25 @@ CommandResponse BPF::CommandAdd(const bess::pb::BPFArg &arg) {
...
+    }
+
+    for (auto i = filters_.begin(); i != filters_.end(); ++i) {
+        if (f.priority() == i->priority &&
+            f.gate() ==i->gate &&
+            f.filter() == i->exp) {
+            filters_.erase(i);
+            i--;
+        }
+    }
```
melvinw, line 143: I don't think `i` is guaranteed to be valid after calling `filters_.erase()` perhaps use this instead?
 
```suggestion
    for (auto i = filters_.begin(); i != filters_.end();) {
        if (f.priority() == i->priority &&
            f.gate() ==i->gate &&
            f.filter() == i->exp) {
            i = filters_.erase(i);
        } else {
            i++;
        }
    }
```

shinae-woo: Oh, thank you. It actually needs to break the loop :)

PR #1006, opened 2020-07-25
Do not send log to stdio when run as daemon

shinae-woo:
The combination of running BESS as a daemon and forward the log to stderr break the process.


sangjinhan:
Merged, thanks

PR #1015, opened 2020-09-01
Fix Mellanox PMDs Not Being Enabled

adithyaphilip:
Hi! I'm a PhD student advised by @justinemarie , and I was having trouble getting BESS to work with a Mellanox NIC. I believe I was able to rectify the issue with the changes in this PR. The fix is two-fold:
1. Allow the MLX4 and MLX5 PMDs to be enabled individually, we don't require both to be present.
2. Enable the the PMDs explicitly while configuring DPDK (DPDK disables them by default).

Tested on Ubuntu 18.04 with a ConnectX-4 Lx NIC running at 25 Gbps, using the latest MLX5 PMD (5.1-0.6.6.0).

### DIFF
--- a/build.py
+++ b/build.py
@@ -232,20 +232,31 @@ def check_bnx():
 
 
 def check_mlx():
-    if check_header('infiniband/ib.h', 'gcc') and check_c_lib('mlx4') and \
-            check_c_lib('mlx5'):
-        extra_libs.add('ibverbs')
-        extra_libs.add('mlx4')
-        extra_libs.add('mlx5')
-    else:
-        print(' - "Mellanox OFED" is not available. '
-              'Disabling MLX4 and MLX5 PMDs...')
-        if check_header('infiniband/verbs.h', 'gcc'):
+    ibverbs_avail = check_header('infiniband/ib.h', 'gcc')
+    mlx4_avail = check_c_lib('mlx4')
+    mlx5_avail = check_c_lib('mlx5')
+
+    if ibverbs_avail:
+        if mlx4_avail:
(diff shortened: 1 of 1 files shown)


tbarbette:
This patch was needed for me. My OFED is installed, but it would tell me it's not without this patch, because I don't have mlx4. I think with recent OFED only what you need is installed. Thanks!

hello0x00:
Thanks for your code, it is very helpful to me

MrWLeo:
Sorry to bother you. I have tried the solution but it still failed. It dropped an error : "Cannot attach PCI device"
And the log writes:
I1224 19:56:38.874467 686 dpdk.cc:72] EAL: Driver cannot attach the device (00000000:af:00.01)
I1224 19:56:38.874562 686 dpdk.cc:72] EAL: Failed to attach device on primary process

hello0x00:
> Sorry to bother you. I have tried the solution but it still failed. It dropped an error : "Cannot attach PCI device" And the log writes: I1224 19:56:38.874467 686 dpdk.cc:72] EAL: Driver cannot attach the device (00000000:af:00.01) I1224 19:56:38.874562 686 dpdk.cc:72] EAL: Failed to attach device on primary process

Check the NIC dmesg information to make sure that the NIC starts correctly. Upgrade bess to the latest version, used with dpdk20.11. Install the latest mlx driver  for your operating system version according to the contents of this document
http://doc.dpdk.org/guides/nics/mlx5.html

py-162157:
thanks, that's work for me

py-162157:
thanks, that works for me

